PR #47, opened 2018-03-25
Applies fixes from calculation maiden voyage

jessex:
@cawarren: this puts the code in a state where it works out of the box. Specific fixes commented inline.

### DIFF
--- a/app.yaml
+++ b/app.yaml
@@ -1,10 +1,13 @@
 runtime: python27
 api_version: 1
 threadsafe: false
-# It is imperative that the module and version values are set to default,
-# due to a GAE MapReduce bug: https://github.com/GoogleCloudPlatform/appengine-mapreduce/issues/103
-module: default
-version: default
+instance_class: F2
+
+# It is imperative that the module and version values are set to default for local and not production,
+# due to a GAE MapReduce bug: https://github.com/GoogleCloudPlatform/appengine-mapreduce/issues/103.
+# Comment these out to run locally. TODO: make this dynamic by environment.
+#module: default
+#version: default
 
 includes:
 - lib/mapreduce/include.yaml
--- a/calculator/calculator.py
+++ b/calculator/calculator.py
@@ -17,6 +17,7 @@
 
 
 from datetime import date
+from dateutil.relativedelta import relativedelta
 import logging
 import webapp2
 
@@ -135,7 +136,7 @@ def count_releases_in_window(start_date, follow_up_period, all_release_dates):
     the given start date
     """
     releases_in_window = [release_date for release_date in all_release_dates
-                          if start_date.replace(year=start_date.year + follow_up_period)
+                          if start_date + relativedelta(years=follow_up_period)
                           > release_date >= start_date]
     return len(releases_in_window)
 
@@ -182,7 +183,7 @@ def relevant_follow_up_periods(release_date, follow_up_periods):
     :return: the array of follow up periods which are relevant to track, i.e. completed or in progress
     """
     return [period for period in follow_up_periods
-            if release_date.replace(year=release_date.year + period - 1) <= date.today()]
+            if release_date + relativedelta(years=period - 1) <= date.today()]
 
 
 def age_at_date(inmate, date):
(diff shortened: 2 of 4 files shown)


app.yaml
```diff
@@ -1,10 +1,13 @@
 runtime: python27
 api_version: 1
 threadsafe: false
-# It is imperative that the module and version values are set to default,
-# due to a GAE MapReduce bug: https://github.com/GoogleCloudPlatform/appengine-mapreduce/issues/103
-module: default
-version: default
+instance_class: F2
+
+# It is imperative that the module and version values are set to default for local and not production,
```
jessex: If these lines are uncommented, we cannot deploy to production as they are deprecated. However, GAE MapReduce relies on them to run locally. Note the TODO to make this dynamic by environment which I have not had a chance to do yet. Also note that this only impacts runtime, i.e. when you trigger the calculator pipeline. So if you are not running calculation locally, you can leave these commented out.

calculator/calculator.py
```diff
@@ -135,7 +136,7 @@ def count_releases_in_window(start_date, follow_up_period, all_release_dates):
     the given start date
     """
     releases_in_window = [release_date for release_date in all_release_dates
-                          if start_date.replace(year=start_date.year + follow_up_period)
+                          if start_date + relativedelta(years=follow_up_period)
```
jessex: Replacing components by hand was not a great idea to begin with. You can get into situations where the date is no longer valid, e.g. if you have February 29 on an actual leap year and add to the year component by a number not divisible by 4. I've updated all date manipulation logic in the calculator, but we should make similar changes in the ingest pipeline.

calculator/calculator.py
```diff
@@ -268,6 +269,10 @@ def find_recidivism(inmate, include_conditional_violations=False):
         record_id = record.key.id()
 
         original_entry_date = first_entrance(record)
+        if original_entry_date is None:
+            # If there is no entry date on the record, there is nothing we can process. Skip it.
```
jessex: There are on the order of 50 records _total_ without an entry date on their record. This is definitely a data issue but it is in super low volume, fortunately. This works around it.

calculator/calculator.py
```diff
@@ -355,7 +360,10 @@ def final_release(record):
     if not record.is_released:
         return None
 
-    return record.last_release_date
+    if record.last_release_date:
+        return record.last_release_date
+    else:
+        return record.last_custody_date
```
jessex: I don't love this and would love suggestions. But some records are marked as no longer in custody but don't have a `last_release_date` set, leading to an error at runtime. In the spur of the moment, I defaulted to `last_custody_date` instead of `last_release_date`, but maybe we should throw these records on the floor. I haven't checked how many are effected, but will do so before we merge this.

calculator/pipeline.py
```diff
@@ -170,7 +170,7 @@ def run(self, region, include_conditional_violations):
             mapper_spec="calculator.pipeline.map_inmate",
             mapper_params=mapper_params,
             reducer_spec="calculator.pipeline.reduce_recidivism_events",
-            shards=8)
+            shards=64)
```
jessex: I suspect this will keep scaling linearly with another couple of powers of 2, but 64 shards was fast enough. More interesting was that the memory issues in the shuffle step were quelled specifically by increasing the shard count past 32 to 64 at our current scale. Simply increasing shards is not a viable strategy in the long-term, as our load becomes more dynamic with more regions, which is another reason to migrate to Cloud Dataflow.

calculator/calculator.py
cawarren: There's something odd in here... Most of those are from records stored for duplicates of the same inmate. I need to figure out a) why that inmate has been duplicated in the system and whether it affects other inmates as well, and b) why it failed to capture any of the criminal record for him.

Filed issue 49 to investigate.

calculator/calculator.py
cawarren: If you could pull a list of those records I'd love to investigate further. Just file an issue and assign to me.

cawarren: I don't know that last custody date will be a good proxy for release cohort, so I'd suggesting ignoring those records.

If we do decide to include them, we should make an exception for records recorded as 1/1/0001 (or 12/31/0000, see issue #50) - a small but non-zero number of DOCCS records have that as the last custody date (e.g., DIN #83A6604)

calculator/calculator.py
jessex: Thanks for the issue!

calculator/calculator.py
jessex: Yeah, you're right. I'll just ignore these records. Change inbound.

I filed #51 to track this.

The `1/1/0001` explains why I started seeing calculated metrics for the release cohort of `1`. Good to know that's filed.

jessex:
Ready for final review, @cawarren.

Issue #40, opened 2018-01-31
Continuous integration and release


jessex:
Especially once testing is finally added (#13), we will want to enable continuous integration and continuous release. Travis CI is probably our best bet but we should survey the landscape to see what the canonically Pythonic solution is in 2018. We'd like our "builds" to include at least linting, testing, and releasing. We might also want continuous deployment.

jessex:
As noted on #69, continuous integration is now and place and continuous release has been spun out into a dedicated issue: #71. This is done.

PR #58, opened 2018-04-16
Adds unit and light integration tests for the calculator.

jessex:
This provides unit test coverage for all of the helper methods across the calculation pipeline, and some light "integration-y" tests for map and reduce. I'd like to have an integration test that runs a full MR pipeline in memory but I think that can be deferred for now.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,2 @@
+[run]
+omit = lib/*,tests/*
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ eggs/
 .eggs/
 lib64/
 parts/
+.pytest_cache/
 sdist/
 var/
 wheels/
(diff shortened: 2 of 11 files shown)


jessex:
This is ready for review now.

jessex:
Whoops. Travis integration is not ready yet (working on it as part of #40). Just silenced it for now. Please ignore.

jessex:
One note on test coverage as a tool for our project: this uses pytest-cov to output a report showing where there is missing test coverage once pytest completes, but it does not have any impact on "the build" until we have gotten continuous integration into place and it's not as visible or navigable as it could be. As part of getting continuous integration into place with #40, we will use Coveralls to output pretty, visible coverage reports and consider integrating the coverage output into build status.

In the Java world, we tend to use something like Jacoco or Cobertura and configure code coverage thresholds, e.g. percentage of minimum lines or branches covered, such that if the code coverage drops below that threshold, i.e. by adding new code and not testing it, the build will fail. I'd like something similar here.

jessex:
Squashing and merging this as per offline agreement with @cawarren.

Issue #60, opened 2018-04-21
Clean up docstrings to match Google Style Guide


jessex:
I was negligent when I first started documenting and assumed that the PyCharm-suggested format was PEP-approved. We want to defer to the [Google Python Style Guide](https://google.github.io/styleguide/pyguide.html) in general, and [docstrings specifically](http://google.github.io/styleguide/pyguide.html?showone=Comments#Comments). 

Update all docstrings across the board to match.

_Note_: there's also the ability to do "doctests" in Pytest, wherein example input and output documented in a specific format in docstrings are automatically scanned and tested. I don't love that personally. I feel it's a bit too magical. I'm open to debating that, but for now let's skip it.

jessex:
This is officially done.

PR #64, opened 2018-04-23
Linting and Google Style

jessex:
DO NOT MERGE, please. This is issued against the `test-catchup` branch because I also captured linting of the test code. I will reissue this against master once `test-catchup` is merged. But this is helpful for anyone with anything in-flight right now.

The first commit here enables and configures Pylint. The ensuing fixup commits gradually roll out the changes it enforces on separate packages in the codebase. I stop short of the `scraper.us_ny` package, because it is about to undergo a significant overhaul and it will be easier and incur less duplicate effort to simply bring it into stylistic accordance as part of the generalization work.

To run it across all of the code we care about, run the command in the README: `pylint *.py calculator models scraper tests`. It's annoying that we have to plug in each package like this, but there appears to be no way to configure `.pylintrc` to specify packages and files like we can with Pytest configuration, and it's really our own fault: if all the source was inside of one `recidiviz` package, then we could just run it over top of that. We'll save that for another day.

This `.pylintrc` configuration is copied directly from Google. They very their Pylint configuration from artifact to artifact as needed, but that was taken from a fairly general Python project and should work for our purposes. 

All told, this PR handles #60 and a linting-specific issue I neglected to file, and will be incorporated in the continuous integration work of #40, for which I have a proof of concept sitting on my machine waiting for these test and linting PRs to be merged (I opted not to overwhelm with another look-ahead PR).

With this in place, we will be able to maintain stylistic integrity and keep code quality high, while staying true to the style guidelines we have chosen. There are other static analysis tools we may choose to incorporate later alongside Pylint, such as Pyflakes, but those will be separate tasks.


jessex:
Closing this in favor of #66.

PR #66, opened 2018-04-24
Linting to Google Python Style Guide

jessex:
This replaces #64, issued against master.

The first commit here enables and configures Pylint. The ensuing fixup commits gradually roll out the changes it enforces on separate packages in the codebase. I stop short of the scraper.us_ny package, because it is about to undergo a significant overhaul and it will be easier and incur less duplicate effort to simply bring it into stylistic accordance as part of the generalization work.

To run it across all of the code we care about, run the command in the README: pylint *.py calculator models scraper tests. It's annoying that we have to plug in each package like this, but there appears to be no way to configure .pylintrc to specify packages and files like we can with Pytest configuration, and it's really our own fault: if all the source was inside of one recidiviz package, then we could just run it over top of that. We'll save that for another day.

This .pylintrc configuration is copied directly from Google. They very their Pylint configuration from artifact to artifact as needed, but that was taken from a fairly general Python project and should work for our purposes.

All told, this PR handles #60 and a linting-specific issue I neglected to file, and will be incorporated in the continuous integration work of #40, for which I have a proof of concept sitting on my machine waiting for these test and linting PRs to be merged (I opted not to overwhelm with another look-ahead PR).

With this in place, we will be able to maintain stylistic integrity and keep code quality high, while staying true to the style guidelines we have chosen. There are other static analysis tools we may choose to incorporate later alongside Pylint, such as Pyflakes, but those will be separate tasks.


jessex:
@cawarren I think that your mega PR is close enough that we should defer this PR until that one is merged. Once we merge that, I'll resolve merge conflicts here and rebase, then merge.

jessex:
As per our contribution guidelines, it is now officially mandatory to ensure the `pylint` command in the README passes for your change before merge: `pylint *.py calculator models scraper tests utils` (with the exception of scraper for now). The definition of "pass" is fuzzy until we get our CI work completed in #40. For now, it means no increase in the identified lint issues beyond what already exists, with room for exceptions. 

As of now, that output should be (reduced for readability):
```bash
$ pylint *.py calculator models tests utils
Using config file /Users/clementinejacoby/Documents/projects/recidiviz-data/.pylintrc
************* Module calculator.pipeline
W:140,18: Use of eval (eval-used)
```

(There's another reported issue which Andrew is going to take care of, a small mishap from a prior merge, in case you try to replicate this in the next several hours. The above issue is the only "legitimate" one for now.)

jessex:
I'm merging this straight away, as per offline discussion with @cawarren. This gets all of our codebase into proper style, with the exception of the `scraper.us_ny` package which is about to undergo heavy refactoring. As part of that refactoring, it will be brought into stylistic consistency.

PR #68, opened 2018-04-30
Update invalid_params to invalid_input

cawarren:
Fix for method name change.


jessex:
This should be submitted as a pull request from an entirely new branch instead of reusing ny-rescrape. We could squash and merge here and change the commit message to make it seem as though it was a new branch, but as best practice, we should never reuse branches once merged.

I approve the changes (:shipit:) so let's just make a commit from a new branch, reissue a PR, and click the green button.

PR #69, opened 2018-05-02
Adds continuous integration via Travis CI.

jessex:
This addresses #40. We are not pursuing continuous release yet because we do not have any release strategy defined yet. I'm going to file a separate issue for that (#71).

This uses Travis CI for continuous integration, including:
* Failing the build if linting fails or if testing fails
* Tracking code coverage via Coveralls
* PR integration for both Travis build status and Coveralls tracking (as can be seen below on this very PR!)
* Badges for both of the above on the README
* A couple of linting fixes
* A `requirements.txt` file to make installation for the project easier: `pip install -r requirements.txt`

_Note_: There are definitely lines in here that are unnecessary because I `pip freeze`'d what was installed in my non-virtualenv controlled environment. I'm going to file an issue to integrate `virtualenv` and clean up `requirements.txt` in the process (#70).


matthewroot:
@jessex I'm curious if you evaluated other code coverage tools. We were using coveralls for awhile and I found it underwhelming. We're in the process of moving over to [Codecov](https://codecov.io/) which seems nice but I don't have a personal opinion on yet. I've also heard good things about [Code Climate](https://codeclimate.com/) but don't have firsthand experience with it.

jessex:
I briefly evaluated Codecov and actually preferred Coveralls, slightly. But I'm sure your squad has more experience on this so I'm happy to hear why I'm wrong.

I did not look at Code Climate but from a cursory glance, I see that it is _not_ free. Before investing in any such tool with $$, I want to make sure the team is getting any use out of this tooling. Since it's a 2-week free trial, I'd rather use those 2 weeks when we have more varied activity across the team.

jessex:
Admittedly, I am coming from a non-Pythonic background (most production experience in Java, and code coverage usage with either Jacoco or Cobertura) and trying to quickly ramp into a more Pythonic approach. 

Coveralls made me slightly more comfortable because, perhaps, it felt more like the Java coverage tools I had used previously, and had the detail I was looking for as I dug through it. Let's take a spin with Coveralls and if we feel we are not getting the value we want, it won't be hard to try Codecov on a branch or two. Sound fair?

Thanks for the insight!

jessex:
Merging this per offline agreement with @cawarren and implicit approval from @matthewroot.

PR #74, opened 2018-05-04
Adds local.yaml file to skip_files

jessex:
@cawarren: please take a look here.

[See the skip_files docs here](https://cloud.google.com/appengine/docs/standard/python/config/appref).

This should avoid accidentally uploading `local.yaml` to the production GAE deployment. It should not be at too much risk to include it in the deployment, but I think as standard practice we should skip any files that are sensitive and not required in production.


app.yaml
```diff
@@ -61,3 +61,11 @@ libraries:
 
 env_variables:
   GAE_USE_SOCKETS_HTTPLIB : 'true'
+
+skip_files:
+- ^(.*/)?#.*#$
```
jessex: These regular expressions are the default value and if you want to add custom values, you have to make all of the default values explicit, [as described here](https://cloud.google.com/appengine/docs/standard/python/config/appref) under "skip_files".

app.yaml
```diff
@@ -61,3 +61,11 @@ libraries:
...
+- ^(.*/)?.*~$
+- ^(.*/)?.*\.py[co]$
+- ^(.*/)?.*/RCS/.*$
+- ^(.*/)?\..*$
+- local.yaml
```
cawarren: This should work as-is, but just to specific (in case of similar filenames in the future), should we make it something like `^local\.yaml$`?

cawarren:
This is a great find, I didn't know about this feature. Thanks for flagging / submitting the change.

app.yaml
jessex: You're right, that's safer. I just tested that out on a variety of sample file names and it locks specifically to local.yaml. Will change and merge this.

PR #76, opened 2018-05-06
Rescrape cleanup

cawarren:
Cleans up some long lines, fixes a few minor bugs.

Out of band of the code change, the `local.yaml` file has been updated on the internal Sites page as well.

PR #78, opened 2018-05-08
Applies minor fixes left behind by the snapshot scraping lift.

jessex:
@cawarren I know you've got a ton on your plate right now so I took a pass at applying the fixes at the center of #76 which there appear to be branch issues with. If you want to make that PR happen, we can close this out and rebase that one. But we can also just approve this and close that one to move on if you'd like.


cawarren:
Thank heavens (and you).

Someday I'll pick up a giant tome on Git and learn the reason I ran into so many issues cherry-picking those commits, but for now this unblocks and gets me back to a clean state. Thanks for the hand on this one.

PR #79, opened 2018-05-09
Fixes recidivism calculation methodologies

jessex:
As the deleted TODO notes, I had previously spotted a bug in the calculation of offender-based recidivism metrics that was leading to a small reduction in value, i.e. offender-based recidivism was being under-calculated slightly: less than 1% at scale, in later follow-up periods only, leading to recidivism rate graphs that went slightly but noticeably _down_ over time, which is illogical. 

In the course of an evening walk, though, I realized that both of the methodologies were actually slightly reduced in different ways.

**Offender-based**
* I misinterpreted prior art in the field (the _Rhodes_ paper†). In cases where one person was reincarcerated multiple times in a given follow-up period as part of a given release cohort, I was calculating the recidivism rate as 1/k, where k=the number of releases in that period, and adding a single positive instance with that as the value. As can be seen in the example below, that leads to a decrease once you pass into the first period with multiple releases. 
* In reality, we are trying to hold the recidivism count at one instance in that scenario. The 1/k idea came from the fact that the actual calculation mechanism in the referenced paper is fundamentally different. By just adding a positive instance with a value of 1 as soon as we hit our first instance of recidivism after a given release, we fulfill the offender-based methodology.

**Event-based**
* I was not counting each unique positive instance of recidivism relative to each release cohort. In cases where one person was reincarcerated multiple times in a given follow-up period as part of a given release cohort, I was adding a positive instance with a value of 1 as soon as the first instance was reached and not accounting for later instances (they would be accounted for separately as part of calculations for their own release cohort, but we need to count it in both places).
* In reality, as soon as a second instance is reached while still within range of the first release cohort, we should be adding another positive recidivism instance into the array of results, as it is another _event_ to measure. 

This hand-drawn example should help illuminate the fix, as should the test cases I have added and modified (hopefully!). The top left shows Original entry in 2006, Leaving in 2008, and Returning in 2010. Then Leaving again in 2012 and Returning again in 2014.

img_20180509_232707

On the bottom are the incorrect methodologies being done today, with the artificial decreases in recidivism rate pointed out. Above that is what the calculation should be. Event-based can be seen to be fixed by adding _another_ positive recidivism instance once it has been reached in a follow-up period from the release cohort, e.g. period 6 after cohort 2008. Offender-based can be seen to be fixed by holding the positive value of 1 steady instead of inanely decreasing it.

† "Following Incarceration, Most Released Offenders Never Return to Prison." Rhodes et. al., 2014.


jessex:
Planning to do another pass in `calculator.py` with use some well-named extraction to make it easier to follow the logic in this class. It's getting closer to simple but not quite there yet. That will be a subsequent PR.

PR #80, opened 2018-05-11
Makes two small runtime fixes to the scraper

jessex:
As identified by @ohinds-vera:
* While `is not None` is the Pythonic choice, it does not working inside of ndb `filter` calls.
* There was an `UnboundLocalError` happening in the lazy initialization of the `PROXY_URL` module variable in `queue_control.py`. But the parsing of that variable is extremely cheap and the impacted function is called sparsely, so we'll just eagerly initialize it local to the function.

With these two fixes applied, ingest runs successfully locally.


jessex:
That tiny drop in code coverage comes about because of the deletion of covered code in the `queue_control.py` fix. I think that's acceptable.

Issue #34, opened 2018-01-31
Implement conditional violation calculation


jessex:
Add support for splitting out metrics including violations of conditional release, e.g. recidivism from parole. There are obviously strong opinions in both directions as to whether conditional violations should be included in recidivism metrics. Our goal is to provide metrics that omit conditional violations as well as metrics that include conditional violations, in line with our general philosophy of letting the wiser among us choose which metrics are relevant to them.

To be clear, the goal is not to add metrics that measure recidivism _only_ for conditional violations. For each metric, there should be one version that includes conditional violations alongside new violations and one version that includes only new violations. There is already a query parameter that can be provided to the calculation pipeline to toggle whether conditional violations are calculated (defaults to false).

PR #81, opened 2018-05-11
Calculation readability improvements

jessex:
This improves the recidivism calculator readability in two ways:
* First commit: Performs more self-documenting function extraction with an eye towards shortening the longer, primary functions, and making the functions they delegate to read more naturally
* Second commit: Breaks up `calculator.py` into 3 separate classes and re-packages all of it under a `recidivism` subpackage to make way for the upcoming population calculator


calculator/recidivism/calculator.py
```diff
@@ -0,0 +1,411 @@
+# Recidiviz - a platform for tracking granular recidivism metrics in real time
```
jessex: There is no functional code change in here. It was moved in the repackaging, and code was extracted into `recidivism_event.py` and `identifier.py`.

calculator/recidivism/identifier.py
```diff
@@ -0,0 +1,327 @@
+# Recidiviz - a platform for tracking granular recidivism metrics in real time
```
jessex: There is no functional code change in here. It was moved in the repackaging, and it was extracted out of `calculator.py`.

jessex:
I recommend reviewing each commit in order. The first one will show the code improvements. The second one will show the reorganization.

calculator/recidivism/identifier.py
```diff
@@ -0,0 +1,327 @@
...
+        include_conditional_violations: a boolean indicating whether or not
+                to include violations of conditional release in recidivism
+                calculations.
+
+    Returns:
+        A dictionary from release cohorts to recidivism events for the given
+        inmate in that cohort. No more than one event can be returned per
+        release cohort per inmate.
+    """
+    records = Record.query(ancestor=inmate.key)\
```
jbrambleDC: I know its "anti-google" but I'd be in favor of switching to 120 blocks just to avoid having to chain these calls.

calculator/recidivism/identifier.py
```diff
@@ -0,0 +1,327 @@
...
+        .order(Record.custody_date)\
+        .fetch()
+    snapshots = Snapshot.query(ancestor=inmate.key)\
```
jbrambleDC: nit: I prefer whitespace after multi-line function calls. does pylint care?

calculator/recidivism/identifier.py
```diff
@@ -0,0 +1,327 @@
...
+        else:
+            # If there is a record after this one and they have been released,
+            # then they recidivated. Capture the details.
+            if record.is_released:
+                later_records = records[index + 1:]
+
+                event = for_intermediate_record(record, later_records,
+                                                snapshots, original_entry_date,
+                                                release_date, release_facility)
+                recidivism_events[release_cohort] = event
```
jbrambleDC: so, explain to me what happens when there is more than one arrest/sentence/release within one release cohort ?

jbrambleDC: In other words, imagine I got arrested, convicted and released twice in 2012 for instance

calculator/recidivism/identifier.py
```diff
@@ -0,0 +1,327 @@
...
+        release_facility: the facility they were last released from for this
+            record
+
+    Returns:
+        A non-recidivism event if released from this record. None otherwise.
+    """
+    # There is only something to capture
+    # if they are out of prison from this last record
+    if record.is_released:
+        logging.debug("Inmate was released from last or only "
```
jbrambleDC: why don't we start off passing params to logging.debug() instead of subbing before passing?

calculator/recidivism/identifier.py
```diff
@@ -0,0 +1,327 @@
...
+                      "record %s. No recidivism." %
+                      record.key.id())
+
+        return RecidivismEvent.non_recidivism_event(original_entry_date,
+                                                    release_date,
+                                                    release_facility)
+
+    logging.debug("Inmate is still incarcerated for last or only "
```
jbrambleDC: Here to for subbing.

calculator/recidivism/identifier.py
```diff
@@ -0,0 +1,327 @@
...
+            record
+
+    Returns:
+        A recidivism event.
+    """
+    logging.debug("Inmate was released from record %s and went "
+                  "back again. Yes recidivism." %
+                  record.key.id())
+
+    recidivism_record = later_records[0]
```
jbrambleDC: werent these already sorted by time, cant we just pass the recidivism record instead of all later records?

calculator/recidivism/recidivism_event.py
```diff
@@ -0,0 +1,128 @@
...
+                from for this record.
+
+        Returns:
+            A RecidivismEvent for an instance where recidivism did not occur.
+        """
+        return RecidivismEvent(False, original_entry_date, release_date,
+                               release_facility)
+
+    def __repr__(self):
+        return "<RecidivismEvent recidivated:%s, original_entry_date:%s, " \
```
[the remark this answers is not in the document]
jessex: Sure thing.

calculator/recidivism/identifier.py
jessex: Yeah, let's get started on that henceforth.

calculator/recidivism/identifier.py
jessex: Great question. In short, it's not supported yet. What would happen is that the last record encountered, which is to say the last instance of release in the calendar year, would win and the other record would be lost. But it's not possible (without manual error) in the DOCCS system and likely is not possible in any state prison system: you don't go to state prison unless it's for a felony, in which case your sentence is typically 18 months or longer in most cases we have seen. You could potentially get a conditional release (e.g. good behavior or parole) before the year is up, but if you violated those terms then you'd go back _on the same record_, not a new one. I see no way for this to happen with prison data, and that's why it's not supported yet.

However, it certainly could happen with _jail_ data. Furthermore, I am working on #34 as we speak, where extreme edge cases could lead to a conditional release and a full release in the same calendar year in the prison system. As part of those changes, I will be changing this data structure to support multiple recidivism events for a single release cohort. Even in the conditional prison edge cases and in the jail data to come, it would almost never happen, but it will not be impossible. So that support is coming.

Good insight, @jbrambleDC.

jessex:
Tiny decrease because of simplification of code that was covered. Acceptable, _for now_.

jessex:
Build passes and local run of the recidivism pipeline work with no apparent regression. Merging!

PR #91, opened 2018-05-15
Style clean up

jbrambleDC:
This simply cleans up the "style" in calculator.recidivism. for reference here is the linter output

A few changes here are worth to note, and open to debate:
1. Single quotes on every string that is static or behaves like a symbol
2. Double quotes on dictionary keys, and any string thats interpolated
3. Passed string formatting args to logging functions, instead of pre-formatting
4. Use .format() anywhere where strings are formatted, except for logging.


calculator/recidivism/pipeline.py
```diff
@@ -170,13 +170,13 @@ def get(self):
...
         if include_conditional_violations_param not in ['true', 'false']:
             error_message = "include_conditional_values must be true or " \
-                            "false. Bad param: %s" % \
-                            include_conditional_violations_param
-            logging.error(error_message)
+                            "false. Bad param: %s"
+
+            logging.error(error_message, include_conditional_violations_param)
 
             self.response.write(error_message)
```
jessex: Now this error message doesn't have the param included in the response. You could call `.format()` on this directly when passing into `write()`, or call `.format()` at initialization time and pass it in formatted to both places.

calculator/recidivism/pipeline.py
```diff
@@ -170,15 +170,17 @@ def get(self):
...
         if include_conditional_violations_param not in ['true', 'false']:
             error_message = "include_conditional_values must be true or " \
-                            "false. Bad param: %s" % \
-                            include_conditional_violations_param
-            logging.error(error_message)
+                            "false. Bad param: %s"
+
+            logging.error(error_message, include_conditional_violations_param)
+
+            self.response.write(error_message.format(
```
jessex: Just to make sure, Python's `String.format` works with both C-style format args (`%s`) and the format args you're using elsewhere (`{}`)?

jbrambleDC: I chose to use `%`, but since this clearly an error case, and will only format as such, using .format() would be fine. let me clean that up

calculator/recidivism/pipeline.py
```diff
@@ -170,15 +170,17 @@ def get(self):
...
         if include_conditional_violations_param not in ['true', 'false']:
             error_message = "include_conditional_values must be true or " \
-                            "false. Bad param: %s" % \
-                            include_conditional_violations_param
-            logging.error(error_message)
+                            "false. Bad param: %s"
+
+            logging.error(error_message, include_conditional_violations_param)
+
+            self.response.write(error_message %
```
jessex: Sorry for all the crossed wires here, but why not do:
```
error_message = 'include_conditional_values must be true or false. Bad param: {}'.format(include_conditional_values_param)
logging.error(error_message)

self.response.write(error_message)
```
?

jbrambleDC: done, we generally want to avoid that, to avoid formatting strings when they won't be logged. but since this an error case where logging will assuredly happen, it is okay to pre-format.

jessex: You updated as I commented. Good to go on this one.

PR #95, opened 2018-05-16
bumps pyOpenSSL

jbrambleDC:
if you have a newer version of OpenSSL, pyOpenSSL will not compile unless you use >=0.14. Further reading here:
https://github.com/pyca/pyopenssl/issues/276

I have successfully ran tests and built locally, as well as launched the app. We can wait until I have fully run ingestion locally before merging this.


jessex:
I would amend the commit message to also include the dateutil bump here. Once you have successfully run regression locally, let me know and I'll do the same on my machine to confirm what happens with my environment and version of OpenSSL.
```bash
16:12 $ openssl version
LibreSSL 2.2.7
```

jbrambleDC:
commit amended, there is no pressing need to bump dateutil at this time, especially with venv coming, however this version of dateutil will not play nice with jupyter, bokeh, and other interactive/visualization tools, which seems like something one may want at a future date, even if not in this repo.

jbrambleDC:
`works on my machine` . Why don't you independently verify?

Issue #94, opened 2018-05-16
No Documentation on creating "local.yaml"


jbrambleDC:
#74 mentions adding this file to .gitignore and ensuring it doesn't get uploaded to GAE due to its sensitive nature. 

As it stands currently the README.md in the base directory does not document how to create this, so I get the following error when running locally:
```
webapp2/_internal_error : [Errno 2] No such file or directory: 'local.yaml'
Traceback (most recent call last):
  File "/Users/jordanbramble/google-cloud-sdk/platform/google_appengine/lib/webapp2-2.5.2/webapp2.py", line 1535, in __call__
    rv = self.handle_exception(request, response, e)
  File "/Users/jordanbramble/google-cloud-sdk/platform/google_appengine/lib/webapp2-2.5.2/webapp2.py", line 1529, in __call__
    rv = self.router.dispatch(request, response)
  File "/Users/jordanbramble/google-cloud-sdk/platform/google_appengine/lib/webapp2-2.5.2/webapp2.py", line 1278, in default_dispatcher
    return route.handler_adapter(request, response)
  File "/Users/jordanbramble/google-cloud-sdk/platform/google_appengine/lib/webapp2-2.5.2/webapp2.py", line 1102, in __call__
    return handler.dispatch()
  File "/Users/jordanbramble/google-cloud-sdk/platform/google_appengine/lib/webapp2-2.5.2/webapp2.py", line 572, in dispatch
    return self.handle_exception(e, self.app.debug)
  File "/Users/jordanbramble/google-cloud-sdk/platform/google_appengine/lib/webapp2-2.5.2/webapp2.py", line 570, in dispatch
    return method(*args, **kwargs)
  File "/Users/jordanbramble/recidiviz-data/utils/environment.py", line 91, in check_env
    return func(request_handler, *args, **kwargs)
  File "/Users/jordanbramble/recidiviz-data/utils/auth.py", line 110, in auth_and_call
    return func(request_handler, *args, **kwargs)
  File "/Users/jordanbramble/recidiviz-data/tests/utils/populate_test_db.py", line 170, in get
    load_env_vars()
  File "/Users/jordanbramble/recidiviz-data/tests/utils/populate_test_db.py", line 190, in load_env_vars
    env_vars = environment.load_local_vars()
  File "/Users/jordanbramble/recidiviz-data/utils/environment.py", line 108, in load_local_vars
    with open("local.yaml", 'r') as ymlfile:
[...]

PR #96, opened 2018-05-16
Add local yaml docs, caution against forking

cawarren:
This adds two new sections to the documentation:

(1) Warning others considering forking the project for scraping the same systems that we are that they should contact us first.

(2) Adding details on providing environment variables to the local server for development.


cawarren:
This PR should address issue #94

README.md
```diff
@@ -41,6 +56,43 @@ and clone the recidiviz repo from Github.
...
 Then install project dependencies: `pip install -r requirements.txt`.
 
+#### Adding environment variables
+Recidiviz depends on sensitive information to run. This data is stored in environment variables, which should be added
+manually to your production environment (see `models/env_vars` for more information on the datastore kind used).
+
+For local testing, these environment variables are loaded from a `local.yaml` file that should be added to your top-level
+project directory.
+
+_Note: Recidiviz team members and partners can download a pre-populated `local.yaml` file from
```
jessex: As we discussed on Slack, this is the only part that is potentially contentious. In general, we agree about a desire to not directly link to internal docs from externally-facing docs. I think this should go in the onboarding doc for now, but in time we will stand up an internal runbook (in fact, we could repurpose our "Scrapo" doc maybe).

In short, I'd remove this and focus on including this in appropriate internal docs. Sound fair?

cawarren:
I've kept the note for team members, but removed the link. Check the
current version, let me know if you still see qualms with this.

README.md
```diff
@@ -41,6 +56,43 @@ and clone the recidiviz repo from Github.
...
+For local testing, these environment variables are loaded from a `local.yaml` file that should be added to your top-level
+project directory.
+
+_Note: Recidiviz team members and partners can download a pre-populated `local.yaml` for local testing - check your
+onboarding document for details._
+
+If you're building your own `local.yaml` file from scratch, it should look similar to the following:
+
+```
+# Test and configuration vars for local development with <your project>.
```
jessex: Was just inspired to try something that I think is a better experience: create a `local.example.yaml` file checked into the repository that has the basic structure outlined below with placeholders for the values (you could just copy and paste that `env_vars` block into the file). Then in the instructions here, instruct users to do `cp local.example.yaml local.yaml` and fill in the placeholders. As we add new required or optional environment variables, we just add them to the example so it's always clear what needs to be set, without updating these docs.

Sound reasonable?

cawarren: Yep, that does seem quite a bit cleaner. Updated to take that approach.

jbrambleDC:
Going to go ahead and close #94 as this is approved.

jbrambleDC:
The medium term win here would be to manage local.yaml with secrets management. A reasonable approach may be to just encrypt it in a cloud storage (is this the name?) bucket and allow the application to pull the secret from the bucket for users who are authenticated. 

Unauth'd users can be instructed via documentation to create their own local.yaml from the example and place it in their own storage. 

This will become increasingly easier to manage if we eventually added some generic terraform.

I also had a second thought, isn't this file already in some form of cloud storage? How about encrypt it, and have the app pull from that?

jessex:
In the medium term, I agree! Secret management is critical for managing credentials once you've reached a certain complexity and size threshold, and it's "merely" important prior to that threshold. In the short term, with our extremely small number of environment variables and even smaller number of credentials to manage (1), I think the added complexity of integrating secret management is not _yet_ warranted.

There are a couple of high-level options, none _trivial_ to integrate with App Engine, when the time comes: Google Cloud KMS and a third-party solution like Vault on top of Kubernetes Engine or Compute Engine. Maybe a migration out of App Engine would accompany the integration of secret management in the future and dovetail nicely.

As for storing it encrypted in Cloud Storage and having the app fetch on startup, we are following fairly well-accepted standards here (read: [12 Factor principles on configuration](https://12factor.net/config)). Configuration like this, which could vary from deploy to deploy and definitely varies between environments, is best read from the environment, with secret management integration on startup as an acceptable way to load secrets.

jbrambleDC:
Sorry, I should have specified better about the app fetching on startup, what might make more sense is to have secrets stored with KMS encryption and a deploy script that sets config via fetching the secrets. 

I agree, that for the time being, configuring manually locally is not a major issue.

jessex:
I think we're on the same page here. Secret and key management solutions that I have seen include or integrate with tools that fetch secrets from some manifest and place them into the environment that the app is running within. For example, Vault integrates with Summon to grab mappings from key to Vault path for all secrets in a `secrets.yml` file. That's effectively what you're calling for and I think in time we will use such an approach.

Issue #90, opened 2018-05-14
Ingest generalization (tracking)


cawarren:
This issue tracks progress on generalization of the ingest (scraping) pipeline. 

List is roughly ordered, with more self-contained and dependency tasks towards the top, and more complex and dependent work towards the bottom.

Work:
- [x] Migrate region subclass fields to superclasses: issue #63 
- [x] Rename top-level 'scraper' package to 'ingest': issue #82
- [x] Update region_manifest with new config info : issue #84
- [x] Refactor scraper setup() call to be us_ny specific, not general scraper interface : issue #85
- [x] Pull out docket as class, add helper methods : issue #86
- [x] Pull out scraper as a class all scrapers inherit from : issue #87
- [x] Improve session management, rename queue_control : issue #83
- [ ] Add metrics: issue #59 
- [x] Clean up ingest test coverage : issue #89

Moved out of scope:
- Create queue manager : issue #88 (it was determined that sticking with a single queue for each scraping region was better for now until we were sure that the complexity was a) needed, b) in this particular architecture)

jessex:
The spirit of this is complete and the outstanding task, #59, is part of a larger ingest productionalization movement. Closing this now!

PR #98, opened 2018-05-23
[WIP]: Conditional release recidivism, i.e. revocations

jessex:
This is a first pass at support for detecting and calculating recidivism from conditional releases, i.e. revocation. The implementation relies on iterating over snapshots that we have for the individual, looking at changes in custody dates, release dates, and facilities to detect if there was a conditional release and/or a return following a conditional release. 

This is still a Work In Progress because I need to add new, thorough test coverage for revocation detection. My implementation tries to be defensive against manual data issues and things like consecutive snapshots changing the same date, which is a partial complexity source and will require ample testing. There is a test coverage decrease coming until that is complete.

The first commit enhances the existing calculation code to support multiple recidivism events for a person in a single release cohort (calendar year). This was incredibly unlikely and maybe non-existent within the NY source data, but will happen, albeit infrequently, once we start tracking revocation.

The second commit adds the core logic for revocation detection. It's fairly complicated and I'm not stoked about it, but I don't think we have a markedly better path at the moment. Please tear this apart! Would love to see that I've over-thought this, and would welcome with less enthusiasm being told that I've under-thought this.

### DIFF
--- a/calculator/recidivism/calculator.py
+++ b/calculator/recidivism/calculator.py
@@ -92,21 +92,22 @@ def map_recidivism_combinations(inmate, recidivism_events):
     metrics = []
     all_reincarceration_dates = reincarceration_dates(recidivism_events)
 
-    for release_cohort, event in recidivism_events.iteritems():
-        characteristic_combos = characteristic_combinations(inmate, event)
+    for release_cohort, events in recidivism_events.iteritems():
+        for event in events:
+            characteristic_combos = characteristic_combinations(inmate, event)
 
-        earliest_recidivism_period = earliest_recidivated_follow_up_period(
-            event.release_date, event.reincarceration_date)
+            earliest_recidivism_period = earliest_recidivated_follow_up_period(
+                event.release_date, event.reincarceration_date)
 
-        relevant_periods = relevant_follow_up_periods(
-            event.release_date, date.today(), FOLLOW_UP_PERIODS)
+            relevant_periods = relevant_follow_up_periods(
+                event.release_date, date.today(), FOLLOW_UP_PERIODS)
 
-        for combo in characteristic_combos:
-            combo["release_cohort"] = release_cohort
+            for combo in characteristic_combos:
+                combo["release_cohort"] = release_cohort
 
-            metrics.extend(combination_metrics(
-                combo, event, all_reincarceration_dates,
-                earliest_recidivism_period, relevant_periods))
+                metrics.extend(combination_metrics(
+                    combo, event, all_reincarceration_dates,
+                    earliest_recidivism_period, relevant_periods))
 
     return metrics
 
@@ -126,9 +127,11 @@ def reincarceration_dates(recidivism_events):
         A list of reincarceration dates, in the order in which they appear in
         the given list of objects.
     """
-    return [event.reincarceration_date
-            for _cohort, event in recidivism_events.iteritems()
-            if event.reincarceration_date]
+    dates = []
+    for _cohort, events in recidivism_events.iteritems():
+        dates.extend([event.reincarceration_date for event in events
+                      if event.reincarceration_date])
+    return dates
 
 
 def count_reincarcerations_in_window(start_date,
--- a/calculator/recidivism/identifier.py
+++ b/calculator/recidivism/identifier.py
@@ -33,10 +33,12 @@ Example:
 
 import logging
 from collections import defaultdict
+from dateutil.relativedelta import relativedelta
 from models.snapshot import Snapshot
 from models.record import Record
-# This is required to permit access to PolyModel attributes on UsNyRecord
+# These are required to permit access to PolyModel attributes on UsNy models
 from scraper.us_ny.us_ny_record import UsNyRecord #pylint: disable=unused-import
+from scraper.us_ny.us_ny_snapshot import UsNySnapshot #pylint: disable=unused-import
 from .recidivism_event import RecidivismEvent
 
 
@@ -122,12 +124,12 @@ def find_recidivism(inmate, include_conditional_violations=False):
                 recidivism_events[release_cohort].append(event)
 
                 if include_conditional_violations:
-                    (conditional_release_cohort, conditional_event) = \
-                        for_conditional_release(record,
-                                                original_entry_date,
-                                                release_facility)
-                    recidivism_events[conditional_release_cohort]\
-                        .append(conditional_event)
+                    conditional_events = for_conditional_release(
+                        record, original_entry_date, original_entry_date,
+                        release_date, release_facility)
+
+                    merge_recidivism_events(recidivism_events,
+                                            conditional_events)
 
     return recidivism_events
 
@@ -153,8 +155,9 @@ def subsequent_entrance(record):
     """The date of most recent entrance into prison for the given record.
 
     A subsequent entrance is when an inmate returns to prison for a given
-    sentence, after having already been conditionally for that same sentence.
-    It is not returning to prison for a brand new sentence.
+    sentence, after having already been released for that same sentence, e.g.
+    onto parole or a conditional release. It is not returning to prison for a
+    brand new sentence.
 
     Args:
         record: a single record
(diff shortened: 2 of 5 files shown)


calculator/recidivism/calculator.py
```diff
@@ -92,21 +92,22 @@ def map_recidivism_combinations(inmate, recidivism_events):
     metrics = []
     all_reincarceration_dates = reincarceration_dates(recidivism_events)
 
-    for release_cohort, event in recidivism_events.iteritems():
-        characteristic_combos = characteristic_combinations(inmate, event)
+    for release_cohort, events in recidivism_events.iteritems():
```
jbrambleDC: do we care about this being python 3 compatible? iteritems() no longer exists in python 3, and items was made into an iterator. If you care, I would suggest using six.iteritems(recidivism_events), so that it ports cross version

calculator/recidivism/calculator.py
```diff
@@ -126,9 +127,11 @@ def reincarceration_dates(recidivism_events):
         A list of reincarceration dates, in the order in which they appear in
         the given list of objects.
     """
-    return [event.reincarceration_date
-            for _cohort, event in recidivism_events.iteritems()
-            if event.reincarceration_date]
+    dates = []
+    for _cohort, events in recidivism_events.iteritems():
```
jbrambleDC: _cohort is unused, you could do:
```python
dates = [event.reincarceration_date for event in recidivism_events.values() if event.reincarceration_date]
```

jbrambleDC:
I left a few comments on this. I haven't finished reviewing as I am packing to move, so feel free to move forward without my input.

calculator/recidivism/calculator.py
jessex: Yes, I would like to avoid code that will need to be changed at eventual upgrade. I'll take a look at Six, thanks!

calculator/recidivism/calculator.py
jessex: So, yes, that's an improved version of the list comprehension I was doing before. However, I've expanded this to be a list comprehension inside of a for loop because there's now nested iteration (list values in a dictionary), and when I composed this as a comprehension inside of a comprehension, I thought it was unnecessarily difficult to read. 

But I could follow the spirit of your comment as such:
```python
for events in recidivism_events.values():
    dates.extend([event.reincarceration_date for event in events if event.reincarceration_date])
```

So using `values` over `iteritems` but not a single comprehension.

calculator/recidivism/identifier.py
```diff
@@ -44,22 +45,31 @@ def find_recidivism(inmate, include_conditional_violations=False):
     recidivism or not.
 
     Transforms each sentence from which the inmate has been released into a
-    mapping from its release cohort to the details of the event. The release
-    cohort is an integer for the year, e.g. 2006. The event details are a
-    RecidivismEvent object, which represents events of both recidivism and
-    non-recidivism. That is, each inmate sentence is transformed into a
-    recidivism event unless it is the most recent sentence and they are still
-    incarcerated.
+    mapping from a release cohort to the details of all event in that cohort.
```
ohinds-vera: typo: "events"

calculator/recidivism/identifier.py
```diff
@@ -44,22 +45,31 @@ def find_recidivism(inmate, include_conditional_violations=False):
...
+    The release cohort is an integer for the year, e.g. 2006. The event details
+    are RecidivismEvent objects, which represents events of both recidivism and
+    non-recidivism. That is, each inmate sentence is transformed into one or
+    more recidivism event unless it is the most recent sentence and they are
```
ohinds-vera: ditto, "events"

calculator/recidivism/identifier.py
```diff
@@ -44,22 +45,31 @@ def find_recidivism(inmate, include_conditional_violations=False):
...
-      2012: RecidivismEvent(recidivated=False, original_entry_date="2010-09-17",
-                            ...)
+      2008:[RecidivismEvent(recidivated=True, original_entry_date="2006-04-05",
+                            ...)],
+      2012:[RecidivismEvent(recidivated=False, original_entry_date="2010-09-17",
+                            ...)]
     }
 
+    The reason the value in the returned dictionary is a list of
+    RecidivismEvents instead of a single event is that, in rare corner cases,
```
ohinds-vera: These might not be rare corner cases in some jurisdictions.

calculator/recidivism/identifier.py
```diff
@@ -306,28 +330,184 @@ def for_intermediate_record(record, recidivism_record, snapshots,
...
+            # Update to the original custody date (unlikely but could be a
+            # manual fix applied in source system)
+            prior_entry_date = snapshot.custody_date
+
+        if snapshot.latest_release_date:
+            # Snapshot introduces new release date or updates previously
+            # introduced release date if it is a manual fix in source data
+            most_recent_release_date = snapshot.latest_release_date
+
+        if snapshot.latest_facility:
```
ohinds-vera: I'm curious why the facility is needed here.

calculator/recidivism/identifier.py
```diff
@@ -306,28 +330,184 @@ def for_intermediate_record(record, recidivism_record, snapshots,
...
+    Because this implementation relies on custody and release date changes
+    captured in successive snapshots, we need enough historical (assuming the
+    historical data does not commit data loss, as New York State's DOCCS does)
+    and/or ongoing data to reach meaningful quantities.
+
+    Additionally, in the case where there is no subsequent custody date after a
+    final detected release date, i.e. apparent non-revocation, we check if the
+    overall time spent incarcerated is less than the max sentence length to
+    detect if this was an actual conditional release and not an unconditional
+    release. This may not be accurate in all jurisdictions.
```
ohinds-vera: Yes, there is a ton of variation in how this is handled. Feds and some states sentence people to mandatory post-sentence supervision that has nothing to do with the length of the prison sentence. Also, it's common to have suspended sentences, where the person goes straight to supervision. How should those revocations be handled?

calculator/recidivism/calculator.py
jbrambleDC: to be fair, you could also accomplish the same with a nested comprehension. Some people believe that to be bad form however, so if you believe this to be more readable, I like that!

jbrambleDC:
Just want to jump in and say its awesome to see @ohinds-vera sprinkling domain knowledge into this PR!

calculator/recidivism/calculator.py
jessex: >when I composed this as a comprehension inside of a comprehension, I thought it was unnecessarily difficult to read

Yep! I tried the nested comprehension route and didn't like the readability.

calculator/recidivism/identifier.py
jessex: That's fair. I'll remove the hemming and hawing on that front and keep it simple. Thanks!

calculator/recidivism/identifier.py
```diff
@@ -234,13 +237,32 @@ def last_facility(record, snapshots):
...
-    return True
+def incarcerated_multiple_times(record):
+    """Returns whether or not the individual was incarcerated multiple distinct
+    times for this given record.
+
+    A person is incarcerated multiple times for the same record if they are
+    released to some form of supervision or with some conditions, and then
+    violate the terms of the release and are reincarcerated.
+
+    This works by checking if the original entry date for the record is equal to
```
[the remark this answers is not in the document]
jessex: Wow, yes, great scenario @ohinds-vera. We'll protect against pre-sentence snapshots here, and possibly elsewhere in calculation. This gets back to the conversation we had at the last weekly about how we'll end up accounting for jail versus prison in snapshots and records. That will propagate here, but I don't think we have anything we can do about it right now.

calculator/recidivism/identifier.py
jessex: This is to make sure that if the `latest_facility` changed in a given snapshot, that we capture that change in our accumulation.

calculator/recidivism/identifier.py
jessex: At a high level, how I would like to handle this going forward is by clearly, explicitly defining the standard model/contract for what a `Record` is, and compelling individual scrapers to ensure that they adapt the incoming source data to that contract. In this case, that would mean some required field or set of fields that determine whether a given release was conditional, clearly documenting that field, and ensuring each scraper is filling in each required field. So the logic for setting that field would differ in jurisdictions like, say, New York and Vermont. That standard model/contract will become more clear as we progress into further regions.

In this particular example of suspended sentences, that's a great question. Practically speaking, would we expect folks with suspended sentences to show up in publicly scraped inmate location system data? Regardless, I would think that if we have a record that indicates that you went straight to supervision and then had it revoked, entering prison for the first time, we may want to do two things:
1. Count it as a conditional recidivism instances (i.e. treat it like any other revocation as we are here)
2. Provide a finer grained dimension in recidivism (and population?) calculations corresponding to specific release type, wherein we could distinctly count, say, parole revocation from suspended-sentence-revocation from probation returns from violation of merit release terms, etc.

We are seeking to take a pass at 1) here but are not making any gestures towards 2) yet.

jessex:
I ran this successfully locally, encountering a handful of as of yet untested bugs along the way. As noted in our meeting, I'm putting this on hold until the migration is finished, while I pivot to focus on the generalization work as part of #90. I'll return to this to fill out the tests and fulfill feedback later.

Issue #88, opened 2018-05-14
Create queue manager


cawarren:
_This is part of ingest generalization work_

**Background:**
To date, each scraper has been expected to get its own scraping (request throttling) queue. This was partly a matter of convenience while getting set-up, and partly because state and federal prison systems are large enough to keep a queue occupied around the clock.

As we look at how to scale this system to jails as well as prisons, this is no longer viable - there are ~3,000 county jails in the US, and each would take less than 5min to scrape.

As a result, we're switching to a new model. While some regions will continue to have special, dedicated queues (e.g., `us_ny`), others will lease queues which are pre-configured for the specific rate they need (e.g., 15 queries per minute). The list of acceptable queue / queue rates for a region will be listed in the `region_manifest` (see issue #84). Many regions may have multiple acceptable rates (e.g., 15/m and 20/m).

This requires a new part of the system, a `queue_manager` module.

**Expected workflow:**
The `queue_manager` is expected to work something like this:
- When scraper starts, `scraper_control` (`queue_control`, if issue #83 isn't complete yet) submits the region and requested lease period to the `queue_manager` queue (this may be a queue, or just a FIFO list on a datastore object - see _Scope of work_ below)
- At some interval (or when poked by `scraper_control`), `queue_manager` checks whether any of the queues it manages are free, and if so attempts to retrieve the next task for that queue type from its own `queue_manager` queue
- If there is such a task, it calls `scraper_control` with the region and new queue name that has been leased, which kicks off that scrape. `queue_manager` also creates a deferred task for itself, for the time of lease expiration, to wake back up and evict the scraper if it's still using that queue.
- When scraper finishes or stops, it calls `scraper_control.stop_scraper()` on itself, which calls `queue_manager` to terminate the lease and purge the queue (and check if others are waiting on it, to re-lease the queue).

_Note: GAE recommends not adding new tasks to a queue for up to 1m after purging it - use the [GAE deferred library](https://cloud.google.com/appengine/articles/deferred) to add delays without exceeding request deadlines._

For more details, see the design doc's [Life of a Scraper](https://docs.google.com/document/d/11fv-hLwLOFcM5XSE16lYne5hnqFzcQK1vwvTBVVyhn8/edit#heading=h.qz0bdjml50ah) and [Queues](https://docs.google.com/document/d/11fv-hLwLOFcM5XSE16lYne5hnqFzcQK1vwvTBVVyhn8/edit#heading=h.uwhtuqwgoo6) sections.

[...]

PR #101, opened 2018-05-31
Updates region manifest

jessex:
This PR is ready to go, but it is branched off of the `rename-scraper-ingest` branch until that is merged. Once that is merged, I'll reissue this against master.

This does a few things:
* Adds a few new typed fields: `timezone`, `scraper_package`, and `scraper_class`
* Changes `scraper_queue_name` to an array called `queues` to permit regions that can be run on multiple queues, e.g. throttled queues
* Refactors the dynamic scraper loading methods in `regions.py` to make use of the new `scraper_*` fields, which will be necessary when we have regions for jails which use region-agnostic vendor scraping code
* Adds support for dynamic key-value pairs for regions that need them (@ohinds-vera: check out the second commit specifically and let me know if that would suit your use case)

### DIFF
--- a/ingest/us_ny/us_ny_scraper.py
+++ b/ingest/us_ny/us_ny_scraper.py
@@ -143,7 +143,7 @@ def start_scrape(scrape_type):
     serial_params = json.dumps(params)
 
     taskqueue.add(url=SCRAPER_WORK_URL,
-                  queue_name=REGION.scrape_queue_name,
+                  queue_name=REGION.queues[0],
                   params={'region': REGION.region_code,
                           'task': "scrape_search_page",
                           'params': serial_params})
@@ -215,7 +215,7 @@ def stop_scrape(scrape_types):
             "type: %s." % str(scrape))
         deferred.defer(resume_scrape, scrape, _countdown=60)
 
-    q = taskqueue.Queue(REGION.scrape_queue_name)
+    q = taskqueue.Queue(REGION.queues[0])
     q.purge()
 
 
@@ -275,7 +275,7 @@ def resume_scrape(scrape_type):
     serial_params = json.dumps(params)
 
     taskqueue.add(url=SCRAPER_WORK_URL,
-                  queue_name=REGION.scrape_queue_name,
+                  queue_name=REGION.queues[0],
                   params={'region': REGION.region_code,
                           'task': "scrape_search_page",
                           'params': serial_params})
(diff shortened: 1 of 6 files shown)


utils/regions.py
```diff
@@ -132,40 +145,87 @@ def get_subkind(region_code, parent_kind_name):
...
+    logging.debug('Fetching scraper package and class from cache '
+                  'for region %s', region_code)
+
+    memcache_package_name = region_code + "_scraper_package"
+    scraper_package = memcache.get(memcache_package_name)
+
+    memcache_class_name = region_code + "_scraper_class"
+    scraper_class = memcache.get(memcache_class_name)
+
+    if not scraper_package and not scraper_class:
```
cawarren: These values could fall out of memcache at separate times if they're pushed out by memory constraints, in which case this `if` statement won't run.

Suggest changing this to:
 `if not scraper_package or not scraper_class:`

ingest/us_ny/us_ny_scraper.py
```diff
@@ -143,7 +143,7 @@ def start_scrape(scrape_type):
     serial_params = json.dumps(params)
 
     taskqueue.add(url=SCRAPER_WORK_URL,
-                  queue_name=REGION.scrape_queue_name,
+                  queue_name=REGION.queues[0],
```
cawarren: This is temporary until we get queue leasing in place, right? Otherwise maybe I'm misunderstanding.

ingest/worker.py
```diff
@@ -20,6 +20,7 @@
 
 import logging
 import webapp2
+from google.appengine.api import memcache
```
cawarren: Not used in worker.py

ingest/us_ny/us_ny_scraper.py
jessex: Is the NY system going to use multiple queues, or just the NY queue at all times? I was updating the `us_ny_scraper.py` to maintain its current logic, but if this functionality in particular gets extracted out into `scraper.py` then it will need to be updated to handle choosing between multiple queues, which will come in #88.

ingest/worker.py
jessex: Whoops not sure how that got there. Will remove.

utils/regions.py
jessex: Good call! Will change.

ingest/us_ny/us_ny_scraper.py
cawarren: You're right that it will use the NY queue at all times, it should just get routed there by the `queue_manager` in #88 rather than selecting the queue itself (same pattern as every other scraper). 

So this makes sense for now, we'll just update it slightly once #88 is ready.

jessex: I'll put a TODO referencing that issue on this for now.

jessex:
This is being replaced by #102, so I am closing it. All feedback will be handled on that PR, and all new conversation should happen there, unless you're interested in a particular thread and its context here before it gets resolved (hopefully tomorrow June 7th).

Issue #32, opened 2018-01-31
Implement sentence breakdowns


jessex:
Enhance the calculation pipeline to include sentencing breakdowns as described in our measurement paper. This includes breakdown by actual sentence length, maximum sentence length, maximum supervision length, release type, parole denial, and total historical sentences.

arian487:
Closing in favour of #400

jessex:
I think this was an errant close. This is related to recidivism calculation, not to the database schema.

PR #108, opened 2018-07-08
Adds facility stay length

bpacker:
This is the first part of #32.

This simply adds the length of prison stay (in months) to the calculated metrics associated with a given recidivism event, bucketing according to the following intervals: <12, 12-24, 24-36, 36-48, 48-60, 60-72, 72-84, 84-96, 96-108, 108-120, 120+. This follows the specification under the Stay Breakdown section of the "Measuring Recidivism" doc.


calculator/recidivism/calculator.py
```diff
@@ -266,6 +266,61 @@ def age_bucket(age):
...
+
+    Returns:
+	The length of the prison stay in months. None if the original entry date or
+        release date is not known.
+    """
+    if event.original_entry_date is None or event.release_date is None:
+        return None
+
+    delta = relativedelta(event.release_date, event.original_entry_date)
+    return delta.years * 12 + delta.months
```
bpacker: This metric is calculated in months, following the Measuring Recidivism doc, but since the current buckets are all at year boundaries, this could be changed to years if that's simpler.

calculator/recidivism/calculator.py
```diff
@@ -266,6 +266,61 @@ def age_bucket(age):
...
+
+    Args:
+        event: the event
+
+    Returns:
+	The length of the prison stay in months. None if the original entry date or
+        release date is not known.
+    """
+    if event.original_entry_date is None or event.release_date is None:
+        return None
```
bpacker: Is this handling of missing entries necessary? Not sure what the story is in general in terms of missing data or validation.

calculator/recidivism/calculator.py
```diff
@@ -278,22 +333,19 @@ def characteristic_combinations(inmate, event):
...
-        {"race": "black", "sex": "female", "age": "<25"}) =
-            [{},
-            {'age': '<25'}, {'race': 'black'}, {'sex': 'female'},
-            {'age': '<25', 'race': 'black'}, {'age': '<25', 'sex': 'female'},
-            {'race': 'black', 'sex': 'female'},
-            {'age': '<25', 'race': 'black', 'sex': 'female'}]
+    The output for a black female age 24 and an incarceration that began in
+    January 2008 and ended in February 2009 is equal to the output of:
+            for_characteristics({'age': '<25', 'race': 'black',
+                                 'sex': 'female', 'stay_length': '12-24'})
```
bpacker: Note: this comment change wasn't related to the scope of this PR, it was simply to better describe the output of this method (and the original comment was moved down to the for_characteristics method).

tests/calculator/recidivism/pipeline_test.py
```diff
@@ -105,21 +106,30 @@ def test_map_inmate(self):
                     else:
                         assert value == 1
 
-        # 16 combinations * 2 methodologies * 10 periods = 320 combinations
+        # Get the number of combinations of inmate-event characteristics.
+        original_entry_date = date(2013, 6, 17)
+        event = recidivism_event.RecidivismEvent(False, original_entry_date,
+                                                 None, "Sing Sing")
+        num_combinations = len(calculator.characteristic_combinations(
```
bpacker: Doing this is necessary if we don't want to manually change these numbers every time the metrics change.

Another possibility is to mock pipeline's calls to calculator, which breaks the dependency on calculator entirely. But currently this test serves as more of an end-to-end test, which might have been the intent.

calculator/recidivism/calculator.py
```diff
@@ -266,6 +266,61 @@ def age_bucket(age):
...
 
+def stay_length_from_event(event):
+    """Length of prison stay of a given event.
+
+    Args:
+        event: the event
+
+    Returns:
+	The length of the prison stay in months. None if the original entry date or
+        release date is not known.
```
jessex: This is indented one space too far 🦅

calculator/recidivism/calculator.py
jessex: Let's keep it as months for now. We may adjust these boundaries in the near future--or make them configurable--but I think months is a reasonable unit for a while.

calculator/recidivism/calculator.py
```diff
@@ -278,22 +333,19 @@ def characteristic_combinations(inmate, event):
     For each event, we need to calculate metrics across combinations of:
     Release Cohort; Follow-up Period (up to 10 years);
     Methodology (Event-based, Offender-based);
-    Demographics (age, race, sex); Location (prison, region); ...
+    Demographics (age, race, sex); Location (prison, region);
```
jessex: Would you mind changing this "prison" reference to "facility" while you're here? We're going to generalize.

jessex: Pursuant to this, can we change your new comment below to "Facility Stay Breakdown?"

calculator/recidivism/calculator.py
```diff
@@ -317,6 +372,16 @@ def characteristic_combinations(inmate, event):
 def for_characteristics(characteristics):
     """The list of all combinations of the given metric characteristics.
 
+    Example:
```
jessex: Thanks, this documentation switcharoo is a good call.

tests/calculator/recidivism/pipeline_test.py
jessex: That was the intent, yes. Wiring up a full integration test with MapReduce actually running is a more long-term goal, but it's fairly tricky to pull off and a migration to something like Cloud Dataflow may not be so far off, so I thought this was a fair compromise for now.

I think what you've done here is reasonable.

tests/calculator/recidivism/calculator_test.py
```diff
@@ -245,6 +245,95 @@ def test_age_bucket():
...
+    assert calculator.stay_length_from_event(event) == 9
+
+
+def test_stay_length_from_event_same_month_earlier_date():
+    original_entry_date = date(2013, 6, 17)
+    release_date = date(2014, 6, 16)
+    event = recidivism_event.RecidivismEvent(
+        False, original_entry_date, release_date, "Sing Sing")
+
+    assert calculator.stay_length_from_event(event) == 11
```
jessex: Talking myself through this: I see that we are not incrementing the month total until we have actually hit the same date of the month, i.e. the anniversary. So even though this is closer to 12 months than 11 in absolute terms, it is calculated as 11 and then bucketed as '<12'. Similarly, if the release date had a year of 2015 instead, then this would get calculated as 23 and bucketed as '12-24'. This is the behavior we want because the upper bound in the bucketing is exclusive. 

Okay, cool, sounds great.

calculator/recidivism/calculator.py
jessex: Thanks, this documentation switcharoo is a good call.

calculator/recidivism/calculator.py
jessex: It shouldn't be necessary, but the defensiveness can't hurt given that we don't have an airtight validation story in place yet. Take a look at `identifier.py`, starting line 85. We don't make the call that creates the `RecidivismEvent` you are operating on here if one of the two dates is missing. But I'd keep it nonetheless.

calculator/recidivism/calculator.py
bpacker: I do not understand this emoji but I made the change anyway.

calculator/recidivism/calculator.py
bpacker: Changed "prison" to "facility" here and everywhere else I used it.

tests/calculator/recidivism/calculator_test.py
bpacker: Added an explanation along these lines to stay_length_from_event docstring.

bpacker:
Addressed comments and pushed changes.

bpacker:
No I figured we'd just test in prod.

Jk, here it is showing up in the datastore:
image

About to 🍆

Issue #89, opened 2018-05-14
Clean up ingest test coverage


cawarren:
_This is part of ingest generalization work_

This is a cleanup task, intended to occur after the major changes involved in ingest generalization (otherwise, much of what gets new tests will be removed / majorly renovated, and require new tests again).

After all other generalization work is done (it can be tracked in issue #90), add test coverage for remaining uncovered components of the `ingest` system (including `us_ny` scraper).

jessex:
You know, as engineering practice, I think it would be better, albeit more difficult now, to try to fill out test coverage as we go. With issues that extract logic from one place into another, or that create an interface where there is today only implementation, or that add new functionality--more or less all of the issues we've filed for the generalization work--having test coverage from the jump reduces the likelihood of bugs in a number of ways: shorter feedback loop, less context switching, less context loss. 

As long as we are testing only in scope of the issue being worked on, I don't think we'll have much test churn. And if there _is_ test churn, you could argue that it's even more important to have incremental test coverage improvement from issue to issue, as a sort of historical record of the impact of each commit.

cawarren:
Sorry, I should have been clearer - most of the issues mentioned in the tracking bug have an explicit item 'Add test coverage', for the ones where they're not making a one-line change.

I think we're on the same page here - all of those should build test coverage of the portion their work pertains to. This bug is only for clean-up (that is, after all of those are done, see where we still don't have coverage and go add it there).

jessex:
Ah, perfect. Same page. Thanks for filing all of these so diligently, @cawarren!

jessex:
This was closed by #134 and also #148.

PR #109, opened 2018-07-17
Session Extraction

### LINKED ISSUE #83: Improve session management, rename queue_control
_This is part of ingest generalization work_

Background:
We're about to introduce a `queue_manager` module as part of the generalization work, so the current naming of `queue_control` is likely to be confusing.

We also need sessions to be more centrally managed, rather than allowing scrapers to create / clean up their own sessions as `us_ny` does today. Sessions should be created by `scraper_control` on `start_scrape` calls, and ended by `scraper_control` on `stop_scrape` calls. Scrapers should only be responsible for updating the session entity they’ve been given (e.g., to store the new `last_scraped` value as they continue scraping). See the Scraper and Sessions sections in the design doc.

Scope of work:
- Rename `queue_control` to `scraper_control` to better reflect its current use
- Pull out the current `scraper/models/scrape_session.py` module to be `ingest/session.py` (see Modules section in design doc) -- _note: if issue #82 isn't complete yet, this will be `scraper/session.py`_ 
- Add `create_session`, `end_session`, `get_last_session`, `update_session` handlers to the `ingest/sessions.py` module
- Update `scraper_control` to use the new helper functions to create and delete sessions when scraping is started and stopped, and to get / pass to the scraper the last active session when it's resuming.
- Add relevant test coverage

### LINKED ISSUE #85: Refactor scraper setup() call to be us_ny specific, not general scraper interface
_This is part of ingest generalization work_

Background:
Some scrapers may require certain work to be done before starting or resuming a scraping session, e.g. fetching session cookies from the site that's being scraped.

Currently, this is an expected scraper interface and `queue_control` (soon to be `scraper_control`) calls it directly before starting/resuming a scraper. Since this won't apply to all scrapers, however, and since its logic is scraper-specific regardless, this should not be a required scraper interface - it should be something local to each scraper, if needed at all.

Scope of work:
- Remove references to `scraper.setup()` in `queue_control` (/`scraper_control`)
- Update `us_ny` scraper to call its own `setup()` method from its `start` or `resume` functions, rather than expecting this to be called elsewhere.

jessex:
This resolves #83.

The commits are self-contained and follow the bullet points outlined in the issue description, more or less in order. There is more code coverage I'd like to add to the request handlers in `scraper_control` but it will be a bit laborious to pull off without actually invoking a live scraper so I am submitting the PR now since it should be ready for review otherwise.

One point of review is whether I have extracted enough logic out of `us_ny_scraper`'s start, stop, and resume calls into `scraper_control`, or whether I should pull more out now. I am confident that much of the logic currently in there is going to become general-purpose for most or all scrapers, but without an actual second scraper to prove that, there is a tightrope to walk between forward-thinking and YAGNI.

Note: This also resolves #85 by getting rid of the `setup` call from the to-be-implemented scraper interface, as I found it was wholly extractable into `scraper_control` and `sessions`.


jessex:
Verified that session behavior is functioning as expected in a variety of start-stop-resume sequences. There are more thorough integration tests to be written here but I'm going to defer those to a later issue, such as #89, in hopes of accelerating this a bit. Merging.

Issue #110, opened 2018-08-03
Refactor directory structure to create a top-level `recidiviz` package


ohinds-vera:
This will mean moving calculator, ingest, models, and, utils into the new top-level directory called recidiviz, and creating an `__init__.py` in that directory. This will allow unambiguous imports across the whole codebase, while maintaining the ability run ingest and calculation independently.

PR #111, opened 2018-08-06
Refactor into single `recidiviz` package (#110)

ohinds-vera:
This creates a top-level package called `recidiviz`. Lots of code moves, but most of the actual code changes are just modifying imports to reflect the new package structure. 

* Moves all code from calculator, ingest, models, tests, and utils
  into a new top-level directory called `recidiviz`.

* Creates an `__init__.py` in that top-level directory.

* Modify the yaml config to load the modules from the `recidiviz`
  package.

* Modify all relevant imports in the code to reference the new
  top-level package.

* Update the documentation to reflect the new top-level package where
  appropriate.

* Update the pylint command in the travis config to reflect the refactor.

* Modify the pytest config to point to the correct directory.


.travis.yml
```diff
@@ -4,7 +4,8 @@ python:
 install:
 - pip install -r requirements.txt
 script:
-- pylint *.py calculator models tests utils
+# TODO: this should be "pylint *.py recidiviz" after ingest gets up to lint.
```
jessex: :+1:

Issue #87, opened 2018-05-14
Pull out scraper as a class all scrapers inherit from


cawarren:
_This is part of ingest generalization work_

Background:
Currently, the `us_ny` scraper is standalone - it doesn't inherit anything, and there's no common interface defined for scrapers to be used by the rest of the ingest system.

The design doc calls for a common Scraper superclass to be used to ensure consistent implementation and interfaces for scrapers, with stubs for those interfaces that must be implemented by inheritors.

From Josh's comment on issue #55 :

> Another technical point: we should likely be making use of Python Abstract Base Classes here. Check out the [PEP](https://www.python.org/dev/peps/pep-3119/) and the [module docs](https://docs.python.org/2/library/abc.html). I could see an interface around individual region scrapers to the effect of start, stop, and resume commands, and/or more fine-grained interfacing around navigation and parse functionality. We'll need to feel it out, but having a discrete "interface" for region scrapers to extend will provide much more in the way of structure and safety than our introspection-based model we have right now.

Scope of work:
- Read the documents described in issue #55, and the design doc [Scraper section](https://docs.google.com/document/d/11fv-hLwLOFcM5XSE16lYne5hnqFzcQK1vwvTBVVyhn8/edit#heading=h.90ynq2mmh781)
- Create a new `scraper` module under the `ingest` package (if issue #82 isn't finished yet, the package might still be called `scraper`) - see design doc [Modules section](https://docs.google.com/document/d/11fv-hLwLOFcM5XSE16lYne5hnqFzcQK1vwvTBVVyhn8/edit#heading=h.v6780khuazyt)
- In the new `scraper` module, create a scraper class that future scrapers will implement, with stubbed methods for start/stop/resume
- Update the `us_ny` scraper to use / inherit from the new `scraper` module
- Add relevant tests

ohinds-vera:
@cawarren @jessex 

As I get around to the task spawning part of this work, I have some design decisions that it seems worthwhile to talk through here.  

Previously, all state was passed through params from one task function to the next. Now, we have an object where we could store this state. However, we don't have the option of traditional scoping, so we'd probably have to play some trick where we rebuild the object at the beginning of each task and pass around a pickled copy. This would have some advantages from a programming standpoint in that handling the data passed from task to task would be through a familiar mechanism (member variables). 

[...]

PR #112, opened 2018-08-07
Create a Scraper class for regional scrapers to inherit from (#87)

ohinds-vera:
This PR addresses issue #87 by adding a Scraper class, and modifying UsNyScraper to inherit from it.

* Add ingest/scraper.py that defines the Scraper class.

* Remove the functionality from UsNyScraper that was pulled into
  Scraper.

* Lint-ify UsNyScraper and small things in the rest of ingest, then
  modify the travis config to include ingest.

* Modify the regions utility to return a region-specific Scraper
  object.

* Modify the worker to use the scraper object.

* Modify the region tests to reflect that getting a scraper gets an
  object.

### DIFF
--- a/recidiviz/ingest/scraper.py
+++ b/recidiviz/ingest/scraper.py
@@ -68,7 +68,7 @@ class Scraper(object):
         Args: (any type) the inmate ID to transform
 
         Returns:
-            N/A
+            A record id mapped from the given inmate id.
 
         """
         pass
@@ -80,7 +80,7 @@ class Scraper(object):
         task to run in the scraper.
 
         Returns:
-            The name of the function to run as the first task
+            The name of the function to run as the first task.
 
         """
         pass
--- a/recidiviz/ingest/scraper.py
+++ b/recidiviz/ingest/scraper.py
@@ -21,29 +21,23 @@ inherit from.
 """
 
 import abc
-from datetime import date
 import json
 import logging
-import string
-import random
 
-import dateutil.parser as parser
 import requests
 
 from google.appengine.ext import deferred
-from google.appengine.ext import ndb
 from google.appengine.api import taskqueue
-from recidiviz.models import env_vars
+from recidiviz.ingest import scraper_utils
 from recidiviz.ingest import sessions
 from recidiviz.ingest.models.scrape_key import ScrapeKey
 from recidiviz.ingest import tracker
-from recidiviz.utils import environment
 from recidiviz.utils.regions import Region
 
 
 class Scraper(object):
     """The base for all scraper objects. It handles basic setup, scrape
-    process control (start, pause, resume, stop), web requests, task
+    process control (start, resume, stop), web requests, task
     queueing, state tracking, and a bunch of static convenience
     methods for data manipulation.
 
@@ -56,13 +50,24 @@ class Scraper(object):
 
         self.REGION = Region(region_name)
         self.FAIL_COUNTER = (
-            self.REGION.region_code + "_next_page_fail_counter")
+            self.get_region().region_code + "_next_page_fail_counter")
         self.SCRAPER_WORK_URL = '/scraper/work'
 
 
     @abc.abstractmethod
     def inmate_id_to_record_id(self, inmate_id):
         pass
+
+
+    @abc.abstractmethod
+    def get_initial_task(self):
+        pass
+
+
+    def get_region(self):
+        """Retrieve the region object associated with this scraper.
+        """
+        return self.REGION
 
 
     def start_scrape(self, scrape_type):
@@ -80,15 +85,15 @@ class Scraper(object):
         """
         docket_item = self.iterate_docket_item(scrape_type)
         if not docket_item:
-            logging.error("Found no %s docket items for %s, shutting down." % (
-                scrape_type, self.REGION.region_code))
-            sessions.end_session(ScrapeKey(self.REGION.region_code,
+            logging.error("Found no %s docket items for %s, shutting down.",
+                          scrape_type, self.get_region().region_code)
+            sessions.end_session(ScrapeKey(self.get_region().region_code,
                                            scrape_type))
             return
 
         params = {"scrape_type": scrape_type, "content": docket_item}
 
-        self.add_task(self.INITIAL_TASK, params)
+        self.add_task(self.get_initial_task(), params)
 
 
     def stop_scrape(self, scrape_types):
@@ -116,17 +121,18 @@ class Scraper(object):
         # resume for them since the taskqueue purge below will kill them.
         other_scrapes = set([])
 
-        open_sessions = sessions.get_open_sessions(self.REGION.region_code)
+        open_sessions = sessions.get_open_sessions(
+            self.get_region().region_code)
         for session in open_sessions:
             if session.scrape_type not in scrape_types:
                 other_scrapes.add(session.scrape_type)
 
         for scrape in other_scrapes:
             logging.info("Setting 60s deferred task to resume unaffected "
-                         "scrape type: %s." % str(scrape))
+                         "scrape type: %s.", str(scrape))
             deferred.defer(self.resume_scrape, scrape, _countdown=60)
 
-        q = taskqueue.Queue(self.REGION.queues[0])
+        q = taskqueue.Queue(self.get_region().queues[0])
         q.purge()
 
 
@@ -144,7 +150,7 @@ class Scraper(object):
             N/A
         """
         recent_sessions = sessions.get_recent_sessions(ScrapeKey(
-            self.REGION.region_code, scrape_type))
+            self.get_region().region_code, scrape_type))
 
         if scrape_type == "background":
             # Background scrape
@@ -181,11 +187,11 @@ class Scraper(object):
             content = self.iterate_docket_item(scrape_type)
             if not content:
                 sessions.end_session(
-                    ScrapeKey(self.REGION.region_code, scrape_type))
+                    ScrapeKey(self.get_region().region_code, scrape_type))
                 return
 
         params = {'scrape_type': scrape_type, 'content': content}
-        self.add_task(self.INITIAL_TASK, params)
+        self.add_task(self.get_initial_task(), params)
 
 
     def fetch_page(self, url, data=None):
@@ -194,15 +200,15 @@ class Scraper(object):
         to use as POST data in a POST request to the url.
 
         Args:
-            url: URL to fetch content from
-            data: POST data to send
+            url: (string) URL to fetch content from
+            data: (string) POST data to send (optional, default None)
 
         Returns:
             The content if successful, -1 if fails.
 
         """
-        proxies = self.get_proxies()
-        headers = self.get_headers()
+        proxies = scraper_utils.get_proxies()
+        headers = scraper_utils.get_headers()
 
         try:
             if data is None:
@@ -233,7 +239,7 @@ class Scraper(object):
                     ce.response.text)
 
             logging.warning("Problem retrieving page, failing task to "
-                            "retry. \n\n%s" % log_error)
+                            "retry. \n\n%s", log_error)
             return -1
 
         return page
@@ -241,6 +247,15 @@ class Scraper(object):
 
     def add_task(self, task_name, params):
         """ Add a task to the task queue.
+
+        Args:
+            task_name: (string) name of the function in the scraper class to
+                       be invoked
+            params: (dict) parameters to be passed to the function
+
+        Returns:
+            The content if successful, -1 if fails.
+
         """
 
         params_serial = json.dumps(params)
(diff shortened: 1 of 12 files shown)


recidiviz/ingest/scraper.py
```diff
@@ -0,0 +1,479 @@
...
+        Args:
+            scrape_type: (string) The type of scrape to start
+
+        Returns:
+            N/A
+
+        """
+        docket_item = self.iterate_docket_item(scrape_type)
+        if not docket_item:
+            logging.error("Found no %s docket items for %s, shutting down." % (
```
jessex: If you take a look at the `calculator` package, you'll see what we later found out was the Pythonic way to inject arguments into logging statements:
`logging.error("Found no %s docket items for %s, shutting down.", scrape_type, self.REGION.region_code)`

recidiviz/ingest/scraper.py
```diff
@@ -0,0 +1,479 @@
...
+        self.add_task(self.INITIAL_TASK, params)
+
+
+    def fetch_page(self, url, data=None):
+        """Fetch content from a URL. If data is None (the default), we perform
+        a GET for the page. If the data is set, it must be a dict of parameters
+        to use as POST data in a POST request to the url.
+
+        Args:
+            url: URL to fetch content from
```
jessex: We should be consistent here and put the type of the arg in parentheses like in other `Args` lines in the class.

recidiviz/ingest/scraper.py
jessex: We can apply that generally to all of the logging calls throughout the PR.

recidiviz/ingest/scraper.py
```diff
@@ -0,0 +1,479 @@
...
+
+            logging.warning("Problem retrieving page, failing task to "
+                            "retry. \n\n%s" % log_error)
+            return -1
+
+        return page
+
+
+    def add_task(self, task_name, params):
+        """ Add a task to the task queue.
```
jessex: Let's fill this out with `Args` and `Returns` before we merge.

recidiviz/ingest/scraper.py
```diff
@@ -0,0 +1,479 @@
...
+            If successful:
+                Background scrape: ("surname", "given names")
+                Snapshot scrape:   ("record_id", ["records to ignore", ...])
+
+        """
+
+        item_content = tracker.iterate_docket_item(
+            ScrapeKey(self.REGION.region_code, scrape_type))
+
+        if not item_content:
```
jessex: Does the upstream failure that leads to this result get adequately logged/instrumented there? I wonder if we should log something here regardless, especially if we consider `False` to mean "failure."

recidiviz/ingest/scraper.py
```diff
@@ -0,0 +1,479 @@
...
+                              % str(item_content))
+                return False
+
+            return (record_id, item_content[1])
+
+        return item_content
+
+
+    @staticmethod
+    def parse_date_string(date_string, inmate_id):
```
jessex: I wonder if it would be cleaner to extract these static utility functions out to a utility file, e.g. `scraper_utils`? By including them in the base scraper class, we have a few downsides:
- We will inevitably keep coming up with utility functions and they'll keep being added here, which makes this crucial interface harder to read.
- We blur the lines between interface and implementation.
- We are mixing and matching in this class between inheritance and composition. By extracting these static functions out to a helper class, we are more solidly coalescing around composition for shared functionality, which is a good thing.

recidiviz/ingest/scraper.py
```diff
@@ -0,0 +1,479 @@
...
+from recidiviz.ingest import sessions
+from recidiviz.ingest.models.scrape_key import ScrapeKey
+from recidiviz.ingest import tracker
+from recidiviz.utils import environment
+from recidiviz.utils.regions import Region
+
+
+class Scraper(object):
+    """The base for all scraper objects. It handles basic setup, scrape
+    process control (start, pause, resume, stop), web requests, task
```
jessex: I would remove the reference to "pause" here, since there's no technical "pause" functionality that is distinct from "stop" in our control terminology, for now.

recidiviz/ingest/us_ny/us_ny_scraper.py
```diff
@@ -86,1600 +78,1224 @@
     requests.packages.urllib3.contrib.appengine.AppEnginePlatformWarning
 )
 
-# TODO(andrew) - Create class to surround this with, with requirements for
```
jessex: 🙌

recidiviz/ingest/us_ny/us_ny_scraper.py
```diff
@@ -86,1600 +78,1224 @@
...
-    different times of day.
+        try:
+            # Extract said tokens and construct the URL for searches
+            session_K01 = search_tree.cssselect('[name=K01]')[0].get("value")
+            session_token = search_tree.cssselect(
+                '[name=DFH_STATE_TOKEN]')[0].get("value")
+            session_action = search_tree.xpath(
+                '//div[@id="content"]/form/@action')[0]
+        except IndexError:
+            logging.error("Search page not understood. HTML:\n\n%s" %
```
jessex: Let's fix the logging statements in here as I described elsewhere, to be thorough.

recidiviz/ingest/us_ny/us_ny_scraper.py
```diff
@@ -86,1600 +78,1224 @@
...
-            content: (tuple) Search target, stored as a tuple:
-                Background scrape: ("surname", "given names")
-                Snapshot scrape:   ("record_id", ["records to ignore", ...])
-
-    Returns:
-        Nothing if successful, -1 if fails.
-    """
-
-    params = json.loads(params)
+        return None
```
jessex: A return strategy of `None` if successful is kind of odd, more so when failure is an integer value. If we have nothing to return in the case of success but we might have some sort of failure to broadcast, I would probably throw an exception and document in that in the method signature so callers can handle it. Alternatively, if there's a reason not to do that, I would update this to return negative values in the case of failure and positive values in the case of success.

jessex: (I know that for some of my comments in here, I'm commenting on logic that predates your work. But while we're in here cleaning everything up, it's a good idea for us to leave things in a better state!)

recidiviz/ingest/us_ny/us_ny_scraper.py
```diff
@@ -86,1600 +78,1224 @@
...
+                token: (string) DFH_STATE_TOKEN - session info from the form
+                action: (string) The 'action' attr from the scraped
+                    form - URL to POST our request to
+                scrape_type: (string) Type of scrape to perform
+                content: (tuple) Search target, stored as a tuple:
+                    Background scrape: ("surname", "given names")
+                    Snapshot scrape:   ("record_id", ["records to ignore", ...])
+
+        Returns:
+            Nothing if successful, -1 if fails.
```
jessex: Looking around, I see a few conditionals doing `if x == -1`. Looking at their contexts, they all seem to be propagating that value by returning their own -1, which percolates all the way up to `worker.py`. I'm not entirely sure if the GAE taskqueue system itself is responding to negative values in a particular way. 

@cawarren: are the negative values to signal failure your own mechanism, or is that based on something in the taskqueue API? I'm willing to believe that it makes good sense here as a means of control flow signaling, but it's not what I would first think of as my ideal system.

If it's not required for taskqueue and the buck _is_ stopping with propagation to `worker.py`, then instead of muddying up this ticket with a refactor of how we signal errors in ingest, I would file a separate issue to _evaluate_ something exception-based as a replacement.

recidiviz/ingest/us_ny/us_ny_scraper.py
```diff
@@ -86,1600 +78,1224 @@
...
-            "type: %s." % str(scrape))
-        deferred.defer(resume_scrape, scrape, _countdown=60)
-
-    q = taskqueue.Queue(REGION.queues[0])
-    q.purge()
+        Returns:
+            Nothing if successful, -1 if fails.
 
+        """
+        search_page = self.fetch_page(self.REGION.base_url)
```
jessex: Similarly to my statement on `INITIAL_TASK` but in the opposite direction, I would consider adding a non-abstract `get_region(self)` method in the base scraper which returns the initialized region object. Then children could invoke `get_region(self)` instead of `self.REGION`, which I think is a cleaner, safer interface.

recidiviz/ingest/us_ny/us_ny_scraper.py
```diff
@@ -86,1600 +78,1224 @@
...
-
-def parse_sentence_duration(term_string, inmate_id):
-    """Converts string describing sentence duration to models.SentenceDuration
-
-    For the 'Maximum Aggregate Sentence' and 'Minimum Aggregate Sentence'
-    results, the scraped string often looks like one of these:
-
-        "00 Years, 000 Months, 000 Days",
-        "04 Years, 002 Months, 000 Days",
-        "LIFE Years, 999 Month
```
jessex: Can we place a `TODO` on this and file an issue to investigate a more deterministic method for concluding we've reached the end of the DOCCS alphabet?

ohinds-vera:
Thanks @jessex! Great comments. I implemented most, with the notable exception of proposing that we leave the modification of return value consistency for a separate issue (while agreeing that your concerns should be addressed ASAP).

recidiviz/ingest/scraper.py
```diff
@@ -65,6 +59,17 @@ def inmate_id_to_record_id(self, inmate_id):
         pass
 
 
+    @abc.abstractmethod
+    def get_initial_task(self):
```
jessex: I think we should include docstrings for abstract methods, too. They help future implementors understand the interface that they are supposed to be implementing.

recidiviz/ingest/us_ny/us_ny_scraper.py
```diff
@@ -1031,8 +1036,8 @@ def link_inmate(self, record_list):
                 prior_inmate = prior_inmate_key.get()
                 inmate_id = prior_inmate.inmate_id
 
-                logging.info("Found an earlier record with an inmate ID " +
-                             inmate_id + ", using that.")
+                logging.info("Found an earlier record with an inmate ID %s "
+                             ", using that.", inmate_id)
```
jessex: Is it a stylistic choice to start this line off with the inline comma, or is that just a product of the way the line was physically updated? It jumped out to me so some aspect of my aesthetics must think it should be on the end of the previous line, but consciously I don't have a problem with it. Just want to be consistent from log statement to log statement.

recidiviz/ingest/us_ny/us_ny_scraper.py
```diff
@@ -1102,13 +1107,15 @@ def results_parsing_failure(self):
...
                 recent_sessions = sessions.get_recent_sessions(
-                    ScrapeKey(self.REGION.region_code, scrape_type))
+                    ScrapeKey(self.get_region().region_code, scrape_type))
 
                 for session in recent_sessions:
                     if session.last_scraped:
                         last_scraped = session.last_scraped
                         break
 
+            # TODO (ohinds): we need a more general method for
```
jessex: I'd like to file an issue in the repo for this and reference that with the TODO message, e.g. `# TODO #XYZ: Replace this with a more general method for detecting the end of the roster.` That way future contributors have the context they need and don't assume that you specifically will be addressing this.

Would you mind filing the issue? I can do so if you'd prefer!

ohinds-vera:
Thanks again @jessex! I dunno what I was thinking with those undocumented functions. Also, I filed the two relevant issues.

recidiviz/ingest/scraper.py
```diff
@@ -47,27 +47,53 @@ class Scraper(object):
...
 
     @abc.abstractmethod
     def inmate_id_to_record_id(self, inmate_id):
+        """Abstract method for child classes to map an inmate ID to a DB
+        record ID.
+
+        Args: (any type) the inmate ID to transform
+
+        Returns:
+            N/A
```
jessex: The return type should be describable at this level, since it's an interface. So I would change this to "A record id mapped from the given inmate id."

recidiviz/ingest/us_ny/us_ny_scraper.py
```diff
@@ -1036,8 +1036,8 @@ def link_inmate(self, record_list):
                 prior_inmate = prior_inmate_key.get()
                 inmate_id = prior_inmate.inmate_id
 
-                logging.info("Found an earlier record with an inmate ID %s "
-                             ", using that.", inmate_id)
+                logging.info("Found an earlier record with an inmate ID %s ,"
```
jessex: lol sorry for the eagle eye knit but we don't need that extra space.

jessex:
Two extremely minor comments from me, and then feel free to merge! Thanks, @ohinds-vera!

Issue #115, opened 2018-08-16
Add a scraper for Vermont to ingest


ohinds-vera:
From here:
http://doc.vermont.gov/offender-locator/

jessex:
Ready to close this one, @ohinds-vera?

jessex:
Ping, @ohinds-vera!

Issue #114, opened 2018-08-15
Investigate whether the scraper convention of returning -1 on failure and None on success is consistent with GAE, or arbitrary


ohinds-vera:
And if it's arbitrary, consider refactoring for a convention of throwing exceptions on errors and removing the explicit `return None` statements.

ohinds-vera:
When resolving this issue, make sure to update all children of Scraper, as well as Scraper itself.

jessex:
I agree in principal. What I need to look at in the code again is whether there are places where we are using -1 to mean non-exceptional in addition to exceptional. I think we should throw exceptions where something is exceptional and let it bubble to the right place in the stack (which may be all the way up to the request handler) and I think we should _not_ throw exceptions if it's purely as a means of control flow. In those cases, having some return value (maybe None) is sensible. Probably this isn't worth prioritizing until we've hit a steadier state in the ingest architecture, e.g. once we've gotten the persistence and database layers into place, and we've gotten monitoring into place.

+CC @arian487

Issue #44, opened 2018-01-31
Design validation layer to validate incoming source data records


jessex:
Source data that we ingest often has fairly rich modeling with a variety of fields whose meanings may not be obvious to the layperson. See [New York State's DOCCS definitions](http://www.doccs.ny.gov/univinq/fpmsdoc.htm) to get a feel for the kind of complexity we need to develop a deep understanding of. We have found that we often need to correct our understanding of a given field, or how multiple fields relate. These systems are often driven by manual entry, or hybrid human-machine systems, which are prone to the occasional error. Furthermore, different departments will have different data modeling, varying standards of quality, and different core terminology. All of this adds up to a probability asymptotically approaching 1 that we will misunderstand something in a given source data system.

Having the ability to encode and validate our assumptions within the data would surface potentially critical issues earlier in the development process, and generate a greater collective body of knowledge. For example, if we are relying on the belief that one date in the data model is always after another date, then we want to know if that's not true for a particular record. 

One approach is to add to our ingest platform a region-specific interface for validating assumptions on a schedule, i.e. after we have already ingested. Another approach would be to bake a validation framework directly into ingestion itself, so that we can make a call as to whether a particular record _should_ be ingested. The semantics and implementations would be meaningfully different.

jessex:
A related task would be a tool that we could run on-demand to dump out all unique values that exist for specific fields within a schema. For example, a task that provided all of the unique values for `record.parole_hearing_type` would be useful. We can do this on-demand today by directly querying Datastore with GQL, though, so the utility is limited if there's no use case that calls for doing this automatically // periodically.

jessex:
Another part of this -- perhaps the most important part -- would be live validation in the ingest pipeline. We'd do this in a structured way, with a subsystem that allows us to define validation rules, record metrics for any records failing a particular rule, and logic for determining how to handle offending records. The strategy for the latter may be to mark these records as invalid to be filtered and/or handled specially in downstream pipelines if necessary, i.e. calculation.

jessex:
Closing this in favor of #196. +CC @jamwalla

PR #116, opened 2018-09-05
Add a scraper for Vermont to ingest (#115)

ohinds-vera:
This addresses issue #115. It adds a scraper that ingests the unified jail and prison system of the state of Vermont. That jurisdiction uses JailTracker, so this commit can be the basis for future JailTracker scrapers. 

* Create a `us_vt` directory in ingest
  - Fill with classes to handle the scraping and models for UsVt
  - This jurisdiction uses JailTracker, so this commit can serve as a
    basis for future jurisdictions that use the JailTracker system.

* Update the web app configuration to add queues for scraping `us_vt`.

* Update a test to include the new 'us_vt' region in the list of all
  expected regions.

### DIFF
--- a/cron.yaml
+++ b/cron.yaml
@@ -26,3 +26,21 @@ cron:
   url: /calculator_pipeline?region=us_ny
   schedule: every saturday 12:00
   timezone: America/New_York
+
+- description: Start VT background scraper every day at 9pm
+  url: /scraper/start?region=us_vt&scrape_type=background
+  schedule: every day 21:00
+  timezone: America/New_York
+  retry_parameters:
+    min_backoff_seconds: 2.5
+    max_doublings: 5
+    job_age_limit: 9h
+
+- description: Stop VT scraper every day at 9am
+  url: /scraper/stop?region=us_vt&scrape_type=background
+  schedule: every day 09:00
+  timezone: America/New_York
+  retry_parameters:
+    min_backoff_seconds: 2.5
+    max_doublings: 5
+    job_age_limit: 9h
--- a/recidiviz/ingest/scraper.py
+++ b/recidiviz/ingest/scraper.py
@@ -24,15 +24,28 @@ import abc
 import json
 import logging
 
+import requests_toolbelt.adapters.appengine
 import requests
 
 from google.appengine.ext import deferred
 from google.appengine.api import taskqueue
+from google.appengine.api import urlfetch
 from recidiviz.ingest import scraper_utils
 from recidiviz.ingest import sessions
 from recidiviz.ingest.models.scrape_key import ScrapeKey
 from recidiviz.ingest import tracker
 from recidiviz.utils.regions import Region
+
+
+# Use the App Engine Requests adapter to make sure that Requests plays
+# nice with GAE
+requests_toolbelt.adapters.appengine.monkeypatch()
+urlfetch.set_default_fetch_deadline(60)
+
+# Squelch urllib3 / sockets platform warning
+requests.packages.urllib3.disable_warnings(
+    requests.packages.urllib3.contrib.appengine.AppEnginePlatformWarning
+)
 
 
 class Scraper(object):
@@ -59,7 +72,6 @@ class Scraper(object):
             self.get_region().region_code + "_next_page_fail_counter")
         self.scraper_work_url = '/scraper/work'
 
-
     @abc.abstractmethod
     def inmate_id_to_record_id(self, inmate_id):
         """Abstract method for child classes to map an inmate ID to a DB
@@ -73,7 +85,6 @@ class Scraper(object):
         """
         pass
 
-
     @abc.abstractmethod
     def get_initial_task(self):
         """Abstract method for child classes to specify the name of the first
@@ -85,7 +96,6 @@ class Scraper(object):
         """
         pass
 
-
     def get_region(self):
         """Retrieve the region object associated with this scraper.
 
@@ -94,7 +104,6 @@ class Scraper(object):
 
         """
         return self.region
-
 
     def start_scrape(self, scrape_type):
         """Start new scrape session / query against corrections site
@@ -120,7 +129,6 @@ class Scraper(object):
         params = {"scrape_type": scrape_type, "content": docket_item}
 
         self.add_task(self.get_initial_task(), params)
-
 
     def stop_scrape(self, scrape_types):
         """Stops all active scraping tasks, resume non-targeted scrape types
@@ -160,7 +168,6 @@ class Scraper(object):
 
         q = taskqueue.Queue(self.get_region().queues[0])
         q.purge()
-
 
     def resume_scrape(self, scrape_type):
         """Resume a stopped scrape from where it left off
@@ -219,7 +226,6 @@ class Scraper(object):
         params = {'scrape_type': scrape_type, 'content': content}
         self.add_task(self.get_initial_task(), params)
 
-
     def fetch_page(self, url, data=None):
         """Fetch content from a URL. If data is None (the default), we perform
         a GET for the page. If the data is set, it must be a dict of parameters
@@ -270,7 +276,6 @@ class Scraper(object):
 
         return page
 
-
     def add_task(self, task_name, params):
         """ Add a task to the task queue.
 
@@ -293,7 +298,6 @@ class Scraper(object):
                       params={'region': self.get_region().region_code,
                               'task': task_name,
                               'params': params_serial})
-
 
     def iterate_docket_item(self, scrape_type):
         """Leases new docket item, updates current session, returns item
--- a/recidiviz/ingest/us_vt/__init__.py
+++ b/recidiviz/ingest/us_vt/__init__.py
@@ -15,19 +15,15 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 # =============================================================================
 
-"""The ingest portion of the Recidiviz data platform.
+"""us_vt package
 
-This includes infrastructure, logic, and models for ingesting, validating,
-normalizing, and storing records ingested from various criminal justice data
-sources.
+Vermont has a unified jail and prison system, so this package handles
+scraping and storing data of both people who would be in jail and
+prison in another state. In addition, Vermont has information on
+people on probabtion and parole.
+
+Vermont uses the JailTracker system.
+
 """
 
-
-import recidiviz.ingest.docket
-import recidiviz.ingest.models
-import recidiviz.ingest.scraper_control
-import recidiviz.ingest.sessions
-import recidiviz.ingest.tracker
-import recidiviz.ingest.us_ny
-import recidiviz.ingest.us_vt
-import recidiviz.ingest.worker
+from recidiviz.ingest.us_vt import us_vt_scraper
--- a/recidiviz/ingest/us_vt/us_vt_record.py
+++ b/recidiviz/ingest/us_vt/us_vt_record.py
@@ -1,0 +1,63 @@
+# Recidiviz - a platform for tracking granular recidivism metrics in real time
+# Copyright (C) 2018 Recidiviz, Inc.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+# =============================================================================
+
+"""us_vt record-specific functionality.
+"""
+
+
+from google.appengine.ext import ndb
+from recidiviz.models.record import Offense
+
+
+class UsVtOffense(Offense):
+    """UsVt-specific model to describe a particular crime of conviction
+
+    TODO #124 promote non-Vermont specific fields to Charge,
+    Conviction, and Arrest when #124 is closed.
+
+    Attributes:
+        status: (string) The legal status of the charge in court.
+        case_number: (string) Case number of this offense.
+        bond_amount: (float) Dollar amount of bond on this charge, if any.
+        bond_type: (string) How the bond must be secured.
+        number_of_counts: (integer) Counts in this offense.
+        arresting_agency: (string) Name of the agency that arrested this person
+            on this charge.
+        court_type: (string) Type of court that will hear/heard this case.
+        warrant_number: (string) The warrant number for this charge, if any.
+        court_time: (date) Date of the next court appearance.
+        control_number: (string) Unsure what this is.
+        crime_type: (string) Type of crime.
+        arrest_code: (string) Identifier of the arrest event.
+        modifier: (string) Unsure what this is.
+        arrest_date: (date) Date of arrest.
+
+    """
+    status = ndb.StringProperty()
+    case_number = ndb.StringProperty()
+    bond_amount = ndb.FloatProperty()
+    bond_type = ndb.StringProperty()
+    number_of_counts = ndb.IntegerProperty()
+    arresting_agency = ndb.StringProperty()
+    court_type = ndb.StringProperty()
+    warrant_number = ndb.StringProperty()
+    court_time = ndb.DateProperty()
+    control_number = ndb.StringProperty()
+    crime_type = ndb.StringProperty()
+    arrest_code = ndb.StringProperty()
+    modifier = ndb.StringProperty()
+    arrest_date = ndb.DateProperty()
--- a/recidiviz/ingest/us_vt/us_vt_scraper.py
+++ b/recidiviz/ingest/us_vt/us_vt_scraper.py
@@ -1,0 +1,730 @@
+# Recidiviz - a platform for tracking granular recidivism metrics in real time
+# Copyright (C) 2018 Recidiviz, Inc.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+# =============================================================================
+
+
+"""Scraper implementation for people incarcerated in Vermont.
+
+Region-specific notes:
+
+    - Scrape the front page to get the session key
+    - Make a requests to the GetInmates endpoint N records at a time
+      (say 500) until all rows of the roster have been retrieved.
+    - For each person on the roster, use the ArrestNo in the person's
+      row to grab the GetInmate, GetCases, and GetCharges info.
+    - Create the person and record objects and store them in the datastore.
+
+"""
+
+
+from copy import deepcopy
+import json
+import logging
+import re
+
+from lxml import html
+from lxml.etree import XMLSyntaxError # pylint:disable=no-name-in-module
+
+from google.appengine.ext import ndb
+from google.appengine.ext.db import InternalError
+from google.appengine.ext.db import Timeout, TransactionFailedError
+
+from recidiviz.ingest import scraper_utils
+from recidiviz.ingest.scraper import Scraper
+from recidiviz.ingest.us_vt.us_vt_person import UsVtPerson
+from recidiviz.ingest.us_vt.us_vt_record import UsVtOffense
+from recidiviz.ingest.us_vt.us_vt_snapshot import \
+    UsVtSnapshot
+from recidiviz.models.record import Record
+
+
+class UsVtScraper(Scraper):
+    """Class that scrapes information on people held in Vermont facilities.
+
+    """
+
+    def __init__(self):
+
+        self.front_url = '/jailtracker/index/Vermont'
+        self.roster_url = ('(S({session}))//(S({session}))/JailTracker/'
+                           'GetInmates?start={start}&limit={limit}'
+                           '&sort=LastName&dir=ASC')
+        self.person_url = ('(S({session}))//(S({session}))/JailTracker/'
+                           'GetInmate?arrestNo={arrest}')
+        self.cases_url = ('(S({session}))//(S({session}))/JailTracker/'
+                          'GetCases?arrestNo={arrest}')
+        self.charges_url = ('(S({session}))//(S({session}))/JailTracker/'
+                            'GetCharges')
+
+        super(UsVtScraper, self).__init__('us_vt')
+
+    def get_initial_task(self):
+        """
+        Get the name of the first task to be run.
+
+        Returns:
+            The name of the task to run first.
+
+        """
+        return 'scrape_front_page'
+
+    def scrape_front_page(self, params):
+        """Scrapes the front page for the VT DOC site to get
+        the session key. Enqueues tasks to scrape the roster.
+
+        Args:
+            params: Dict of parameters, includes:
+                scrape_type: 'background' or 'snapshot'
+
+        Returns:
+            Nothing if successful, -1 if fails.
+
+        """
+
+        def get_session_key(html_tree):
+            """Helper function to extract the session key from the html of the
+            front page of the Vermont DOC site.
+
+            """
+
+            # The session key is in a string parameter in a javascript
+            # call that initializes the jailtracker system.
+            body_script = html.tostring(
+                html_tree.xpath("//body/div/script")[0])
+            try:
+                session_key = re.search(
+                    r"JailTracker.Web.Settings.init\('(.*)'",
+                    body_script).group(1)
+            except AttributeError:
+                logging.error("Error, could not parse session key from the "
+                              "front page HTML")
+                return None
+
+            return session_key
+
+        url = '/'.join([self.get_region().base_url, self.front_url])
+        front_page = self.fetch_page(url)
+        if front_page == -1:
+            return -1
+
+        try:
+            front_tree = html.fromstring(front_page.content)
+        except XMLSyntaxError as e:
+            logging.error("Error parsing front page. Error: %s\nPage:\n\n%s",
+                          e, front_page.content)
+            return -1
+
+        session = get_session_key(front_tree)
+        if session is None:
+            return -1
+
+        first_roster_params = {
+            'session': session,
+            'start': 0,
+            'limit': 500,
+            'scrape_type': params['scrape_type']}
+
+        self.add_task('scrape_roster', first_roster_params)
+
+        return None
+
+    def scrape_roster(self, params):
+        """Scrapes the roster for VT to get a list of people. Enqueues tasks
+        to scrape the individual listings.
+
+        Args:
+            params: Dict of parameters, includes:
+                scrape_type: 'background' or 'snapshot'
+                session: (string) session key (required)
+                start: (int) index of the first row of the roster to retrieve
+                       (required)
+                limit: (int) the number of rows to retrieve (required)
+
+        Returns:
+            Nothing if successful, -1 if fails.
+
+        """
+
+        url = '/'.join([self.get_region().base_url, self.roster_url]).format(
+            session=params['session'], start=params['start'],
+            limit=params['limit'])
+
+        roster_response = self.fetch_page(url)
+
+        roster = json.loads(roster_response.content)
+
+        # for termination condition testing and to know which start
+        # row to request next
+        max_row_index = -1
+
+        # add a task for each person in the roster
+        for person in roster['data']:
+            if person['RowIndex'] > max_row_index:
+                max_row_index = person['RowIndex']
+
+            next_person_params = {
+                'session': params['session'],
+                'roster_entry': person,
+                'scrape_type': params['scrape_type']
+            }
+
+            self.add_task('scrape_person', next_person_params)
+
+        # if we aren't done reading the roster, read another page
+        if roster['totalCount'] > max_row_index:
+            next_roster_params = {
+                'start': max_row_index,  # 1- vs. 0-based indexing :(
+                'limit': params['limit'],
+                'session': params['session'],
+                'scrape_type': params['scrape_type']}
+
+            self.add_task('scrape_roster', next_roster_params)
+
+        return None
+
+    def scrape_person(self, params):
+        """Scrape the information about an individual.
+
+        Fetches the general information about an individual, and
+        enqueues the task to scrape the cases information associated
+        with the individual.
+
+        Args:
+            params: Dict of parameters.
+                scrape_type: 'background' or 'snapshot'
+                session: (string) session key (required)
+                roster_entry: (dict) key/value pairs with info contained in
+                    the roster entry
+        """
+
+        url = '/'.join([self.get_region().base_url, self.person_url]).format(
+            session=params['session'],
+            arrest=params['roster_entry']['ArrestNo'])
+        person_response = self.fetch_page(url)
+        person = json.loads(person_response.content)
+
+        cases_params = {
+            'session': params['session'],
+            'roster_entry': params['roster_entry'],
+            'person': person,
+            'scrape_type': params['scrape_type']}
+
+        self.add_task('scrape_cases', cases_params)
+
+
+    def scrape_cases(self, params):
+        """Scrape the case information about an individual case.
+
+        Args:
+            params: Dict of parameters.
+                scrape_type: 'background' or 'snapshot'
+                session: (string) session key (required)
+                roster_entry: (dict) key/value pairs with info contained in
+                    the roster entry
+                person: (dict) key/value pairs with info on the person panel
+        """
+
+        url = '/'.join([self.get_region().base_url, self.cases_url]).format(
+            session=params['session'],
+            arrest=params['roster_entry']['ArrestNo'])
+        cases_response = self.fetch_page(url)
+        cases = json.loads(cases_response.content)
+
+        charges_params = {
+            'session': params['session'],
+            'roster_entry': params['roster_entry'],
+            'person': params['person'],
+            'cases': cases,
+            'scrape_type': params['scrape_type']}
+
+        self.add_task('scrape_charges', charges_params)
+
+    def scrape_charges(self, params):
+        """Scrape the charge information about an individual charge.
+
+        Args:
+            params: Dict of parameters.
+                scrape_type: 'background' or 'snapshot'
+                session: (string) session key (required)
+                roster_entry: (dict) key/value pairs with info contained in
+                    the roster entry
+                person: (dict) key/value pairs with info on the person panel
+                cases: (dict) key/value pairs with info from each of the cases
+                    from the cases list
+
+        """
+
+        url = '/'.join([self.get_region().base_url, self.charges_url]).format(
+            session=params['session'])
+        data = {'arrestNo': params['roster_entry']['ArrestNo']}
+
+        charges_response = self.fetch_page(url, data=data)
+        charges = json.loads(charges_response.content)
+
+        self.store_record(params['roster_entry'], params['person'],
+                          params['cases'], charges)
+
+    @staticmethod
+    def extract_agencies(person_data, person_id):
+        """Get the list of agencies with jurisdiction over this person. There
+        can be more than one agency with jurisdiction if someone is
+        incarcerated but already has a parole agency assigned.
+
+        Args:
+
+            person_data: (list of key/value pairs) The data scraped
+                from the person's detail page.
+
+            person_id: (string) Person identifier for logging purposes.
+
+        Returns:
+
+            A tuple where the first entry is the name of the
+            residential facility with custody over this person (or
+            None) and the second entry is the probation or parole
+            office in charge of supervising this person (or None).
+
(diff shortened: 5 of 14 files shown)


cron.yaml
```diff
@@ -26,3 +26,12 @@ cron:
   url: /calculator_pipeline?region=us_ny
   schedule: every saturday 12:00
   timezone: America/New_York
+
+- description: Start VT background scraper every day at 9pm
+  url: /scraper/start?region=us_vt&scrape_type=background
```
jessex: And there's no automatic configured stop because we believe we can crawl the entirety of their system in a matter of a few hours? Is there data non-historical?

recidiviz/ingest/scraper.py
```diff
@@ -242,6 +242,8 @@ def fetch_page(self, url, data=None):
             else:
                 page = requests.post(url, proxies=proxies, headers=headers,
                                      data=data)
+            import time
+            time.sleep(1)
```
jessex: Was this only for local debugging? If so, let's make sure we drop this before merging.

recidiviz/ingest/us_vt/__init__.py
```diff
@@ -0,0 +1,21 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+# =============================================================================
+
+"""us_vt package
```
jessex: I think I would use these top-level package docs for jurisdiction or vendor packages to discuss any high level details of that system that are worth mentioning, vis-a-vis the code found within the package. For example, it's worth calling out that this Vermont system combines local jail with state prison data, and that the ingest code is covering both. There could be other things worth mentioning.

recidiviz/ingest/us_vt/us_vt_inmate.py
```diff
@@ -0,0 +1,43 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+# =============================================================================
+
+"""Tools for environment variables in Google App Engine."""
```
jessex: Copy paste detritus.

recidiviz/ingest/us_vt/us_vt_inmate.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+from google.appengine.ext import ndb
+from recidiviz.models.inmate import Inmate
+
+
+class UsVtInmate(Inmate):
```
jessex: Even if the code you're inheriting from is not yet updated to use the `Person` terminology on master (it is on my branch, though!), we should go ahead and update everything we can. So this line would be `class UsVtPerson(Inmate)` and the doc below would avoid use of the term inmate, and the first field would be `us_vt_person_id`, so forth.

recidiviz/ingest/us_vt/us_vt_inmate.py
```diff
@@ -0,0 +1,43 @@
...
+    Vermont. This extends the Inmate class.
+
+    Attributes:
+
+        us_vt_inmate_id: (string) Same as person_id, but used as
+            key for this entity type to force uniqueness / prevent
+            collisions within the us_vt records (see
+            models.inmate for inherited attributes)
+        us_vt_person_suffix: (string) Suffix portion of the person's name.
+        us_vt_person_alias: (string) Alias(es) known to be used by this person.
```
jessex: These feel like the kind of fields we'd want to promote straight up to `Person`. Aliases and name suffices, while not in New York DOCCS, will likely show up elsewhere and are general, non-jurisdictional data fields. @cawarren to confirm I'm on the right track here.

recidiviz/ingest/us_vt/us_vt_record.py
```diff
@@ -0,0 +1,94 @@
...
+    arrest_date = ndb.DateProperty()
+
+
+class UsVtRecord(Record):
+    """A subclass of Record that adds Vermont specific fields
+
+    Datastore model for Vermont-specific fields in incarceration
+    records. A single record describes one sentencing event. E.g., if
+    a person were to commit armed robbery and larceny in one night,
+    arrested, and given a sentence for N years for those crimes, this
```
jessex: `be arrested` or `get arrested`

recidiviz/ingest/us_vt/us_vt_record.py
```diff
@@ -0,0 +1,94 @@
...
+    would be described in a single Record.
+
+    UsVtRecord entities are stored as children of their
+    respective inmate entities. There may be multiple records per
```
jessex: Scrub inmate in favor of person across the board in this PR.

recidiviz/ingest/us_vt/us_vt_record.py
```diff
@@ -0,0 +1,94 @@
...
+    inmate.
+
+    Attributes:
+        status: (string) Status of the person at this time (sentenced,
+            detained, etc.)
+        release_date: (date) Date of actual release.
+        earliest_release_date: (date) Earliest date release is possible.
+        community_supervision_agency: (string) The parole or probation office
```
jessex: Hey! Parole and probation fields! All three of these are fantastic. I hope we see them show up in more systems.

recidiviz/ingest/us_vt/us_vt_record.py
```diff
@@ -0,0 +1,94 @@
...
+    a person were to commit armed robbery and larceny in one night,
+    arrested, and given a sentence for N years for those crimes, this
+    would be described in a single Record.
+
+    UsVtRecord entities are stored as children of their
+    respective inmate entities. There may be multiple records per
+    inmate.
+
+    Attributes:
+        status: (string) Status of the person at this time (sentenced,
```
jessex: Similar to my comment on the fields on `UsVtPerson`, I think you could argue that all of these fields could be promoted to `Record` straight-away. None of them are Vermont-specific, conceptually. @cawarren to provide input here, too.

recidiviz/ingest/us_vt/us_vt_record.py
```diff
@@ -0,0 +1,94 @@
...
+            detained, etc.)
+        release_date: (date) Date of actual release.
+        earliest_release_date: (date) Earliest date release is possible.
+        community_supervision_agency: (string) The parole or probation office
+            that does or will have jurisdiction over this person after release.
+        parole_officer: (string) Name of the parole officer.
+        case_worker: (string) Name of the case worker.
+        earliest_release_date: (date) Earliest date to be released based on
```
jessex: This is copy paste detritus.

recidiviz/ingest/us_vt/us_vt_record.py
```diff
@@ -0,0 +1,94 @@
...
+        court_type: (string) Type of court that will hear/heard this case.
+        warrant_number: (string) The warrant number for this charge, if any.
+        court_time: (date) Date of the next court appearance.
+        control_number: (string) Unsure what this is.
+        crime_type: (string) Type of crime.
+        arrest_code: (string) Identifier of the arrest event.
+        modifier: (string) Unsure what this is.
+        arrest_date: (date) Date of arrest.
+    """
+    status = ndb.StringProperty()
```
jessex: I don't yet know enough about correctional data systems to say whether I believe _all_ of the fields here could be promoted straight away to `Offense`, i.e. are any of these fields particular to the Vermont penal system, but I get the feeling that we'd likely want to promote most or all of these. Where discussion is helpful here is in the fact that some of these fields are specific to jails and not prisons, e.g. the fields related to cash bonds. How do we want to demarcate records (and offenses) that are jail-specific, or prison-specific, or show someone "graduating" from jail to prison?

recidiviz/ingest/us_vt/us_vt_record.py
```diff
@@ -0,0 +1,94 @@
...
+
+from google.appengine.ext import ndb
+from recidiviz.models.record import Offense, Record
+
+
+class UsVtOffense(Offense):
+    """UsVt-specific model to describe a particular crime of conviction
+
+    Attributes:
+        status: (string) The legal status of the charge in court.
```
jessex: Some of the fields in this offense sub-class will change, which was not the case with UsNy. But we should be good here since `Offense` is included in `Snapshot`. So if we run another scrape and see that the legal status has changed since the last scrape, we'll create a new `Snapshot` which includes an `Offense` field that itself includes a `status` field. _However_, today there are no _nested_ fields inside of `Record`'s object graph that could change (sentence and offense are effectively immutable), so what we _should_ do is make sure that for nested fields we follow our modeling for the `Snapshot` itself: if only the `UsVtOffense.status` field changes between scrapes, then `Snapshot.offense.status` should be set and nothing else in `Snapshot` should be set.

Do we agree on that?

recidiviz/ingest/us_vt/us_vt_snapshot.py
```diff
@@ -0,0 +1,48 @@
...
+# =============================================================================
+
+"""us_vt snapshot-specific functionality
+"""
+
+from google.appengine.ext import ndb
+from recidiviz.models.snapshot import Snapshot
+
+
+class UsVtSnapshot(Snapshot):
```
jessex: If we promote those fields from `UsVtRecord` straight to `Record`, then this goes away, too.

recidiviz/ingest/us_vt/us_vt_scraper.py
```diff
@@ -0,0 +1,614 @@
...
+from recidiviz.ingest.scraper import Scraper
+from recidiviz.ingest.us_vt.us_vt_inmate import UsVtInmate
+from recidiviz.ingest.us_vt.us_vt_record import UsVtOffense
+from recidiviz.ingest.us_vt.us_vt_record import UsVtRecord
+from recidiviz.ingest.us_vt.us_vt_snapshot import \
+    UsVtSnapshot
+
+# Use the App Engine Requests adapter to make sure that Requests plays
+# nice with GAE
+requests_toolbelt.adapters.appengine.monkeypatch()
```
jessex: If we shift this up to `Scraper`, will all children inherit it nicely?

jessex: Same goes for the two lines directly below.

recidiviz/ingest/us_vt/us_vt_scraper.py
```diff
@@ -0,0 +1,614 @@
...
+
+class UsVtScraper(Scraper):
+    """Class that scrapes information on people held in Vermont facilities.
+
+    """
+
+    def __init__(self):
+
+        self.front_url = '/jailtracker/index/Vermont'
+        self.roster_url = ('(S({session}))//(S({session}))/JailTracker/'
```
jessex: Woah! Can't say I've ever used this syntax before in Python. Can you link me to some docs to learn more about what's going on here under the hood?

recidiviz/ingest/us_vt/us_vt_scraper.py
```diff
@@ -0,0 +1,614 @@
...
+        Get the name of the first task to be run.
+
+        Returns:
+            The name of the task to run first.
+
+        """
+        return 'scrape_front_page'
+
+    def scrape_front_page(self, params):
+        """Scrapes the front page for the VT DOC page to get
```
jessex: "front page for the VT DOC _site_" maybe?

recidiviz/ingest/us_vt/us_vt_scraper.py
```diff
@@ -0,0 +1,614 @@
...
+        the session key. Enqueues tasks to scrape the roster.
+
+        Args:
+            params: Dict of parameters, includes:
+                scrape_type: 'background' or 'snapshot'
+
+        Returns:
+            Nothing if successful, -1 if fails.
```
jessex: I know we discussed this on the other PR related to `Scraper`'s -1 error handling. Are you following the leader on this one until #114 or did you proactively choose this approach?

recidiviz/ingest/us_vt/us_vt_scraper.py
```diff
@@ -0,0 +1,614 @@
...
+        """
+        url = '/'.join([self.REGION.base_url, self.front_url])
+        front_page = self.fetch_page(url)
+        if front_page == -1:
+            return -1
+
+        try:
+            front_tree = html.fromstring(front_page.content)
+        except XMLSyntaxError as e:
+            logging.error("Error parsing front page. Error: %s\nPage:\n\n%s" %
```
jessex: Remember to use the more modern logging convention:
`logging.error("Error parsing front page. Error: %s\nPage:\n\n%s", e, front_page.content)`

You shouldn't need the `str` calls with this approach, either, but test that.

recidiviz/ingest/us_vt/us_vt_scraper.py
```diff
@@ -0,0 +1,614 @@
...
+                          (str(e), str(front_page.content)))
+            return -1
+
+        try:
+            body_script = html.tostring(
```
jessex: I think it would be worth extracting this session finding bit into a small utility function, with a doc on what exactly it's looking for with the page content.

recidiviz/ingest/us_vt/us_vt_scraper.py
```diff
@@ -0,0 +1,614 @@
...
+
+        """
+
+        url = '/'.join([self.REGION.base_url, self.roster_url]).format(
+            session=params['session'], start=params['start'],
+            limit=params['limit'])
+
+        roster_response = self.fetch_page(url)
+
+        logging.info(roster_response.content)
```
jessex: This feels like a `debug` to me, but you may already intend to make that change prior to merging.

recidiviz/ingest/us_vt/us_vt_scraper.py
```diff
@@ -0,0 +1,614 @@
...
+        # if we aren't done reading the roster, read another page
+        if roster['totalCount'] > roster['data'][-1]['RowIndex']:
+            next_roster_params = {
+                'start': roster['data'][-1]['RowIndex'] + 1,
+                'limit': params['limit'],
+                'session': params['session'],
+                'scrape_type': params['scrape_type'],
+                'content': params['content']}
+
+            self.add_task('scrape_roster', next_roster_params)
```
jessex: Since the system at large is asynchronous, enqueuing a task to grab another chunk of the roster before we have enqueued every individual task in this chunk means that we will end up with individual tasks from different chunks interleaved, breaking the natural ordering of the source data system. Is this acceptable?

recidiviz/ingest/us_vt/us_vt_scraper.py
```diff
@@ -0,0 +1,614 @@
...
+
+    def scrape_person(self, params):
+        """Scrape the information about an individual.
+
+        Fetches the general information about an individual, and
+        enqueues the task to scrape the cases information associated
+        with the individual.
+
+        Args:
+            params: Dict of parameters.
```
jessex: Let's describe the params in detail like in other method docs in here.

recidiviz/ingest/us_vt/us_vt_scraper.py
```diff
@@ -0,0 +1,614 @@
...
+        person = json.loads(person_response.content)
+
+        cases_params = {
+            'session': params['session'],
+            'roster_entry': params['roster_entry'],
+            'person': person,
+            'scrape_type': params['scrape_type'],
+            'content': params['content']}
+
+        self.add_task('scrape_cases', cases_params)
```
jessex: Re-affirming our previous discussion, I think this approach of isolating fetches 1:1 to an enqueued task by basically daisy-chaining these calls together makes more sense than having some globby model where one task makes many requests. :+1:

recidiviz/ingest/us_vt/us_vt_scraper.py
jessex: I'm guessing it must be fine since we're yielding to the same forces with the daisy-chaining below, but want to be sure we've made that as an explicit choice.

recidiviz/ingest/us_vt/us_vt_scraper.py
```diff
@@ -0,0 +1,614 @@
...
+        """
+
+        def field_val_arr_to_dict(field_val_arr):
+            """Convert an array of dict entries where each dict consists of only
+            the fields 'Field' and 'Value' to a dictionary with the
+            values of those fields as the keys and values.
+
+            """
+            dictionary = {}
+            for el in field_val_arr:
```
jessex: I don't love Hungarian notation, which is what `field_val_arr` basically is. I also don't love the admittedly widespread convention of referring to elements as `el` or `elem`. Could we change this to, maybe, `for pair in field_value_pairs`?

recidiviz/ingest/us_vt/us_vt_scraper.py
```diff
@@ -0,0 +1,614 @@
...
+            return dictionary
+
+        # extract the agency list for future processing, then collapse
+        # the field/value array into a dicitonary
+
+        # when multiple agencies are present, 'Field' is None for
+        # agencies after the first, but they are sequential.
+        agencies = []
+        last_was_agency = False
+        for entry in person['data']:
```
jessex: I think this logic is tricky enough that it deserves method extraction and dedicated unit tests.

recidiviz/ingest/us_vt/us_vt_scraper.py
```diff
@@ -0,0 +1,614 @@
...
+            person: (dict) Key/value results parsed from the person's page
+            cases: (dict) Key/value results parsed from the cases page
+            charges: (dict) Key/value results parsed from the charges page
+
+        Returns:
+            Nothing if successful, -1 if fails.
+
+        """
+
+        def field_val_arr_to_dict(field_val_arr):
```
jessex: Similarly, I'd call this `field_value_pairs_to_dict`.

recidiviz/ingest/us_vt/us_vt_scraper.py
```diff
@@ -0,0 +1,614 @@
...
+            inmate.race = 'white'
+        else:
+            inmate.race = person['Race'].lower()
+
+        inmate.inmate_id = person_id
+        inmate.us_vt_inmate_id = person_id
+        inmate.inmate_id_is_fuzzy = True
+        inmate.region = self.get_region().region_code
+
+        logging.info(inmate)
```
jessex: Similarly, let's make this `debug` before merging?

recidiviz/ingest/us_vt/us_vt_scraper.py
```diff
@@ -0,0 +1,614 @@
...
+
+        try:
+            person_key = inmate.put()
+        except (Timeout, TransactionFailedError, InternalError):
+            # Datastore error - fail task to trigger queue retry + backoff
+            logging.warning("Couldn't persist person: %s" % person_id)
```
jessex: Let's adjust all logging statements like I've called out previously.

recidiviz/ingest/us_vt/us_vt_scraper.py
```diff
@@ -0,0 +1,614 @@
...
+        old_record = deepcopy(record)
+
+        record.custody_date = scraper_utils.parse_date_string(
+            person['Booking Date'], arrest_no)
+
+        # store each charge as an offense to be added to the record
+        record_offenses = []
+        for charge in charges:
+            offense = UsVtOffense(
+                crime_description=charge['ChargeDescription'],
```
jessex: I'd like to see fields in all uses (declaration in Polymodel classes, usage in methods, etc.) to be alphabetically ordered with exceptions for fields that _obviously_ (to everyone) go together, such as surname and given names.

recidiviz/ingest/us_vt/us_vt_scraper.py
```diff
@@ -0,0 +1,614 @@
...
+
+        # guess facility vs. community supervision agency by looking
+        # for the string 'parole' in the agency name, and taking the
+        # first entry without it.
+        #
+        # this is a hacky method that should be verified.
+        facilities = [agency for agency in agencies
+                      if 'parole' not in agency.lower().rstrip()]
+        if not facilities.empty():
+            record.latest_facility = facilities[0]
```
jessex: Is it possible for a Vermont record to have multiple incarceration facilities listed?

recidiviz/ingest/us_vt/us_vt_scraper.py
```diff
@@ -0,0 +1,614 @@
...
+
+        cs_agencies = [agency for agency in agencies
+                       if 'parole' in agency.lower().rstrip()]
```
jessex: Rather than duplicate the logic but include `not` in one of the calls, I'd rather extract this out as a function that we can pass into filter calls. I think the clean approach is to declare and initialize a lambda that returns `parole in agency.lower().rstrip()` and pass that into both comprehensions individually.

recidiviz/ingest/us_vt/us_vt_scraper.py
```diff
@@ -0,0 +1,614 @@
...
+        if not cs_agencies.empty():
+            record.community_supervision_agency = cs_agencies[0]
```
jessex: Can there be multiple supervision agencies in here?

recidiviz/ingest/us_vt/us_vt_scraper.py
```diff
@@ -0,0 +1,614 @@
...
+            return -1
+
+        # PERSON RECORD SNAPSHOT
+        new_snapshot = self.record_to_snapshot(record)
+
+        self.compare_and_set_snapshot(old_record, new_snapshot)
+
+        return None
+
+    def link_person(self, arrest_nos):
```
jessex: I don't see this being used. Am I missing it?

recidiviz/ingest/us_vt/us_vt_scraper.py
```diff
@@ -0,0 +1,614 @@
...
+        """Checks for prior records matching newly scraped ones, returns
+        person ID
+
+        Matches DIN (record IDs) to previously scraped records, looks
+        up associated persons, then returns that person_id so we can
+        update the same person rather than duplicating the person.
+
+        Args:
+            arrest_nos: (list of strings) List of arrest numers to check
```
jessex: numbers

recidiviz/ingest/us_vt/us_vt_scraper.py
```diff
@@ -0,0 +1,614 @@
...
+
+        Returns:
+            The found person_id from prior instance, or None if not found
+
```
jessex: Extra line of whitespace.

recidiviz/ingest/us_vt/us_vt_scraper.py
```diff
@@ -0,0 +1,614 @@
...
+                prior_person = prior_person_key.get()
+                person_id = prior_person.inmate_id
+
+                logging.info("Found an earlier record with an person ID %s "
+                             ", using that.", person_id)
+                return person_id
+
+        return None
+
+    def inmate_id_to_record_id(self, inmate_id):
```
jessex: I don't see this being used, either.

recidiviz/ingest/us_vt/us_vt_scraper.py
```diff
@@ -0,0 +1,614 @@
...
+
+        Args:
+            record: A Record object to mirror
+
+        Returns:
+            A Snapshot entity populated with the same details as the Record
+
+        """
+        snapshot = UsVtSnapshot(
+            parent=record.key,
```
jessex: Same goes for alphabetical ordering here. I know that hasn't been a convention so far but I'd like to push us in that direction and plan to do so on my own migration branch with some other new code.

recidiviz/ingest/us_vt/us_vt_scraper.py
```diff
@@ -0,0 +1,614 @@
...
+
+        return snapshot
+
+    def compare_and_set_snapshot(self, old_record, snapshot):
+        """Check for updates since last scrape, if found persist new snapshot
+        Checks the last snapshot taken for this record, and if fields
+        have changed since the last time the record was updated (or
+        the record has no old snapshots) stores new snapshot.
+
+        TODO: this function should be abstracted into the Scraper class.
```
jessex: Yes! Do you plan to do so in this PR? If not, let's make sure it's documented on some issue and the issue number is referenced on the TODO. In general, I don't want to see any TODO's on master that don't have an issue number documented.

cron.yaml
ohinds-vera: That was the original intent, but we might have to be aggressive backing off the rate, so I'll put a stop in in case we have to wait really long between requests.

recidiviz/ingest/us_vt/us_vt_record.py
ohinds-vera: In that case, should I even have a UsVtRecord?

recidiviz/ingest/us_vt/us_vt_record.py
ohinds-vera: An additional discussion is how we limit the valid values for a field, if at all. I can see certain cases where we would want this, e.g. race, where ingesting inconsistent identifiers from different systems could impede our ability to do analysis. On the other hand, a lot of the fields here (court type, warrant number, control number, arrest code) could have analogous information that would have completely unrelated values, and it wouldn't make sense to constrain the values.

ohinds-vera: To your original comments, though, I suspect many jail systems will have fields similar to these. I think it would make sense to have separate models for Charge (an offense that is awaiting resolution) and Conviction (an offense for which guilt has been determined). This way we don't have to make an explicit distinction between jail and prison, which are actually more about where someone is held than what they are being held for, while still having the benefit of tracking the differences between these two states and their transition. If that sounds good I can file an issue to  do that generalization.

recidiviz/ingest/us_vt/us_vt_record.py
```diff
@@ -0,0 +1,94 @@
...
+        bond_type: (string) How the bond must be secured.
+        number_of_counts: (integer) Counts in this offense.
+        arresting_agency: (string) Name of the agency that arrested this person
+            on this charge.
+        court_type: (string) Type of court that will hear/heard this case.
+        warrant_number: (string) The warrant number for this charge, if any.
+        court_time: (date) Date of the next court appearance.
+        control_number: (string) Unsure what this is.
+        crime_type: (string) Type of crime.
+        arrest_code: (string) Identifier of the arrest event.
```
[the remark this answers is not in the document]
ohinds-vera: I think it does make sense to have an Arrest model, separate from the Charge model I suggest above. The fields arrest_date, arrest_code, and arresting_agency (we'd also add arresting_officer) would go there. That could be part of the same issue I propose in the previous comment, as well.

recidiviz/ingest/us_vt/us_vt_record.py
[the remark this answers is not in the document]
ohinds-vera: Yes, I think for this case that is fine. I worry that sentences *may* be mutable in situations like appeal or clemency though. And what about the case when a charge is dropped altogether?

recidiviz/ingest/us_vt/us_vt_scraper.py
ohinds-vera: Oh yeah, I think for now I'm tied to this interface. Should I drop a note in #114 pointing out that this class need changing, too?

recidiviz/ingest/us_vt/us_vt_record.py
jessex: That sounds like a good idea. So on `Record` we would generalize `Record.offenses` to `Record.charges` and `Record.convictions`, which should allow `Record` to continue to apply to both jail and prison records? That sounds good. If you file that issue, add a note that we would want to write a short migration that converts all of the us-ny `Record.offenses` fields to `Record.convictions`.

jessex: On your former point, yes I do think we'll need some lightweight, per-field validation system. Generally speaking, there are:
* Some fields where we can define a jurisdiction-agnostic validation standard (e.g. we know that birthdates should not be in the future), 
* Some fields where we could generalize towards a jurisdiction-agnostic validation from jurisdiction-varying fields (e.g. different nomenclature for race among different regions that can be generalized to one canonical list with some effort), and 
* Some fields where we can define a jurisdiction-specific validation standard (e.g. there's a hard-coded list of offense classes or security levels).

I think a validation layer could eventually handle all three, but I would argue that those are in descending order of effort and we should start easy. If you agree, I can file an issue for producing an ingest validation design doc. Could be a good task for one of our upcoming fellows!

recidiviz/ingest/us_vt/us_vt_record.py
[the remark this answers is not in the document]
jessex: Agreed! Let's include it in that issue, too. Do you foresee including `arrests` on `Record` alongside `charges` and `convictions`? It makes sense to my more simple understanding of the criminal justice pipeline.

recidiviz/ingest/us_vt/us_vt_record.py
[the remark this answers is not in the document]
jessex: Yes, I think that we are about to see the need for handling of nested mutability become more pervasive, per some of the comment threads on this PR. It sounds like we agree that any nested fields that can change from snapshot to snapshot should be handled as we would a flat field, with only those nested fields being changed instead of the entire nesting object.

recidiviz/ingest/us_vt/us_vt_scraper.py
jessex: Sure, that sounds good. Thanks!

recidiviz/ingest/us_vt/us_vt_scraper.py
jessex: You (I) learn something new every day.

recidiviz/ingest/us_vt/us_vt_scraper.py
ohinds-vera: I think it's fine, but I still like the idea of queuing the roster entries first anyhow, so I'm switching it.

recidiviz/ingest/us_vt/us_vt_scraper.py
ohinds-vera: AFAIK there will only be one facility and/or one supervision agency. I added a clarifying comment.

ohinds-vera: I'm reconsidering this. Perhaps in cases like this where we know assumptions are being made that will silently discard data if not met (more than one facility, for example), we should test for those conditions explicitly and fail so we catch it (here test that there is no more than a single entry in the facilities or cs_facilities arrays). Thoughts?

recidiviz/ingest/us_vt/us_vt_scraper.py
ohinds-vera: This one is an implementation of an abstract method in Scraper. I added a comment.

recidiviz/ingest/us_vt/us_vt_scraper.py
ohinds-vera: I like it.

recidiviz/ingest/us_vt/us_vt_scraper.py
ohinds-vera: I will file an issue and reference it here. There's going to be a little too much extra work in  util/regions.py to get the snapshot class referenced in this function for me to feel comfortable doing this in this PR.

recidiviz/ingest/us_vt/__init__.py
```diff
@@ -16,6 +16,14 @@
 # =============================================================================
 
 """us_vt package
+
+Vermont is has a unified jail and prison system, so this package
```
jessex: "Vermont has a..."

recidiviz/ingest/us_vt/us_vt_scraper.py
```diff
@@ -155,36 +158,33 @@ def scrape_roster(self, params):
...
 
-        logging.info(roster_response.content)
         roster = json.loads(roster_response.content)
 
+        # add a task for each person in the roster
+        for person in roster['data']:
+            next_person_params = {
+                'session': params['session'],
+                'roster_entry': person,
+                'scrape_type': params['scrape_type']}
```
jessex: Nit: drop the closing curly brace to the next line.

recidiviz/ingest/us_vt/us_vt_scraper.py
```diff
@@ -0,0 +1,614 @@
...
+                            offense_changed = True
+
+                    # Check if any old offenses have been removed
+                    for charge in old_record.offense:
+                        if charge in snapshot.offense:
+                            pass
+                        else:
+                            offense_changed = True
+
+                    if offense_changed:
```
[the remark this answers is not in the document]
jessex: I don't think there's any impact on migration since the migration will be complete prior to this being run in production. The potential impact is if we run this in production and generate Vermont snapshots that capture single-field changes to an offense by copying the entire offense into the new snapshot. Even then, I don't think there would be a meaningful impact on calculation yet that I can foresee. So I think we can defer for now. File an issue for that one and we'll get it prioritized.

recidiviz/ingest/us_vt/us_vt_record.py
jessex: Also, we already have https://github.com/Recidiviz/recidiviz-data/issues/44 filed, which could be repurposed for this new validation issues, or is maybe just a piece of it if our vision is more expansive.

recidiviz/ingest/us_vt/us_vt_scraper.py
```diff
@@ -553,24 +560,35 @@ def compare_and_set_snapshot(self, old_record, snapshot):
...
+
+        skip_attrs = ['class', 'created_on']
+        nested_attrs = ['offense']
         if last_snapshot:
             for attribute in snapshot_attrs:
-                if attribute not in ["class", "created_on", "offense"]:
+                if attribute in skip_attrs:
+                    continue
+                elif attribute in nested_attrs:
+                    set_new_attributes(
```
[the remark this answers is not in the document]
ohinds-vera: Sorry, this was just me messing around with ideas, and wasn't meant to be reviewed.

recidiviz/ingest/us_vt/us_vt_scraper.py
ohinds-vera: Yeah, typo. Whoops!

recidiviz/ingest/us_vt/us_vt_record.py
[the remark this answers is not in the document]
ohinds-vera: Yup! Filed #123

recidiviz/ingest/us_vt/us_vt_record.py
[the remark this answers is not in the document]
ohinds-vera: Yes, arrests should be at that same level. Filed #124

jessex:
I clicked through to those commits so I could see their changes in isolation and left a handful of comments--mostly in the auto-collapsed `us_vt_scraper.py` FYI--and am surprised that they didn't show up in the PR view. Weird, but that's where my latest comments are.

Almost there!

recidiviz/ingest/us_vt/us_vt_scraper.py
```diff
@@ -271,128 +278,182 @@ def scrape_charges(self, params):
...
+                if is_parole_agency(agency_name):
+                    # repeated names happen
+                    if agency_name == parole_agency:
+                        continue
+
+                    # protect against the so-far-unencountered
+                    # situation of multiple distinct community
+                    # supervision agencies.
+                    if parole_agency is not None:
+                        logging.error("Error: found mulitple community "
```
jessex: Instead of logging these three on separate lines without enough context, let's log this as one statement:
`logging.error("Error: found multiple community super vision agencies for person id [%s]: [%s, %s]", person_id, agency_name, parole_agency)`

recidiviz/ingest/us_vt/us_vt_scraper.py
```diff
@@ -271,128 +278,182 @@ def scrape_charges(self, params):
...
+                        logging.error(parole_agency)
+                        return None
+
+                    parole_agency = agency_name
+
+                else:
+                    # protect against the so-far-unencountered
+                    # situation of multiple residential facilities.
+                    if facility is not None:
+                        logging.error("Error: found mulitple residential "
```
jessex: General nit related to the above comment: let's drop prefixes like "Error" from the front of log messages. It's unnecessary.

jessex: Also, can we expand this log message similar to the one in my comment above?

recidiviz/ingest/us_vt/us_vt_scraper.py
```diff
@@ -416,70 +477,101 @@ def field_val_pairs_to_dict(field_val_pair):
...
+        charge_data = charge_data['data']
+
+        # So far, all case data has been empty. Check for non-empty
+        # case data and fail so we know to investigate what should be
+        # stored.
+        case_data = case_data['data']
+        if case_data:
+            logging.error("Non-empty case data found, inspect it and "
+                          "investigate storing the info it contains.")
+            logging.error(case_data)
```
jessex: One of the subtle issues about this pattern of multiple consecutive log statements is that some of the unseen operational tooling, e.g. around log correlation or the use of injected metadata context, is made more difficult when one loggable event it split into multiple logged events. There's also the potential for timing weirdness if the two logs get split unluckily on other side of a time range bound in some search, or for interleaving in a multi-threaded or multi-process context to cause the two statements to actually materialize quite far apart. This all makes troubleshooting more frictiony than it needs to be.

So as a rule of thumb, we should make sure all of the events we log are encapsulated in one log statement. We can use formatting if having it print on multiple lines is beneficial.

PR #118, opened 2018-09-07
Data migration: second pass

### LINKED ISSUE #63: Migrate any non-NY-specific fields to top-level models
Our approach to date has been to put very common fields into top-level models (`Record`, `Inmate`, and `Snapshot`), and fields that are less common / seem specific to New York to the region-specific subclasses.

Instead, we should have any field that is not explicitly region-specific be placed in the top-level model and left as null for regions which don't provide it.

Fields to migrate from `UsNyRecord` and `UsNySnapshot` to `Record` and `Snapshot`:
```
    last_custody_date
    admission_type
    county_of_commit
    custody_status
    earliest_release_date
    earliest_release_type
    parole_hearing_date
    parole_hearing_type
    parole_elig_date
    cond_release_date
    max_expir_date
    max_expir_date_parole
    max_expir_date_superv
    parole_discharge_date
```

In the future, only truly state-specific fields should live in the region-specific subclasses for these models.

jessex:
See #117 for more details on the roll out plan. This is being issued now against the `63-migration` branch for pre-approval but will be reissued against master and merged straight away when time to run the second pass.


jessex:
That build failed for pylint reasons that are resolved on the branch this is branched from. Not sure why that slipped past Travis but it's not an issue locally.

jessex:
This is replaced by #125.

PR #119, opened 2018-09-08
Data migration: cleanup after the second pass

jessex:
This is further clean up of decommissioned code and terminology to be merged after the data migration is complete. It's chiefly focused on removal of `Inmate` and friends, but do let me know if you think I've missed anything else.

This is issued against `63-migration-second-pass` for pre-approval but will be reissued against master and merged straight away after the migration is complete.


jessex:
This is closed in favor of #127.

Issue #126, opened 2018-09-23
Design way to ensure all possible fields are set on each instance of Person, Record, and Snapshot


jessex:
Our hierarchical object model supports the diversity of fields from incoming data streams, but it makes no attempt at validating that all fields that a stream can set are actually being set. This means it is remarkably easy for code for an earlier scraper, for example, to drift as we make changes to the data model. It is not as simple as validating that all fields are set on construction, due to the aforementioned variety. But we do want to know that each ingest integration is setting everything that it can. This could be accomplished through a combination of the to-be-designed validation layer (#44) and unit tests baked into the build.

jessex:
The generic scraper PR (#147) centralized entity creation logic which goes a long way towards solving this problem. However, it could still grow stale if engineers update the fields on core entities without updating the methods on generic scraper.

In this comment thread (https://github.com/Recidiviz/recidiviz-data/pull/147#discussion_r226818174), we discussed creating a lightweight generator that could be run to automatically fill in the required field setters based on what is currently in the entities. However, there is also a need to create a unit test case for generic scraper which explicitly cross-references the list of attributes on the core entities with the list of setters in the generic scraper to ensure that all are included, failing the build if not. We can write this test case independently of the work to create the generator (#153).

arian487:
Obsolete, most of this has been fixed already with the converter/schema work.

PR #127, opened 2018-09-23
Performs some cleanup after the big data migration

jessex:
This does the following:
* Removes nearly all references to "inmate" in favor of "person"
* Ensures that the NY and VT scrapers are setting all of the fields they can set, with the right names
* Reorders the Snapshot and Record entities, and calls to their constructors, to order mass field settings alphabetically
* Removes migration-specific configuration and modeling no longer needed


jessex:
@ohinds-vera I'd like for you to take a look at the changes I made in Vermont to make sure it looks right by you. 

Also, I have a question after having looked through everything post-migration: I noticed that the `UsVtSnapshot` has 6 fields, and those same fields are promoted to the top-level `Record` but not to the top-level `Snapshot`. I don't totally follow the reasoning behind that. If they're not top-level fields then they probably shouldn't be on either `Record` or `Snapshot`, no? (Also, a minor inconsistency here, as of those 6 fields, `earliest_release_date` actually _is_ on `Snapshot`.

jessex:
Also, @ohinds-vera, some thoughts after running the Vermont scraper for the first time locally (against a post-migration Datastore):

I just ran the Vermont scraper myself and noticed that the persisted records and snapshots have some fields that are not set that I think could probably be derived from the other fields I see being set, such as `latest_facility` and `custody_status` and `is_released`. This relates to the issue I recently filed, #126.

Related to this: I noticed that on consecutive runs when we scrape the same record again, `compare_and_set_snapshot` will think that a field which used to be set has become `None` if we set a field in a record we instantiate and mistakenly don't also set it in the snapshot we instantiate. So it's very important that we set everything in both places. We should try to get copying from record to snapshot into one centralized location where we centrally control the copy operation. (I think we should file an issue for centralizing the logic for copying a `Record` to a `Snapshot`.)

Furthermore, after running for a while I started seeing errors like this in the logs:
```
webapp2/_internal_error : 'data'
Traceback (most recent call last):
  File "/Users/clementinejacoby/google-cloud-sdk/platform/google_appengine/lib/webapp2-2.5.2/webapp2.py", line 1535, in __call__
    rv = self.handle_exception(request, response, e)
.....
  File "/Users/clementinejacoby/Documents/projects/recidiviz-data/recidiviz/utils/auth.py", line 110, in auth_and_call
    return func(request_handler, *args, **kwargs)
  File "/Users/clementinejacoby/Documents/projects/recidiviz-data/recidiviz/ingest/worker.py", line 90, in post
    result = scraper_task(params)
  File "/Users/clementinejacoby/Documents/projects/recidiviz-data/recidiviz/ingest/us_vt/us_vt_scraper.py", line 278, in scrape_charges
    params['cases'], charges)
  File "/Users/clementinejacoby/Documents/projects/recidiviz-data/recidiviz/ingest/us_vt/us_vt_scraper.py", line 541, in store_record
    agencies = UsVtScraper.extract_agencies(person_data['data'],
KeyError: 'data'
```

These were caused by the fact that the built up dict in memory looked like this (censored):
```
{"person": {"error": "max-requests-for-timeperiod"}, "cases": {"totalCount": 0, "data": [], "success": "true", "error": ""}, "session": "XXX", "roster_entry": {"Jacket": "XXX", "MiddleName": "X", "FinalReleaseDateTime": null, "Suffix": "", "FirstName": "XXX", "OriginalBookDateTime": "XXX", "RowIndex": XX, "LastName": "XXX", "ArrestNo": XXX, "AgencyName": "XXX"}, "scrape_type": "background"}
```

Have you run into that yourself? This is because right now, `scraper.fetch_page` only denotes error if there's an exception explicitly thrown from the http request, but not if, say, we receive a 400- or 500-space error response, much less if someone misuses the protocol and returns, say, a 200/OK with a body that includes an error message. (I think we should file an issue to update fetch_page to handle non-200 responses like errors, but we should also test ramping up the Vermont scraper and ensuring we both a) calibrate to a reasonable scraping rate to avoid those kinds of errors, and b) gracefully handle those errors.)

If you want to move this conversation out somewhere more appropriate, please do! I'm typing here because I was testing Vermont within the context of my post-migration cleanup.

recidiviz/ingest/us_vt/us_vt_scraper.py
```diff
@@ -134,7 +134,7 @@ def get_session_key(html_tree):
         first_roster_params = {
             'session': session,
             'start': 0,
-            'limit': 500,
+            'limit': 10,
```
jessex: @ohinds-vera: speaking of errors about exceeding request rates, I found that bringing this number down to 10 seemed to make for a stable scraping session. I don't know precisely what the impact is on performance, but I suspect it's not terribly high, since it just means more `scrape_roster` requests and no impact on `scrape_person` or further. It also means that we queue up significantly fewer tasks before we get to a wave of `scrape_charges` calls where we finally persist data, meaning lower latency and less data "lost" in the event of a failure which shuts down the entire session.

recidiviz/ingest/us_vt/us_vt_snapshot.py
```diff
@@ -25,9 +25,9 @@
 class UsVtSnapshot(Snapshot):
     """A subclass of Snapshot that adds New York specific fields
```
ohinds-vera: New York -> Vermont

recidiviz/ingest/us_vt/us_vt_snapshot.py
```diff
@@ -25,9 +25,9 @@
...
 
-    Datastore model for the New York-specific fields in snapshots of inmate
+    Datastore model for the Vermont--specific fields in snapshots of person
```
ohinds-vera: "Vermont--specific" --> "Vermont-specific"

recidiviz/ingest/us_vt/us_vt_scraper.py
jessex: Let's go with this for now. We'll want to tweak after running in production.

recidiviz/ingest/README.md
```diff
@@ -88,7 +88,7 @@ is considered a 'scrape session'.
 More details are provided below for components of this system.
 
 #### Regional variation
-Every corrections system provides a different inmate search system, and so requires a custom scraper to collect and
+Every corrections system provides a different person search system, and so requires a custom scraper to collect and
```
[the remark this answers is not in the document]
jessex: Sounds good.

recidiviz/ingest/docket.py
```diff
@@ -502,26 +502,26 @@ def add_to_query_docket(scrape_key, docket_items):
     except (taskqueue.taskqueue.DuplicateTaskNameError,
             taskqueue.taskqueue.TombstonedTaskError,
             taskqueue.taskqueue.TaskAlreadyExistsError):
-        logging.debug("Some inmates been added to the docket already; "
+        logging.debug("Some people been added to the docket already; "
```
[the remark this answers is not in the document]
jessex: Yes, thanks!

recidiviz/ingest/us_ny/us_ny_scraper.py
```diff
@@ -716,155 +714,152 @@ def store_record(self, inmate_details):
         that can be indexed / queried in the future.
 
         Args:
-            inmate_details: (dict) Key/value results parsed from the scrape
+            person_details: (dict) Key/value results parsed from the scrape
 
         Returns:
             Nothing if successful, -1 if fails.
 
         """
```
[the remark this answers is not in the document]
jessex: Done! Todo added above.

PR #134, opened 2018-09-27
Improve ingest test coverage

### LINKED ISSUE #89: Clean up ingest test coverage

jessex:
This resolves #89.

The increase in coverage is actually far higher than the Coveralls report implies. I noticed we were accidentally counting the test code itself towards our code coverage. It was all at or near 100% and artificially inflating our coverage stats. The change to `.coveragerc` in here keeps us honest on that front. Getting rid of the test code from coverage stats dropped us to down around 50%, and this PR climbs back out of that hole.

The only untested ingest code at this point is the New York scraper, which I'd like to save for a follow-up PR:
* This is already an enormous PR with 3000 lines of test code
* The crowning jewel here is the testing of the Vermont scraper, but clearly if all scrapers require a 1600 line test class then we'll have some people scale issues. So I'd like to get people reviewing Vermont while I start working on trying to generalize that work to New York as part of some sort of scraper test suite that reduces required effort


recidiviz/tests/ingest/us_vt/us_vt_scraper_test.py
```diff
@@ -0,0 +1,1671 @@
...
+                "Value": "ST. ALBANS PROBATION & PAROLE"
+            },
+            {
+                "Field": None,
+                "Value": "A DIFFERENT PAROLE AGENCY"
+            }
+        ]
+
+        agencies = self.conduct_test(person_json)
+        assert agencies is None
```
ohinds-vera: Is there a way to test that a logging.error call occurs also?

recidiviz/tests/ingest/us_vt/us_vt_scraper_test.py
```diff
@@ -0,0 +1,1671 @@
...
+                "Value": "ST. ALBANS PROBATION & PAROLE"
+            },
+            {
+                "Field": None,
+                "Value": "A DIFFERENT FACILITY"
+            }
+        ]
+
+        agencies = self.conduct_test(person_json)
+        assert agencies is None
```
ohinds-vera: Ditto on the logging.error

recidiviz/ingest/us_vt/us_vt_scraper.py
```diff
@@ -36,7 +36,7 @@
 import re
 
 from lxml import html
-from lxml.etree import XMLSyntaxError # pylint:disable=no-name-in-module
+from lxml.etree import LxmlError  # pylint: disable=no-name-in-module
```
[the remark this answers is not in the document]
jessex: I did and could not for the life of me figure out why it's panicking. I tried a variety of import paths and forms, and my IDE and python path see the package, file, and class. 
❓ ❓

recidiviz/tests/ingest/us_vt/us_vt_scraper_test.py
jessex: You can do it by mocking out the calls to logging, but I don't think it's worth the effort, personally. We know the logging statement was hit by examining the code coverage. Mocking it out and asserting its arguments lets us check in fine detail that we logged a particular statement, but I don't honestly think that's worthwhile. Let me know if you disagree.

recidiviz/ingest/scraper_utils.py
```diff
@@ -63,7 +63,7 @@ def parse_date_string(date_string, person_id):
             return None
 
         # If month-only date, manually force date to first of the month.
-        if len(date_string.split("/")) == 2:
+        if len(date_string.split("/")) == 2 or len(date_string.split("-")) == 2:
```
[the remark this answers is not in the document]
jessex: The day becomes the first of the month if it is not specified in the string.

I had the same discordant reaction to you when I first saw this method--that there's a sudden shift from general to specific--but the reason it exists is because the date string parser we are using doesn't appear to cover this case and it's a known data case we have encountered. I think the right long-term solution is to extract this into a library somewhere (either find a library that handles it, contribute to a popular date string parsing library, or maintain our own shim) but I think that's out of scope for this PR.

Issue #123, opened 2018-09-19
Snapshots should only save nested fields if they have changed since the last scrape


ohinds-vera:
Currently, `compare_and_set_snapshot` correctly computes differences between snapshot and an old record for scalar fields only. It should also store only fields that have changed between scrapes in more complicated fields such as objects, sequences, and sequences of objects. For example, currently the Record object has a list of 'Offense' objects, and `compare_and_set_snapshot` stores the entire list of offenses if any field in any offense has changed since the last scrape.

ageiduschek:
When investigating this issue I realized that the approach outlined in this task doesn't handle what might happen if an offense is deleted.

From discussion with @jessex:

> We definitely want to track that the removal occurred. Expunging of an offense is one possibility, and there could be sentencing-related ones in the future when we start to ingest systems that separate out, for example, concurrent sentences. As it stands, there is no strict use case for doing any processing with the knowledge that a sub-entity array had an element removed right now, but it's almost certain to come up at some point, and when new ways of looking at criminal justice records come up they tend to be surprising. So we should think about this one now.

> The latest and greatest state of the data is recorded via the top-level Record, but we do want to be able to compare a series of Snapshots to see that some sub-entity was removed.
> * Storing a different sub-entity on the Snapshot (i.e. instead of Snapshot.offenses to match Record.offenses, we'd have Snapshot.offense_changes) that includes a flag like `is_deletion` is technically possible but messy, and makes conversion between Records and Snapshots more difficult because of an extra, nested layer of transformation to perform.
> * A similar approach would be to continue to track Snapshot.offenses but also track Snapshot.offense_removals, which would be an array of removed offenses. It would be easy to work with in the future by doing a simple check on whether Snapshot.offense_removals is empty. Conversion between Records and Snapshots stays straight-forward, but we have the added complexity of including an extra field on the Snapshot for any array sub-entity where removal is possible. 
> * Or, we scrap this plan and just resolve to always store an entire field whenever it changes, where it's a flat field, an array field, or an object field. That way, evaluation of a series of consecutive snapshots is predictable for all fields (compare the entire field in each snapshot, no matter what). It's more consistent from the perspective of the Snapshot field, but less consistent from the perspective of the lowest-level field available, i.e. sub-fields. It has less complexity overall than the other two, but inflates the amount of data we need to store and process as we navigate the object graph in some cases.

We decided that the best plan of action was to do nothing, since it's the simplest and least error-prone right now.

One idea though:
[...]

PR #141, opened 2018-10-15
Move compare_and_set_snapshot out of Scraper subclasses (#121)

### LINKED ISSUE #121: Move compare_and_set_snapshot() into the parent Scraper class
Right now, all child classes have to implement this function, which is only different because of the name of the child snapshot class that must be referenced. This will involve adding functionality to utils/regions.py to retrieve a region's snapshot class.

ageiduschek:
Fixes #121 - Moves some duplicated code out of Snapshot base classes.

Note: When running the us_ny scraper locally on I'm seeing a ton of seemingly unrelated exceptions, but this change doesn't seem to have broken things more...


jessex:
I'm pretty sure I know what the exceptions you saw locally are but if you send them over to me I can confirm. A branch I have in progress fixes them in passing if they're what I think they are. If so, I'll issue a quick PR with just that fix.

recidiviz/ingest/scraper.py
```diff
@@ -85,6 +88,21 @@ def person_id_to_record_id(self, person_id):
         """
         pass
 
+
+    @abc.abstractmethod
+    def get_snapshot_class_name(self):
```
jessex: We can remove this in favor of `Region.get_snapshot_kind()`, which we can call directly from `compare_and_set_snapshot`.

recidiviz/ingest/scraper.py
```diff
@@ -334,3 +352,90 @@ def iterate_docket_item(self, scrape_type):
...
+                current scrape.
+        Returns:
+            True if successful
+            False if datastore errors
+
+        """
+        new_snapshot = False
+
+        # pylint:disable=protected-access
+        snapshot_class = ndb.Model._kind_map[self.get_snapshot_class()]
```
jessex: Once we remove this in favor of `Region.get_snapshot_kind`, we should be able to annotate this method as a `@staticmethod`.

recidiviz/ingest/scraper.py
```diff
@@ -334,3 +352,90 @@ def iterate_docket_item(self, scrape_type):
...
+                        setattr(snapshot, attribute, None)
+
+                elif attribute == "offense":
+                    # Offenses have to be treated differently -
+                    # setting them to None means an empty list or
+                    # tuple instead of None, and an unordered list of
+                    # them can't be compared to another unordered list
+                    # of them.
+                    #
+                    # TODO #123 instead of storing the entire list if
```
jessex: If we're going to _not_ do #123 as discussed offline, would you mind removing this TODO and closing that issue with a comment that explains why?

recidiviz/ingest/scraper.py
jessex: Never mind. As discussed offline, you're right. We'd still be accessing `self.region` here.

recidiviz/ingest/us_ny/us_ny_scraper.py
```diff
@@ -911,7 +911,7 @@ def store_record(self, person_details):
             record.offense = record_offenses
         record.race = person.race
         record.record_id = record_id
-        record.region = self.get_region()
+        record.region = self.get_region().region_code
```
jessex: Thanks for tossing in this fix, @ageiduschek! I'll close #143.

ageiduschek: Oh whooops I thought I had reverted this. Somehow that didn't make it up here...

PR #143, opened 2018-10-16
Fixes typing issue around Region versus String region code in the NY 

jessex:
@ageiduschek: is this the same issue that you saw locally?

Good news: my unit test coverage in progress for this scraper caught this issue, so we're on our way towards fewer hapless bugs in the very near future.


jessex:
Fixed in #141.

Issue #137, opened 2018-10-09
Document data transfer spec


jessex:
This is a bit of a placeholder as we review the data model and ingest interface. Once we are confident in an ingest API, as a more general abstraction above the scraper API, and confident in the underlying data model--importantly including how that model will evolve to accommodate more jurisdictions--we need to draft publish a versioned data transfer spec. Exact medium to be specified later.

jessex:
This should also include a glossary of the terminology we use and why, notably including why we prefer some terms over others, such as _person_ instead of _inmate_, _convict_, or the like.

arian487:
Most of this work is complete with the ingest_info work, we just need to document the possible fields.

jessex:
Going to re-open this because I don't want to lose track of very important documentation work. Please re-close this if you've got a dedicated documentation issue you'd like to use instead.

ohinds:
@jessex, does the existence of this doc satisfy this issue?

https://docs.google.com/document/d/16j-Ng8c4MGyKArDrTsyLjQ-MQfYbxTsoWWmZBBLG0iA/edit#

jessex:
I think that document needs to be public to satisfy this issue. It could be in a README or we could turn on the Wiki feature in this repository (and eventually, it should maybe live in a subdomain on recidiviz.org somewhere), but I think we need a stable, versioned, public spec.

jessex:
Speaking of which: #467. I think we can close this issue with the addition of those docs, as long as they are clearly related to the API.

PR #147, opened 2018-10-18
Add generalized scraper and usage example from pa greene

arian487:
This code attempts to generalize the scraper code into a single place.  It abstracts away the need for any child scrapers to know about how the DB entities are organized.  If in the future we have 500 scrapers, and we change a db entity name or structure it might be a nightmare to propagate that out to all the scrapers.

The individual scrapers need only care about 2 things:

1.  Given the current page, what is the next set of pages I want to scrape, and are there people/records in that page.
2.  Given the current page, how do I extract out information I care about? (person_id, age, etc...)

Child scrapers should NOT know about how scraping is set up from a task perspective, or any generalized structure before a change in that logic might have devastating refactor consequences once we have more scrapers.  

The eventual goal is to move as many of the getters to the top level generic scraper as possible.  Can the generic scraper scrape out age from **any** website given a config mapping.  I think this is the end state we want and I have an idea how it might be possible, but this is the first step in allowing that to happen.  

This also helps in granularly being able to see which piece of data scraping caused something to fail and will make our monitoring easier to read at the dashboard level (when they eventually exist)


recidiviz/ingest/us_pa_greene/us_pa_greene_person.py
```diff
@@ -0,0 +1,41 @@
...
+    Attributes:
+
+        us_pa_greene_person_id: (string) Same as person_id, but used as
+            key for this entity type to force uniqueness / prevent
+            collisions within the us_vt records (see
+            models.person for inherited attributes)
+
+    """
+    us_pa_greene_person_id = ndb.StringProperty()
+    us_pa_greene_inmate_id = ndb.StringProperty()
```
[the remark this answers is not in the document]
jessex: That's right, we _should_ have changed all references to _inmate_ to _person_ but we may have missed some which would be misleading.

recidiviz/ingest/us_pa_greene/us_pa_greene_record.py
```diff
@@ -0,0 +1,50 @@
...
+class UsPaGreeneRecord(Record):
+    """A subclass of Record that adds PA Greene specific fields
+
+    Datastore model for a specific person incarcerated in
+    Vermont. This extends the Person class.
+
+    Attributes:
+		reference_id: (string) The reference id of the record, this is different
+			from booking number which we map to record_id
+        committed_by: (string) The entity that committed the inmate.
```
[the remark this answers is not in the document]
jessex: Yeah it wouldn't be obvious that this is a conscious decision or why, especially when so many of the source data systems we scrape will use that terminology. It would be good to document the terminology we use and why. I've added a note to #137 to include that.

recidiviz/ingest/us_pa_greene/us_pa_greene_scraper.py
```diff
@@ -0,0 +1,384 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+# =============================================================================
+
+
+"""Scraper implementation for PA Greene County (Archonix)
```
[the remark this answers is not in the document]
jessex: I'm in favor, assuming there's a wide portfolio of Archonix jurisdictions.

recidiviz/ingest/scraper_utils.py
```diff
@@ -138,6 +142,19 @@ def calculate_age(birth_date):
     return age
 
 
+def get_id_value_from_html_tree(html_tree, id):
```
jessex: Looking again at this file, I'm seeing that it's going to become a bit monolithic and is ripe to be split into more focused helper classes. We could put this method and others like it into a module focused on common scraping patterns, e.g. `selectors.py`, whereas the two directly below could be put into a module focused on retrieving values from environment variables.

recidiviz/ingest/us_pa_greene/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+# =============================================================================
+
+
+"""us_ny specific regional functionality.
+"""
+
+
+from recidiviz.ingest.us_pa_greene import us_pa_greene_scraper
```
jessex: Cleanup reminder: newlines erry'where.

recidiviz/ingest/us_pa_greene/us_pa_greene_person.py
```diff
@@ -0,0 +1,41 @@
...
+"""Person model for PA Greene Country Jail.
+"""
+
+
+from google.appengine.ext import ndb
+from recidiviz.models.person import Person
+
+
+class UsPaGreenePerson(Person):
+    """A subclass of Person that adds Vermont specific fields
```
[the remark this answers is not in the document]
jessex: Cleanup reminder: rename all copy+pasted references to Vermont from Greene County.

recidiviz/ingest/us_pa_greene/us_pa_greene_person.py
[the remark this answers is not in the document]
jessex: I am curious here, though, if this was meant to track two separate ids/entities?

recidiviz/ingest/us_pa_greene/us_pa_greene_record.py
```diff
@@ -0,0 +1,50 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+# =============================================================================
+
+"""Record model for PA Greene Country Jail.
```
jessex: County

recidiviz/ingest/us_pa_greene/us_pa_greene_record.py
```diff
@@ -0,0 +1,50 @@
...
+    Vermont. This extends the Person class.
+
+    Attributes:
+		reference_id: (string) The reference id of the record, this is different
+			from booking number which we map to record_id
+        committed_by: (string) The entity that committed the inmate.
+
+    """
+    reference_id = ndb.StringProperty()
+    committed_by = ndb.StringProperty()
```
jessex: Pursuant to our conversation on Thursday at GWC, `committed_by` sounds exactly like a field to live at the top-level of the Record hierarchy.

recidiviz/ingest/us_pa_greene/us_pa_greene_record.py
```diff
@@ -0,0 +1,50 @@
...
+class UsPaGreeneOffense(Offense):
+    """UsPaGreene-specific model to describe a particular crime of conviction
+
+    Attributes:
+        case_number: (string) Case number of this offense.
+        bond_amount: (float) Dollar amount of bond on this charge, if any.
+
+    """
+    case_number = ndb.StringProperty()
+    bond_amount = ndb.FloatProperty()
```
jessex: I think both of these fields should live at the top of the Offense hierarchy.

jessex: Idle thought: I think that all of the comments related to where a field should live in the hierarchy will feel jarring at first but I do think that as we expand our coverage, we will reach a point where new jurisdictions do not actually introduce anything new and we've hit on everything we want to track.

region_manifest.yaml
```diff
@@ -46,3 +46,17 @@ regions:
       snapshot: UsVtSnapshot
     scraper_package: us_vt
     timezone: America/New_York
+  us_pa_greene:
+    region_name: UsPaGreene
+    agency_name: Greene County Detention Center
+    region_code: us_pa_greene
+    agency_type: unified
+    queues:
```
jessex: When you rebase you'll see that this is now a singular `queue: us-pa-greene-scraper`.

region_manifest.yaml
```diff
@@ -46,3 +46,17 @@ regions:
...
+    - us-pa-green-scraper
```
jessex: greene, here and in the names file below.

recidiviz/ingest/generic_scraper.py
```diff
@@ -0,0 +1,620 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+# =============================================================================
+
+
+"""Generic implementation of the scraper.  This class abstracts away the need for
+scraper implementations to know or understand tasks.  Child scrapers need only
+to know how to extract data from a given page and whether or not more pages need
+to be scraped.
```
jessex: I think we'll want some more information in here that explains a) its high-level structure, and b) the rough work required to extend this for a specific jurisdiction.

recidiviz/ingest/generic_scraper.py
```diff
@@ -0,0 +1,620 @@
...
+        person.sex = self.get_sex(content, params)
+        person.race = self.get_race(content, params)
+        # Lastly populate any extra params that may be scraper specific.
+        self.populate_extra_person_params(content, params, person)
+
+        try:
+            person.put()
+        except (Timeout, TransactionFailedError, InternalError):
+            # Datastore error - fail task to trigger queue retry + backoff
+            logging.warning("Couldn't persist person: %s", person.person_id)
```
jessex: We should expose some information from the exception on this one instead of swallowing it all. Same goes for the other catches in this class.

recidiviz/ingest/generic_scraper.py
```diff
@@ -0,0 +1,620 @@
...
+            return -1
+        return person
+
+    def _populate_record(self, content, params):
```
jessex: The centralization of entity construction in here is such a huge improvement, in terms of readability, maintainability, operability, every ability. :100:

recidiviz/ingest/generic_scraper.py
```diff
@@ -0,0 +1,620 @@
...
+        record.latest_release_date = self.get_latest_release_date(content, params)
+        record.latest_release_type = self.get_latest_release_type(content, params)
+        record.max_expir_date = self.get_max_expir_date(content, params)
+        record.max_expir_date_parole = self.max_expir_date_parole(content, params)
+        record.max_expir_date_superv = self.max_expir_date_superv(content, params)
+        record.max_sentence_length = self.get_max_sentence_length(content, params)
+        record.min_sentence_length = self.get_min_sentence_length(content, params)
+        record.offense = self.get_offenses(content, params)
+        record.offense_date = self.get_offense_date(content, params)
+        recoprd.parole_discharge_date = self.get_parole_discharge_date(content, params)
```
jessex: Whoopsy-daisy. `record`

recidiviz/ingest/generic_scraper.py
```diff
@@ -0,0 +1,620 @@
...
+        new_snapshot = self.get_snapshot_class()()
+        last_snapshot = self.get_snapshot_class().query(
+            ancestor=record.key).order(
+                -self.get_snapshot_class().created_on).get()
+
+        # We loop through all of the properties of the snapshot class and
+        # Create this snapshot from the scraped record entities.
+        if last_snapshot:
+            for attr in snapshot_attrs:
+                if attr not in ['class', 'created_on', 'offense']:
```
jessex: As @ageiduschek has pointed out, we can get smarter about which fields we choose to handle separately, e.g. instead of special-casing `offense`, we handle all arrays and objects in a particular fashion. I don't think we have an issue filed for that yet. If you'd like to tackle it here then feel free but if not I'm happy to file an issue.

recidiviz/ingest/generic_scraper.py
```diff
@@ -0,0 +1,620 @@
...
+
+        record = get_record_class().query(ancestor=person.key).get()
+        if not record:
+            return None
+
+        return record.record_id
+
+    # Every scraper must implement this.  It should return a list of params
+    # Where each param corresponds to a task we want to run.
+    # Mandatory fields are endpoint and task_type.
```
jessex: Cleanup reminder: shift these comments into the pydoc itself.

recidiviz/ingest/generic_scraper.py
```diff
@@ -0,0 +1,620 @@
...
+        Returns:
+            A hex code from constants telling the generic how to behave on the first call
+        """
+        return constants.INITIAL_TASK_AND_MORE
+
+    # The following can be overriden by the child scrapers to help scrape content
+    # retrieved from an endpoint.
+    # PERSON RELATED GETTERS
+
+    def get_person_class(self):
```
jessex: We should be able to use the `Region.get_person_kind`, `Region.get_record_kind`, and `Region.get_snapshot_kind` methods instead of asking scrapers to define them specifically. Each scraper should be aware of its own region, after all. One approach would be to not have these be overridden generally, but have them call a delegate for retrieving the scraper's region and then checking the kind. 

Something like so:
```
def get_person_class(self):
    return self.get_region().get_person_kind()

def get_record_class(self):
    return self.get_region().get_record_kind()

def get_snapshot_class(self):
    return self.get_region().get_snapshot_kind()
```

recidiviz/ingest/generic_scraper.py
```diff
@@ -0,0 +1,620 @@
...
+                    record_value = getattr(record, attr)
+                    setattr(new_snapshot, attr, record_value)
+
+        if snapshot_changed:
+            try:
+                new_snapshot.put()
+            except (Timeout, TransactionFailedError, InternalError):
+                logging.warning("Couldn't store new snapshot for record %s",
+                                old_record.record_id)
+
```
jessex: When you rebase, you'll see that `compare_and_set_snapshot` now returns `False` inside of the `except` and returns `True` at the end of the method.

recidiviz/ingest/generic_scraper.py
```diff
@@ -0,0 +1,620 @@
...
+    # Every scraper must implement this.  It should return a list of params
+    # Where each param corresponds to a task we want to run.
+    # Mandatory fields are endpoint and task_type.
+    @abc.abstractmethod
+    def get_more_tasks(self, content, params):
+        """
+        Gets more tasks based on the content and params passed in.  This
+        function should determine which task params, if any, should be
+        added to the queue
```
jessex: It's a bit of an annoying comment in some ways, but let's add `Args` statements for all of these methods, even if they feel redundant. If and when we generate browsable pydocs, we'll be glad we did, I think.

recidiviz/ingest/generic_scraper.py
```diff
@@ -0,0 +1,620 @@
...
+
+        Returns:
+            A list of params containing endpoint and task_type at minimum.
```
jessex: An actual list of parameter objects, a list of tuples, a dict?

recidiviz/ingest/generic_scraper.py
```diff
@@ -0,0 +1,620 @@
...
+
+    def _populate_record(self, content, params):
+        """Calls all of the necessary functions to populate a record in the data.
+
+        Returns:
+            A record database object.
+        """
+        person = params['person']
+        record_id = self.get_record_id(content, params)
+        record = self.get_record_class().get_or_insert(record_id, parent=person.key)
```
jessex: Another thing you'll notice on rebase: the Vermont scraper now follows the New York scraper's trail on this one and has a `create_record` method which returns a tuple of `(old_record, record)` where old_record is the result of the get_or_insert call. This is because _that's_ the one you'd want to `deepcopy` when doing `compare_and_get_snapshot`, otherwise you'd be comparing against the thing you just updated anyway and your snapshot would match it every time.

recidiviz/ingest/us_pa_greene/us_pa_greene_scraper.py
```diff
@@ -0,0 +1,384 @@
...
+        return False
+
+    def get_person_id(self, content, params):
+        """Gets person id given a page
+        Returns:
+            A string representing the persons id
+        """
+        return self._get_scraped_value(content, 'ctl00_ContentPlaceHolder1_spnInmateNo')
+
+    def person_id_is_fuzzy(self):
```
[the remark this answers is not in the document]
jessex: Looks like it. Same with `record_id_is_fuzzy`.

recidiviz/ingest/generic_scraper.py
```diff
@@ -0,0 +1,620 @@
...
+    def get_race(self, content, params):
+        """Gets person race given a page
+        Returns:
+            A string representing the persons race
+        """
+        pass
+
+    # RECORD RELATED GETTERS
+
+    def record_id_is_fuzzy(self, content, params):
```
jessex: We've got two `record_id_is_fuzzy` implementations here in generic scraper.

recidiviz/ingest/scraper_utils.py
```diff
@@ -138,6 +142,19 @@ def calculate_age(birth_date):
...
+    """Retrieves the value of the given id from the given html tree.
+
+    Args:
+        html_tree: (string) html of the scraped page.
+        id: (string) the html id we are trying to retrieve.
+
+    Returns:
+        A string representing the value of the id from the html page.
+    """
+    html_tree.cssselect('[id={}]'.format(id))[0].get('value')
```
jessex: We'll need to handle the case when `cssselect` returns an empty list here.

recidiviz/ingest/us_pa_greene/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+# =============================================================================
+
+
+"""us_ny specific regional functionality.
```
jessex: Copy+paste mistake.

region_manifest.yaml
```diff
@@ -46,3 +46,17 @@ regions:
...
+    base_url: http://www.co.greene.pa.us/InmateLocator
+    names_file: us_pa_green_names.csv
```
jessex: Does this exist? It seems like we don't need it for this region.

recidiviz/ingest/us_pa_greene/us_pa_greene_scraper.py
```diff
@@ -0,0 +1,384 @@
...
+        super(UsPaGreeneScraper, self).__init__('us_pa_greene')
+
+    def _retrieve_session_vars(content):
+        """Gets the variables necessary to complete requests.
+
+        Returns:
+            A dict of session vars needed for the next scrape.
+        """
+        data = {}
+        for session_var in self._required_session_vars:
```
jessex: We can one-line this with a dictionary comprehension, if you think it's more readable: 
`{session_var: scraper_utils.get_id_value_from_html_tree(content, session_var) for session_var in self._required_session_vars}`

recidiviz/ingest/us_pa_greene/us_pa_greene_scraper.py
```diff
@@ -0,0 +1,384 @@
...
+                content, session_var)
+        return data
+
+    def _get_all_inmates_params(self, content):
+        """Returns the params needed to search for all inmates.
+
+        Returns:
+            dictionary of data needed for the next scrape session.
+        """
+        params = {}
```
jessex: From a readability standpoint, I've always preferred inlining dict initialization, e.g. 
```
params = {
    'data': self.retrieve_session_vars(content),
    'endpoint': self._initial_endpoint,
    ...
}
```

I also _think_ it's considered the more Pythonic approach, but I could be wrong!

recidiviz/ingest/us_pa_greene/us_pa_greene_record.py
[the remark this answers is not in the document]
jessex: It's worth noting, though, that in cases where inmate is the terminology in a source data system and may even be included as, say, a required key in an API request, we can't avoid its usage. And there's a debate to have about how far we should go in constraining use of that language. Would we, for a contrived example, make an API call that requires us to pass in a field called `inmateName` and returns what it calls an "inmate id" and declare it a `personId` locally in our code? There are cases where it could be tricky to weigh terminology sensitivity and code readability.

recidiviz/ingest/us_pa_greene/us_pa_greene_scraper.py
```diff
@@ -0,0 +1,384 @@
...
+        return params
+
+    def _get_inmate_params(self, content):
+         """Returns the params needed to open the page of a specific inmate.
+
+        Returns:
+            a list of dictionary data needed for the next scrape session.
+        """
+        params_list = []
+        grid_results = content.cssselect('[id=ctl00_ContentPlaceHolder1_gridResults_ctl00]')[0]
```
jessex: This logic is a subset of what we have in `get_id_value_from_html_tree` so there's an extraction we can pull here.

recidiviz/ingest/us_pa_greene/us_pa_greene_scraper.py
```diff
@@ -0,0 +1,384 @@
...
+        table_body = grid_results.find('tbody')
+        for row in table_body:
+            href = row.find('tr').find('a').get('href')
+            endpoint = '/'.join([self._base_endpoint, href])
+            # Extract out the reference and inmate id.
+            ref_id_start = href.index('ReferenceID=') + len('ReferenceID=')
```
jessex: I think this is a supremely useful, extractable utility: parsing ids out of `href` attributes. As we discussed Thursday, there's going to be a whole suite of abstractions we'll identify to form some library around. Definitely don't want to generalize until we have multiple use cases calling for it, but do want to highlight cases I see.

recidiviz/ingest/us_pa_greene/us_pa_greene_scraper.py
```diff
@@ -0,0 +1,384 @@
...
+        return params_list
+
+    def _get_next_page_if_exists(self, content):
+        """Returns the params needed to scrape the next page of results.
+
+        Returns:
+            dictionary of data needed for the next scrape session.
+        """
+        params_list = []
+        current_page = content.cssselect('[class=rgCurrentPage]')[0].text_content()
```
jessex: Cleanup reminder: handling empty lists throughout the code in here.

jessex: Similarly, handling chains like `.find('input').get('name')` below or `.find('tr').find('a').get('href')` above which could return None at any link.

recidiviz/ingest/us_pa_greene/us_pa_greene_scraper.py
```diff
@@ -0,0 +1,384 @@
...
+            params['endpoint'] = self._initial_endpoint
+            params['task_type'] = constants.GET_MORE_TASKS
+            # Effectively this extracts out the next page button to send to the post.
+            params['data']['__EVENTTARGET'] = content.cssselect(
+                'div.rgArrPart2')[0].find('input').get('name')
+            params_list.append(params)
+        return params_list
+
+
+    def get_more_tasks(self, content, params):
```
jessex: The structure in here feels like a good next candidate for further generalization: I can see this become something more declarative where each child scraper just defines the callbacks that determine the path through the task tree, e.g. `INITIAL_TASK -> GET_MORE_TASKS = scraper_utils.get_id_value_from_html_tree(content, self._page_size_id) == '10')`, incorporating the list of params to be found at each new state.

recidiviz/ingest/us_pa_greene/us_pa_greene_scraper.py
```diff
@@ -0,0 +1,384 @@
...
+            params_list.append(self._get_page_size_50_params(content))
+        # We can start sending people to be scraped.
+        elif ((task_type & constants.GET_MORE_TASKS) && (
+            scraper_utils.get_id_value_from_html_tree(content, self._page_size_id) == '50')):
+            # First find all inmates we need to scrape on this page.
+            params_list.extend(self._get_inmate_params(content))
+            # Next check to see if there is another page we can scrape.
+            params_list.extend(self._get_next_page_if_exists(content))
+
+    def _get_scraped_value(self, content, scrape_id):
```
jessex: We should place this alongside `get_id_value_from_html_tree`, probably.

recidiviz/ingest/us_pa_greene/us_pa_greene_scraper.py
```diff
@@ -0,0 +1,384 @@
...
+        """Gets whether or not the person is released.
+        Returns:
+            A bool representing whether the person is released.
+        """
+        # This county only shows record of inmates that are currently in jail, so if we got here
+        # they are not released.
+        return False
+
+    def populate_extra_record_params(self, content, params, record):
+        """Populates any extra params for a person."""
```
jessex: "for a record"

recidiviz/ingest/us_pa_greene/us_pa_greene_scraper.py
```diff
@@ -0,0 +1,384 @@
...
+        return scraper_utils.parse_date_string(date_str[0:10])
+
+    def get_custody_status(self, content, params):
+        """Gets the record custody status given a page.
+        Returns:
+            A string representing the persons custody status.
+        """
+        return self._get_scraped_value(content, 'ctl00_ContentPlaceHolder1_spnCustodyStatus')
+
+    def get_is_released(self, content, params):
```
jessex: @ohinds-vera: I'm guessing that there are other convenience boolean methods that apply to jails generally beyond just `is_released` (which is more prison-centric, probably)? Such as `is_awaiting_trial` or `is_awaiting_transfer`

recidiviz/ingest/us_pa_greene/us_pa_greene_scraper.py
```diff
@@ -0,0 +1,384 @@
...
+            offense.crime_class = row[0].text_content()
+            offense.description = row[1].text_content()
+            offsense.case_number = row[2].text_content()
+            # Note that row[3] and row[4] represent the offense date and arrest date,
+            # But PA Greene leaves this field blank.
+            bond_str = row[5].text_content()
+            # If there is no bound, it is filled with this character.
+            if bond_str != u'\xa0':
+                # Bond is written in this form: $15,000.00
+                offense.bond_amount = float(bond_str[1:].replace(',', ''))
```
jessex: I imagine there's a canonical package for parsing currency text.

recidiviz/ingest/us_pa_greene/us_pa_greene_person.py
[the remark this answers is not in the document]
arian487: Done.

recidiviz/ingest/us_pa_greene/us_pa_greene_person.py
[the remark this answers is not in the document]
arian487: So this is deliberate.  Greene County has 2 IDs for a person.  One is exposed in the actual web page and is called Inmate number, this is what you use to actually search for a person on their page.  Another is an internal database ID used to construct the URL that goes directly to the persons page.  I've labelled the exposed id as person_id but added the field inmate_id to also store that extra, un-exposed ID.  Here is an example:

http://www.co.greene.pa.us/InmateLocator/DetailPages/InmateDetails.aspx?ReferenceID=1306&InmateID=17614

recidiviz/ingest/us_pa_greene/us_pa_greene_record.py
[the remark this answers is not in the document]
arian487: I've changed its usage here.  To the point Josh made, I added an extra field in person called inmate_id as it is an internal DB ID they use that can be uswed to construct the url which directly takes us to a person.  I can change it to something else, but it might be confusing to someone else reading the code which field is represented if there is a mismatch in field.

recidiviz/ingest/us_pa_greene/us_pa_greene_record.py
```diff
@@ -0,0 +1,50 @@
...
+
+class UsPaGreeneRecord(Record):
+    """A subclass of Record that adds PA Greene specific fields
+
+    Datastore model for a specific person incarcerated in
+    Vermont. This extends the Person class.
```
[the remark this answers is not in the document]
arian487: Done.

recidiviz/ingest/us_pa_greene/us_pa_greene_scraper.py
```diff
@@ -0,0 +1,384 @@
...
+            '__EVENTARGUMENT': 'FireCommand:ctl00$ContentPlaceHolder1$gridResults$ctl00;PageSize;50'
+            'ctl00_ContentPlaceHolder1_gridResults_ctl00_ctl03_ctl01_PageSizeComboBox_ClientState':
+                '{"logEntries":[],"value":"50","text":"50","enabled":true,"checkedIndices":[],"checkedItemsTextOverflows":false}'
+
+        }
+        self._page_size_id = 'ctl00_ContentPlaceHolder1_gridResults_ctl00_ctl03_ctl01_PageSizeComboBox_Input'
+        self._total_pages_regex = r'[0-9].*? items in ([0-9].*?) page'
+        super(UsPaGreeneScraper, self).__init__('us_pa_greene')
+
+    def _retrieve_session_vars(content):
```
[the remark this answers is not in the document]
arian487: I think that makes sense generally.  At google we require docs for all public methods and generally more complex private methods as well, but for very small, obvious helpers we usually don't do that.  However I'm fine doing it for everything.  I've updated this PR to have that.

recidiviz/ingest/us_pa_greene/us_pa_greene_scraper.py
```diff
@@ -0,0 +1,384 @@
...
+        """
+        task_type = params['task_type']
+        params_list = []
+        # If it is our first task, we know the next task must be a query to
+        # return all inmates
+        if task_type & constants.INITIAL_TASK:
+            params_list.append(self._get_all_inmates_params(content))
+            return params_list
+        # In this case, we need to return more inmates so scraping will go faster.
+        elif ((task_type & constants.GET_MORE_TASKS) && (
```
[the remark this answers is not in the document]
arian487: Done.

recidiviz/ingest/us_pa_greene/us_pa_greene_scraper.py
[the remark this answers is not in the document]
arian487: Fixed, sorry about that.

recidiviz/ingest/us_pa_greene/us_pa_greene_scraper.py
```diff
@@ -0,0 +1,384 @@
...
+        Returns:
+            A dict of session vars needed for the next scrape.
+        """
+        data = {}
+        for session_var in self._required_session_vars:
+            data[session_var] = scraper_utils.get_id_value_from_html_tree(
+                content, session_var)
+        return data
+
+    def _get_all_inmates_params(self, content):
```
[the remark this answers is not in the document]
arian487: Fixed it in all locations

recidiviz/ingest/us_pa_greene/us_pa_greene_scraper.py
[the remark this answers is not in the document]
arian487: We can definitely subclass the generic scraper to create an archonix scraper and have the children inheric from archonic_scraper.

recidiviz/ingest/us_pa_greene/us_pa_greene_scraper.py
```diff
@@ -0,0 +1,384 @@
...
+        Returns:
+            A string representing the persons record id
+        """
+        return self._get_scraped_value(content, 'ctl00_ContentPlaceHolder1_spnBookingNo')
+
+    def get_record_class(self):
+        """Returns the record subclass to use for this scraper."""
+        return UsPaGreeneRecord
+
+    def record_id_is_fuzzy(self, content, params):
```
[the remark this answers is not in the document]
arian487: Removed.

recidiviz/ingest/us_pa_greene/us_pa_greene_scraper.py
```diff
@@ -0,0 +1,384 @@
...
+            A bool representing whether the record is fuzzy.
+        """
+        pass
+
+    def get_county_of_commit(self, content, params):
+        """Gets county of commit given a page
+        Returns:
+            A string representing the county of commit.
+        """
+        return 'Greene County'
```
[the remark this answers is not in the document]
arian487: Ah gotcha, didn't know.  Thanks for the clarification!

recidiviz/ingest/generic_scraper.py
```diff
@@ -0,0 +1,620 @@
...
+        """Gets the person subtype to use for this scraper.
+        """
+        return Record
+
+    def get_snapshot_class(self):
+        """Gets the snapshot subtype to use for this scraper.
+        """
+        return Snapshot
+
+    def get_person_id(self, content, params):
```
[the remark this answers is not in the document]
arian487: Yup, I have a todo in my head for the generic scraper project to actually write a generator to generate generic_scraper based on the db objects.

recidiviz/ingest/us_pa_greene/us_pa_greene_scraper.py
```diff
@@ -0,0 +1,384 @@
...
+            # Next check to see if there is another page we can scrape.
+            params_list.extend(self._get_next_page_if_exists(content))
+
+    def _get_scraped_value(self, content, scrape_id):
+        """Convenience function to get a scraped value from a row.
+
+        Returns:
+            The value scraped from the page.
+        """
+        text = search_tree.cssselect(
```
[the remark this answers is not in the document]
arian487: Good catch!  That was a testing artifact.  Its fixed now.

recidiviz/ingest/us_pa_greene/us_pa_greene_scraper.py
```diff
@@ -0,0 +1,384 @@
...
+        record.committed_by = self._get_scraped_value(content, 'ctl00_ContentPlaceHolder1_spnHoldForAgency')
+        record.reference_id = params['reference_id']
+
+    def get_offenses(self, content, params):
+        """Gets the list of offenses relevant to the record.
+        Returns:
+            A list of Offense objects for the record.
+        """
+        offenses = []
+        charges = search_tree.cssselect('[id=ctl00_ContentPlaceHolder1_gridCharges_ctl00]')[0]
```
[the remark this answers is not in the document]
arian487: Fixed.

recidiviz/ingest/us_pa_greene/__init__.py
arian487: Done.

recidiviz/ingest/us_pa_greene/us_pa_greene_record.py
arian487: I agree, I've kept it this way for now, I think we should very soon have a conversation about how best to handle unique params and what should go in the upper level and what should go into a specific sub class.

recidiviz/ingest/us_pa_greene/us_pa_greene_record.py
arian487: I agree.  In general it would be nice if everything could live at top level as it would simplify a lot of things, especially as we move to make things as generic as possible.  Lets have a discussion soon regarding this, I'll leave this here for now.

region_manifest.yaml
arian487: Done.

recidiviz/ingest/generic_scraper.py
arian487: Done.

recidiviz/ingest/scraper_utils.py
arian487: Agreed, lets do that in another PR.

recidiviz/ingest/us_pa_greene/us_pa_greene_record.py
arian487: Done.

recidiviz/ingest/generic_scraper.py
```diff
@@ -0,0 +1,620 @@
...
+        task_type = params.get('task_type', self.get_initial_task_type())
+        # We always fetch some content before doing anything.  Note that we
+        # use get here for the data to return a default value of None if
+        # this scraper doesn't set it.
+        content = self._fetch_content(endpoint, data)
+        if content == -1:
+            return -1
+        # For an initial task we just make a call to get initial variables.
+        # This may be a noop for some scrapers.
+        if task_type & constants.INITIAL_TASK:
```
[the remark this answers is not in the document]
arian487: Hm good point.  I've made more specific functions in generic_scraper, but we can refactor that into util helper when we do a cleaner util refactor.

recidiviz/ingest/generic_scraper.py
arian487: Done.

recidiviz/ingest/generic_scraper.py
arian487: Done.

recidiviz/ingest/generic_scraper.py
arian487: Lets file the issue, but also we should discuss snapshot logic a bit better because I don't think we are being as efficient as we can with that right now.

recidiviz/ingest/generic_scraper.py
arian487: Done.

recidiviz/ingest/generic_scraper.py
arian487: Ah didn't know about that, thanks!

recidiviz/ingest/generic_scraper.py
arian487: I think you're right.  Any public method should definitely have it, I was just sloppy.  I don't think it is absolutely necessary for very small and simple private functions though.

arian487: Added it in.

recidiviz/ingest/generic_scraper.py
arian487: Fixed the comment

recidiviz/ingest/generic_scraper.py
arian487: I'm going to remove that, this logic should be in the parent class anyways.

recidiviz/ingest/generic_scraper.py
arian487: Fixed.

recidiviz/ingest/us_pa_greene/__init__.py
arian487: Done.

recidiviz/ingest/scraper_utils.py
arian487: Done.

region_manifest.yaml
arian487: Ah sorry, not needed for this region.

recidiviz/ingest/us_pa_greene/us_pa_greene_scraper.py
arian487: Done.

recidiviz/ingest/us_pa_greene/us_pa_greene_scraper.py
arian487: Sure, happy to do either way.  Changed it :)

recidiviz/ingest/us_pa_greene/us_pa_greene_scraper.py
arian487: Yea I definitely agree, we'll want to have as many useful scraping utils as possible.

recidiviz/ingest/us_pa_greene/us_pa_greene_scraper.py
arian487: I agree, I think some of that generalization will depend on the eventual goal having scrapers be config files.  We can have some way for a scraper to tell the generic scraper its scrape flow.

recidiviz/ingest/us_pa_greene/us_pa_greene_scraper.py
arian487: Agreed, lets have a separate PR where we pull out useful scrape util type things into its own utility class.  I sort of realized this half way and hence stopped adding things to the single the util file.

recidiviz/ingest/us_pa_greene/us_pa_greene_scraper.py
arian487: Done.

recidiviz/ingest/us_pa_greene/us_pa_greene_scraper.py
arian487: Not sure if there is a python module fpor this, but I've moved this to the scraper util for now.

recidiviz/ingest/us_pa_greene/us_pa_greene_scraper.py
```diff
@@ -0,0 +1,384 @@
...
+        """Gets county of commit given a page
+        Returns:
+            A string representing the county of commit.
+        """
+        return 'Greene County'
+
+    def get_custody_date(self, content, params):
+        """Gets custody date given a page
+        Returns:
+            A datetime representing the persons custory date
```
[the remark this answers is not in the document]
arian487: Fixed.

region_manifest.yaml
arian487: Fixed.

recidiviz/ingest/generic_scraper.py
```diff
@@ -0,0 +1,620 @@
...
+                # Always pass along the scrape type as well.
+                task_params['scrape_type'] = params['scrape_type']
+                self.add_task('_generic_scrape', task_params)
+        if task_type & constants.SCRAPE_PERSON:
+            params['person'] = self._populate_person(content, params)
+        if task_type & constants.SCRAPE_RECORD:
+            record = self._populate_record(content, params)
+            self._populate_snapshot(record)
+
+    def _populate_person(self, content, params):
```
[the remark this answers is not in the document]
arian487: As you mentioned, this is much less of a problem when the data model is handled in a parent class.  I had mentioned above that one plan I have for the generic scraper is constructing it automatically based on the db model so a 'presubmit' before any db changes would be running the generator.  Most of that can be constructed automatically, and the generator can in addition find all getters in the children and rename them as well to stay in sync.  This should fix most of the drift issues (and in general it'll be even less of an issue when as many of the getters as possible live in the parent).

recidiviz/ingest/generic_scraper.py
arian487: Thanks :)

This will also pay dividends when my plans for the generalized scraper come to pass, overriding individual might have to be a thing for some child scraper and this was not possible in the previous model.

arian487: individual fields*

recidiviz/ingest/generic_scraper.py
arian487: Its removed as I am just using the base now.

recidiviz/ingest/generic_scraper.py
```diff
@@ -16,10 +16,29 @@
...
+        params and returns a list of params defining what to scrape next.
+        The params must include 'endpoint' and 'task_type' which tell the 
+        generic scraper what endpoint we are getting and what we are doing
+        with the endpoint when we do get it.
+    2.  set_initial_vars:  This function should set any initial vars needed
+        after a first scrape (session vars for example)
+    3.  All of the getters which populate one of the DB fields need also be
+        implemented (for now).  Each getter takes content and params, and the
+        child scrapers need to know how to extract out that data given those
+        two params.
```
jessex: Doc nit: we're using `params` twice in this sentence but in two slightly different ways.

recidiviz/ingest/generic_scraper.py
```diff
@@ -16,10 +16,29 @@
...
+
+This is a class that aims to handle as many DB related operations as possible
+as well as operations around how to get the content of a page.  It provides
+generic implementations of getter functions which aim to scrape out important
+fields we care about.
+
+In order to sublcass this the following functions must be implemented:
+
+    1.  get_more_tasks: This function takes the page content as well as the
+        params and returns a list of params defining what to scrape next.
```
jessex: Related to the doc nit below, it might be good to more explicitly state what the params are here in its first usage, e.g. "... as well as the scrape task parameters, e.g. session variables, ..."

recidiviz/ingest/generic_scraper.py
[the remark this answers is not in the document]
jessex: That sounds good to me. It would be clutch to have a unit test case that checks all of these getters on the generic scraper and compares it against the set of attributes on the core entities, ensuring that we've got a getter for each. That way if someone forgets to re-generate when necessary, the build fails straight away. If you add it to this PR, I think you could close #126 in the process.

recidiviz/ingest/generic_scraper.py
jessex: Agreed that docs shouldn't be required for trivial methods going forward.

recidiviz/ingest/generic_scraper.py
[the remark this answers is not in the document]
jessex: Let's file an issue for that before we close this PR up!

recidiviz/ingest/scraper_utils.py
```diff
@@ -152,7 +152,11 @@ def get_id_value_from_html_tree(html_tree, id):
     Returns:
         A string representing the value of the id from the html page.
     """
-    html_tree.cssselect('[id={}]'.format(id))[0].get('value')
+    html_obj = html_tree.cssselect('[id={}]'.format(id))
+    if html_obj:
+        return [0].get('value')
```
jessex: This should be `html_obj[0]`, right? There's no shorthand for indexing into an array you just checked in a conditional?

recidiviz/ingest/scraper_utils.py
```diff
@@ -226,3 +230,15 @@ def get_headers():
...
+def currency_to_float(currency):
+    """Converts a given currency string value to a float.
+
+    Args:
+        currency: A string representing the dollar currency dollar amount.
+
+    Returns:
+        A float representing the currency.
+    """
+    return float(currency[1:].replace(',', ''))
```
jessex: This is suitable for predictable USD currencies but won't internationalize well, and also wouldn't work for currency strings that don't start with a `$`. We'll want to keep that in mind as we expand. There's surely a package that handles this automatically (I see a few looking at Google).

recidiviz/ingest/generic_scraper.py
```diff
@@ -280,12 +285,12 @@ def _compare_and_set_snapshot(self, record):
             for attr in snapshot_attrs:
                 if attr not in ['class', 'created_on', 'offense']:
                     record_value = getattr(record, attr)
-                    last_snapshot_value = getattr(last_snapshot, attr)
-                    if record_value != last_snapshot_value:
+                    old_record_value = getattr(old_record, attr)
+                    if record_value != old_record_value:
```
jessex: Yeah this method of setting those fields which should be set on a new snapshot instead of creating a full snapshot and then nulling out unchanged fields is obviously more efficient, and I can't think of a reason we did the opposite previously. Good fix!

The only call-out I have here is that in the old `record_to_snapshot` copy, we would explicitly set the snapshot's parent to `record.key`, which we do need to do here to ensure the object hierarchy is maintained.

recidiviz/ingest/us_pa_greene/us_pa_greene_record.py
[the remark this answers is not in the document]
jessex: Yeah that's a very good example of the case I was trying to paint. I think we should probably keep the field named as you have it but comment on what it means and why that's the name.

recidiviz/ingest/us_pa_greene/us_pa_greene_person.py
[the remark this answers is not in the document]
jessex: Oh, got it, that makes perfect sense. As I said on the related throw just below, I propose we keep the name but comment on what it is and why it's named this way, unless @ohinds-vera disagrees.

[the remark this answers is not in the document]
arian487: Ok done, I've added comments regarding this.

recidiviz/ingest/generic_scraper.py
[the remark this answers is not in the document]
arian487: Done, filed here:

https://github.com/Recidiviz/recidiviz-data/issues/153

recidiviz/ingest/scraper_utils.py
arian487: Great.

recidiviz/ingest/generic_scraper.py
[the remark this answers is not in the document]
arian487: Sounds good, if the test sees drift it just recommends to the user that they run the generator script again.

recidiviz/ingest/generic_scraper.py
arian487: Fixed.

recidiviz/ingest/generic_scraper.py
arian487: Done.

recidiviz/ingest/scraper_utils.py
arian487: Yea most of the top responses are to just do it manually.  I've added a comment to the function regarding its uses.

arian487: Also it works for any currency that starts with a symbol, not just $ :)

recidiviz/ingest/scraper_utils.py
arian487: Yes.   Sorry!

recidiviz/ingest/generic_scraper.py
```diff
@@ -0,0 +1,906 @@
...
+                    else:
+                        setattr(new_snapshot, attr, None)
+                elif attr == 'offense':
+                    # Offenses have to be treated differently -
+                    # setting them to None means an empty list or
+                    # tuple instead of None, and an unordered list of
+                    # them can't be compared to another unordered list
+                    # of them.
+                    #
+                    # TODO #123 instead of storing the entire list if
```
jessex: We can remove this TODO. We've closed that issue as WON'T FIX.

recidiviz/ingest/scraper_utils.py
jessex: Yes, true. On the localization point I was thinking more about the fact that commas and decimal points are flipped between different countries, but again we're probably fine here for a while. :+1:

recidiviz/ingest/us_ny/us_ny_scraper.py
```diff
@@ -46,7 +46,6 @@
 import re
 
 from lxml import html
-from lxml.etree import XMLSyntaxError  # pylint:disable=no-name-in-module
```
jessex: It's odd, there is no merge conflict but you've gotten a couple of my own previous commits included as diffs here, specifically:
* https://github.com/Recidiviz/recidiviz-data/pull/147/commits/2cc8f4556014bbc8b6d6cad8c7f6093b22e45760
* https://github.com/Recidiviz/recidiviz-data/pull/147/commits/4c91fdae563b63964518d2e972be7a334b2a6c4c

I'm not sure why you're not getting a giant merge conflict right now but I'd try to rebase against master again and get my own commits out of this PR.

recidiviz/ingest/us_pa_greene/us_pa_greene_person.py
```diff
@@ -0,0 +1,44 @@
...
+    """A subclass of Person that adds Greene County, PA specific fields
+
+    Datastore model for a specific person incarcerated in
+    Vermont. This extends the Person class.
+
+    Attributes:
+
+        us_pa_greene_person_id: (string) Same as person_id, but used as
+            key for this entity type to force uniqueness / prevent
+            collisions within the us_vt records (see
```
jessex: us_pa_greene here instead of us_vt?

recidiviz/ingest/us_pa_greene/us_pa_greene_scraper.py
```diff
@@ -0,0 +1,498 @@
...
+            scrape_id: the html id to scrape out of the page
+
+        Returns:
+            The value scraped from the page.
+        """
+        text = content.cssselect(
+            '[id={}]'.format(scrape_id))[0].text_content()
+        return ''.join(text.split()).split(':')[1]
+
+    # pylint:disable=unused-argument
```
jessex: What is this one referring to? The method directly below does reference its only argument, `self`. The method directly above references both of its arguments.

recidiviz/ingest/us_pa_greene/us_pa_greene_scraper.py
```diff
@@ -0,0 +1,498 @@
...
+            content: An lxml html tree.
+            params: dict of parameters passed from the last scrape session.
+
+        Returns:
+            A string representing the persons given name
+        """
+        full_name = content.cssselect(
+            '[id=ctl00_ContentPlaceHolder1_spnInmateName]')[
+                0].text_content().strip()
+        return full_name.split(',')[1].strip()
```
jessex: This would break for a mononymous person (one name), no?

recidiviz/ingest/us_pa_greene/us_pa_greene_scraper.py
```diff
@@ -0,0 +1,498 @@
...
+        for row in body:
+            offense = UsPaGreeneOffense()
+            offense.crime_class = row[0].text_content()
+            offense.description = row[1].text_content()
+            offense.case_number = row[2].text_content()
+            # Note that row[3] and row[4] represent the offense date and arrest
+            # date, But PA Greene leaves this field blank.
+            bond_str = row[5].text_content()
+            # If there is no bond, it is filled with this character.
+            if bond_str != u'\xa0':
```
jessex: [Looks like](https://stackoverflow.com/questions/10993612/python-removing-xa0-from-string) this corresponds to some odd unicode "non-breaking space." I imagine there's a Pythonic method in the standard library somewhere that captures this instead of a direct character comparison?

jessex: Did a quick test, you can use `bond_str.isspace()` here. It will return true here.

recidiviz/ingest/us_ny/us_ny_scraper.py
jessex: To put a point on it, it looks like a _majority_ of the lines changed in here are from my commits being accidentally included.

jessex:
Only a few more minor remarks from me, and one comment about needing to clean up the commit log in this branch before merging. I've also gone back through all comments and catalogued a list of issues which I will file later this evening. Once we've got these last comments addressed and the commit log contains only your commits on the branch, and the build is still green, I'll squash and merge.

recidiviz/ingest/generic_scraper.py
arian487: Done.

recidiviz/ingest/us_pa_greene/us_pa_greene_person.py
arian487: Done.

recidiviz/ingest/us_pa_greene/us_pa_greene_scraper.py
arian487: Thanks, gotta remember that python has functions for everything!

arian487:
I've not run this e2e yet, was planning on doing that tonight or tomorrow morning, I have run all of the getters as well as the get new tasks on testdata though so those should be fine.

recidiviz/ingest/us_pa_greene/us_pa_greene_scraper.py
jessex: Hm, there are times where Pylint will do something similar in test code with undocumented methods: it will complain about one to a few seemingly arbitrary methods in the class being undocumented but not the rest. Maybe it has something to do with top-to-bottom, line-by-line examination of the file? I haven't looked into it yet.

jessex:
Wonderful work here, @arian487. I'm ready to approve once you've got an end-to-end run in the books locally. Let me know when you've had a chance to confirm and then I'll squash and merge this.

recidiviz/ingest/us_pa_greene/us_pa_greene_scraper.py
arian487: Yea I tried removing it and experimenting a bit to see but it did seem arbitrary functions it decided to complain about.

arian487:
Thats interesting!  So the viewstate is needed definitely needed on every subsequent call, the system always reverts to the homepage with nothing searched if there isn't a proper viewstate passed in.  A quick search on the internet tells me that this is necessary when scraping asp websites, if the website requires it (which archonix does).  I decoded the text and they are unfortunately sending in every variable under the sun when they should just be sending in the delta, but we can't control archonix sadly.  Do you know by how much we are going over the limit?  Using gz compression the viewstate string can be compressed from **27956** bytes to **10035** bytes.  However it is a relatively costly operation, about 1 second by my measurements, decompressing is much faster though.  We can do a bit better because the viewstate is actually base 64 encoded, so we can decode this and compress the decoded text which gets us down to about **6000** bytes.  

We might be able to do better yet by trying to reconstruct the viewstate from the delta of our actions.  IE.  We send in the task params only the delta (search + next page for example) and then construct the viewstate from that in the scraper, base64 encode it, and send it along in the request.  I'll do some experimentation to see how viable that is.  I'll decode it and see if I can understand how the viewstate builds on top of each of the last calls, it should be possible to construct it from our deltas.

Finally, it may be worth setting up a small local db service which the scrapers can dump their params into (something like redis).  If we anticipate issues with our task params length long term, this may be useful for many scrapers.  

Thoughts?  I think reconstructing it from the deltas is the nicest option but the quickest right now would be if the compression is good enough?  Do you know if that is enough compression?  If so I'll add that real quick and then see about the delta reconstruction.

arian487:
Looking at this further, it seems the first viewstate is fine, and its only when you start going down the pages that it exceeds the byte limit.  The compression here would certainly be good enough.  Later tonight I'll push that and continue to investigate constructing the viewstate ourselves using the deltas.  

This viewstate is extremely inefficiently constructed on their end...lots of unnecessary info.

jessex:
On the run right now so I can't give a longer more thoughtful comment, but the last line of the stack trace has the numbers. The total task size is 133598 bytes and the max is 102400, at least for this particular request.

recidiviz/ingest/scraper_utils.py
```diff
@@ -247,3 +248,28 @@ def currency_to_float(currency):
...
+    Args:
+        s: The string to be compressed
+        level: The level of compression, 1 is almost always the right answer
+            here as the multi-level compression is much slower and provides
+            very little benefit.
+
+    Returns:
+        The compressed string
+    """
+    return zlib.compress(s, level)
```
jessex: It looks like both of these `zlib` calls fail if the string arg is None. Should we validate against that in the method itself, or return None if given None?

recidiviz/ingest/scraper_utils.py
```diff
@@ -247,3 +248,28 @@ def currency_to_float(currency):
         A float representing the currency.
     """
     return float(currency[1:].replace(',', ''))
+
+def compress_string(s, level=1):
```
jessex: These should probably be placed into a dedicated module instead of being bundled inside the generic scraper.

jessex:
Thanks a lot, @arian487! What sort of compression do we end up with? How big is the params dictionary before and after the compression, without the decoding/encoding?

recidiviz/ingest/scraper_utils.py
arian487: I think we probably want it to fail actually, if someone is calling this function they probably expect there to be content of some kind.

recidiviz/ingest/scraper_utils.py
arian487: You mean instead of scraper utils right?  Agreed, thats part of the issue which is open to break out scraper utils which we will do soon.

arian487:
It depends on what is passed and the current viewstate which can vary slightly based on the actual html content but when I tested I went from **127944** bytes to **26545** bytes.  Adding another 2 levels of compressions gets us to ~23k bytes but is much slower, so IMO we stick with 1 for now.  The compression is enough to get us well under the limit.

jessex:
Sounds good, @arian487. I pushed up another fix dealing with the fact that the `data` passed to `transform_data` is None on the initial call. However, there is an encoding issue here:
```
  File "/Users/clementinejacoby/Documents/projects/recidiviz-data/recidiviz/ingest/worker.py", line 90, in post
    result = scraper_task(params)
  File "/Users/clementinejacoby/Documents/projects/recidiviz-data/recidiviz/ingest/generic_scraper.py", line 138, in _generic_scrape
    self.add_task('_generic_scrape', task_params)
  File "/Users/clementinejacoby/Documents/projects/recidiviz-data/recidiviz/ingest/scraper.py", line 301, in add_task
    params_serial = json.dumps(params)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/__init__.py", line 243, in dumps
    return _default_encoder.encode(obj)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/encoder.py", line 207, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/encoder.py", line 270, in iterencode
    return _iterencode(o, 0)
UnicodeDecodeError: 'utf8' codec can't decode byte 0xed in position 2: invalid continuation byte
```

We turn our scraper params into json before inserting them into the task params, and that json serialization is choking on the compressed session var.

arian487:
Ah ok, so the zlib encoding is one that json cannot handle so the solution is to base64 encode the zlib compressed data (we have set the level to 9 here so as to not get newlines in the resulting string and ensure and exact matched decompression) and then that'll let us happily dump it.  Just testing in local terminal:

import json
import zlib

# I actually used a real viewstate here to be sure :)
s = 'long_viewstate'
compressed = zlib.compress(s, 9).encode('base64')
params = {
  data = {
    '__VIEWSTATE':  s
  }
}
js = json.dumps(params)
params2 = json.loads(params2)
s2 = params2['data']['__VIEWSTATE']
zlip.decompress(s2.decode('base64')) == s

>> True

Let me whip up a quick CL to fix this.

arian487:
Correction I used:

params = {
data = {
'__VIEWSTATE': compressed
}
}

Also note this worsens the compression a bit, but only by 1-2% so its well under our max

jessex:
Here are some more fixes for you, @arian487! Good news: the compression is working and we're able to scrape end-to-end! Data is in Datastore and the system is working. Room for improvement news: there are some issues in parsing of the page that are causing gaps and errors in some cases. Some examples I've noticed so far:
* Persons matching some characteristics I didn't determine yet are failing in `get_offenses` because the `row` that we iterate over does not have more than 1 internal entry, leading to an `IndexError` (line 510).
* We are leaving gaps in the offense data we do parse, notably we are missing the bond amount and crime description each time.
* We are missing at least `surname` and `us_pa_greene_person_id` for all Persons and Records. Potentially others.
* Some Persons and Records are having every single custom field (i.e. not `created_on` or `class` or the ids) set to None.
* Every single `race` field is set to "White/Eurp/N.Afr/MidEas" (minus those who have it set to None due to the previous bug).

There could be more. It's a bit of whack-a-mole at the moment from my end (going to be happy when we sort out known issues and you can run this yourself) but we should probably pair up on this early this week.

arian487:
That sounds good!  Thanks for the help.  Do you think we should merge this now so the fellows have it as a model this week when they write their own scrapers?

jessex:
I can merge, but would you file an issue to fix the parsing bugs for us_pa_greene? I don't love merging a PR with known bugs and with a decrease in test coverage but we can do this as a one-time measure as long as we have the requisite issues filed, given the circumstances. (We've already got the issue for test coverage, just need one for fixing the parsing issues I highlighted above.)

arian487:
I agree for sure, I've filed bug:

https://github.com/Recidiviz/recidiviz-data/issues/160

I'll work on those asap, thanks for helping get this through before the fellows arrive.  They should have a solid model to base their own scrapers off of now!

PR #163, opened 2018-11-02
Added pipenv workflow.

rasmi:


README.md
```diff
@@ -58,7 +58,9 @@ Your mileage may vary here. If you've installed via the interactive installer an
 `google-cloud-sdk/platform/`, then you will need to install google_appengine separately by doing:
 `gcloud components install app-engine-python`
 
-Then install project dependencies: `pip install -r requirements.txt`.
+Then install project dependencies with [`pipenv`](https://pipenv.readthedocs.io/en/latest/): `pipenv install`.
+
```
jessex: Can we add one more brief line in here specifying that commands later in the doc, e.g. to run `dev_appserver`, should be launched from within the `pipenv shell`? Might not be obvious to future contributors.

PR #164, opened 2018-11-02
Added surname return statement to us_pa_greene

rasmi:


recidiviz/ingest/us_pa_greene/us_pa_greene_scraper.py
```diff
@@ -350,6 +350,9 @@ def get_surname(self, content, params):
             '[id=ctl00_ContentPlaceHolder1_spnInmateName]')[
                 0].text_content().strip()
         full_name.split(',')[0].strip()
+        last_name = full_name.split(',')[0].strip()
+
+        return last_name
```
jessex: Oh, nice. Thanks for the fix, @rasmi!

PR #168, opened 2018-11-06
Scraper implementation for Mesa County, Colorado

colincadams:
You'll notice there are quite a few todos in here. The mesa county setup doesn't play nice with our current schema (re: one bond -> many charges), so once we have the new schema decided we will need to go through and fix some of these.


recidiviz/ingest/generic_scraper.py
```diff
@@ -153,6 +153,8 @@ def _populate_person(self, content, params):
             A person database object that is passed through to future functions.
         """
         person_id = self.get_person_id(content, params)
+        # TODO: Does the person id get prefixed by the region automatically? If
```
jessex: No this does not happen automatically yet, but it should. Check out `us_ny_scraper` lines 763-764 for a current example. We could put a decorator around the getter which prepends our region code to the front of all returned person ids (and similarly for record ids, maybe).

recidiviz/ingest/us_co_mesa/us_co_mesa_scraper.py
```diff
@@ -0,0 +1,196 @@
...
+    # - 'The inmate currently has a hold that could prevent him from being
+    #    eligible for release after posting bond...'
+
+    @staticmethod
+    def _scrape_person(content):
+        person = Person()
+
+        table = content.cssselect('table')[0]
+        for row in table:
+            if len(row) == 2:
```
jessex: If the row has only one element then the value we want to extract is blank? I'd add an inline comment on that, since `if length is 2` otherwise appears magical.

recidiviz/ingest/us_co_mesa/us_co_mesa_scraper.py
```diff
@@ -0,0 +1,196 @@
...
+    #    has not been set at this time...'
+    # - 'The inmate currently is eligible for release after posting bond...'
+    # - 'The inmate currently has a hold that could prevent him from being
+    #    eligible for release after posting bond...'
+
+    @staticmethod
+    def _scrape_person(content):
+        person = Person()
+
+        table = content.cssselect('table')[0]
```
jessex: If the `table` selector isn't found then this will error out with a fairly unhelpful error message. But adding checks around every call and refusing to chain commands together makes for reduced readability.

It begs a larger question about how defensively we code across scrapers, how much we assume that the page structure we see will remain the same versus conditionally checking every little feature. Do you have any thoughts on this in your experience so far, @colincadams?

recidiviz/ingest/us_co_mesa/us_co_mesa_scraper.py
```diff
@@ -0,0 +1,196 @@
...
+
+        if self.is_initial_task(task_type):
+            params_list.append(self._get_all_people_params())
+        elif self.should_get_more_tasks(task_type):
+            params_list.extend(self._get_person_params(content))
+
+        if self.should_scrape_person(task_type):
+            self._person = UsCoMesaScraper._scrape_person(content)
+        if self.should_scrape_record(task_type):
+            self._record = UsCoMesaScraper._scraper_record(content)
```
jessex: Typo: `_scrape_record`

recidiviz/ingest/us_co_mesa/us_co_mesa_scraper.py
```diff
@@ -0,0 +1,196 @@
...
+
+        Returns:
+            A class representing the person DB object.
+        """
+        return Person
+
+    def get_record_class(self):
+        return Record
+
+    def get_person_id(self, content, params):
```
jessex: This scraper is working slightly differently from its forerunner in us_pa_greene, I see. That one is taking in the content and pulling specific fields out in each getter. This one on the other hand is directly scraping data inside of `get_more_tasks` and then holding the entity in memory instead of letting `generic_scraper` handle it. Is that deviation a conscious design decision?

recidiviz/ingest/us_co_mesa/us_co_mesa_scraper.py
arian487: To comment about this, I think we should defensively if the failure should not halt the execution of the task.  We should rely on our monitoring to see these issues rather than let the entire execution not be scraped properly.  In the future having consistent data will actually be important so a small bug in one field for example shouldn't result in entire people not being scraped but rather a single offending field. 

Conditionally checking every little feature makes the code less readable, but it'll also make us very happy from a monitoriong standpoint.  Someone can easily look at our dashboards and see what is failing.

region_manifest.yaml
```diff
@@ -58,3 +58,16 @@ regions:
...
     scraper_package: us_pa_greene
     timezone: America/New_York
+  us_co_mesa:
+    region_name: UsCoMesa
+    agency_name: Mesa County Sheriff's Office
+    region_code: us_co_mesa
+    agency_type: jail
+    queue: us-co-mesa-scraper
+    base_url: http://inmateinfo.mesacounty.us
+    names_file: us_ny_names.csv # TODO: should not be required
```
jessex: Agreed. Are you planning to tackle making this optional in the current PR? If not, I think as a general guideline we should file issues for future work and reference the issue number in the TODO, e.g. `# TODO (456): Make names_file optional`.

recidiviz/ingest/us_co_mesa/us_co_mesa_scraper.py
jessex: I lean in that direction too, but am interested to see if we can somehow structure the code to hide that defensiveness in, say, decorators so that actual scraper code reads more humanely.

recidiviz/ingest/us_co_mesa/us_co_mesa_scraper.py
arian487: So we've discussed this in person here.  The current design makes it hard if there is precomputation to be done.  For example you dont want to have to walk the content for every field as it will be a lot of duplicate code and also generally more inefficient.  You might want to walk the html and pull out fields as you go (which the getter logic doesn't do).

Colin is temporarily is doing it this way but we are changing the logic from getters, to setters that are called in a single overloaded function called populate_x.  This makes scrapers smaller and allows us to define an ingest data model (a mostly flat one) that people call setters on all in one place, the setters will be strongly typed and allow the user to easily see what options are available, and has the added benefit for allowing someone to walk content and pull out values.

@zacharylawrence is working on the flat ingest data model, and we will restructure generic_scraper once that is done.

jessex: Arian and I discussed this offline and I see now that this foreshadows a structural change coming in generic_scraper. Feel free to comment if you have thoughts you'd like to share, or just ignore this now.

recidiviz/ingest/generic_scraper.py
colincadams: I added an issue to this todo to generify it.

I've also copied the code from NY but I think there is an issue with how it handles conflicts. scraper_utils.generate_id checks to see if the id it has generated already exists, but at this point the region_code has not been prepended so it isn't checking the final id for conflicts.

recidiviz/ingest/us_co_mesa/us_co_mesa_scraper.py
colincadams: I've actually just deleted this line so that we don't silently skip data if the page format change (i.e. all rows are changed to have 3 values), this is inline with my next comment. I do think that this is a part of a bigger assumption about how we handle page format changes and a potential abundance of KeyErrors :)

recidiviz/ingest/us_co_mesa/us_co_mesa_scraper.py
colincadams: Skipping talking about this generically for a second, this is what I would expect to happen in this particular case:

* No table exists, this line throws a KeyError
* Generic scraper catches the error
  * Increments the error count metric, with region code "us_co_mesa" and error code "FAILED_TO_PARSE_PERSON"
  * Logs error and the stack trace

Then if the format truly changed and this was happening all of the time, an alert would fire. The maintainer would see that all of the errors are for the given combination. They would look through the logs and see that all of the logs for that region show errors occurring on this particular line, so the format must have changed such that the cssselector is failing and investigate how to fix the scraper.

Does that make sense? I think that skipping info is generally bad in this case, but it is definitely a skipping information vs maintenance burden issue. I would opt for not skipping any information, but trying to make it as easy as possible to identify the underlying issue. Open to other thoughts!

recidiviz/ingest/us_co_mesa/us_co_mesa_scraper.py
colincadams: Good catch!

region_manifest.yaml
colincadams: Done, moved to regions.py with issue number #169

region_manifest.yaml
```diff
@@ -65,7 +65,7 @@ regions:
     agency_type: jail
     queue: us-co-mesa-scraper
     base_url: http://inmateinfo.mesacounty.us
-    names_file: us_ny_names.csv # TODO(#169): should not be required
```
jessex: Sorry, why did we remove the TODO here? We still want it because we're still waiting on #169 to make this optional, right?

recidiviz/ingest/generic_scraper.py
jessex: You're right, that should probably be fixed as part of the same task. I'll comment on #169.

recidiviz/ingest/us_co_mesa/us_co_mesa_scraper.py
jessex: I agree with that expected flow, @colincadams. But going from there onward, would you have the underlying task eventually fail and become a candidate for restart, or would you have it return a "normal" value and let the task silently complete? I'm guessing the latter since it's the lowest friction route while accomplishing your expected flow. It seems like, generally speaking again, we want to only instigate retry for issues that we expect to be transient, and to track/count/monitor/alert on issues that we expect to be permanent without resolution. But with Task Queue, it's easy to get into the pattern of failing tasks for both classes and thus retrying everything. So we need guidance and consistency on this.

colincadams: (I have not added this to the doc but I will add it)

I think conceptually this should return a non-retryable error. That way we won't keep retrying it, but we also won't think that the full scrape is successful. I think that distinction is meaningful if for example we add a metric for last full scrape completion time and use it to track how fresh our data is for each region. If we were failing to scrape persons for a particular region we wouldn't want to update that metric.

How we implement this though might be that we return a "normal" value so that task queue doesn't retry, but somehow pass along information so that our code knows there was an error. I haven't looked through task queue or our code that deals with it enough to have a sense of how this would actually work.

colincadams: I think this makes sense as is to me, but open to changing it to conform with how other child scrapers work currently.

PR #171, opened 2018-11-06
Add jailtracker base scraper

zdg2102:
Currently a no-op, because I figured this was already long enough for someone to have to read without also updating the Vermont scraper to use it. I'll update the Vermont scraper separately once this is checked in.


recidiviz/ingest/vendors/jailtracker/jailtracker_scraper.py
```diff
@@ -0,0 +1,425 @@
...
+
+import abc
+import re
+
+from lxml import html
+from recidiviz.ingest import constants
+from recidiviz.ingest.generic_scraper import GenericScraper
+
+
+class JailtrackerScraper(GenericScraper):
```
jessex: Very minor nit: since the name is stylized JailTracker and we spell it that way ourselves elsewhere, I think the class name should be `JailTrackerScraper` with the capitalized T.

recidiviz/ingest/vendors/jailtracker/jailtracker_scraper.py
```diff
@@ -0,0 +1,425 @@
...
+                },
+            _ENDPOINT: charges_request_endpoint,
+            _PERSON: params[_PERSON],
+            _REQUEST_TARGET: _CHARGES_REQUEST,
+            _SCRAPE_TYPE: params[_SCRAPE_TYPE],
+            _SESSION_TOKEN: params[_SESSION_TOKEN]
+            }
+
+
+    # Overrides method in GenericScraper to handle JSON responses.
```
jessex: I'm guessing we're not going to extract JSON-handling upwards until we have more scrapers/use cases for it, right?

recidiviz/ingest/vendors/jailtracker/jailtracker_scraper.py
zdg2102: Makes sense to me, updated

recidiviz/ingest/vendors/jailtracker/jailtracker_scraper.py
zdg2102: Yeah, that was my thinking. I could definitely see us ending up moving JSON handling to the generic scraper if other sites also use it, but since this is the only one for now I figured it should still stay in the child

PR #174, opened 2018-11-08
Create ingest_info to describe the scraped information.

zacharylawrence:
Created ingest_info to represent the data scraped from a website and defined a common directory for related constants. 

PAIR = Terin


arian487:
This looks good to me, you can go ahead and merge, so we can use it in the generic_extractor :)

recidiviz/ingest/models/ingest_info.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+class IngestInfo(object):
+    def __init__(self):
+        self.people = []  # type: List[Person]
+
+
+class Person(object):
+    def __init__(self):
+        self.last_name = None  # type: str
```
jessex: Localization note that I'm not an expert on: I think `surname` is preferred because it is more generalizable, e.g. to some Latin cultures where the surname is a construction based on maternal and paternal family names with spaces. Some Brazilians might have a surname that is "multiple names" and would be addressed as such. Reducing to just a singular last name would distort the proper reading/ordering of the name, and could potentially cause name matching issues with diverse data sources later on.

zacharylawrence: Great to know, thanks for the clarification.

PR #178, opened 2018-11-12
Data Extractor

jamwalla:
Adds a DataExtractor class that scrapers can use to automatically get data from HTML tables.


recidiviz/ingest/base_scraper.py
```diff
@@ -0,0 +1,501 @@
...
+                self.add_task('_generic_scrape', task_params)
+        # If we don't have any more tasks to scrape, we are at a leaf node and
+        # we are ready to populate the database.
+        else:
+            # Something is wrong if we get here but don't have ingest_info
+            # to work with.
+            if not ingest_info:
+                raise ValueError(
+                    'IngestInfo must be populated if there are no more tasks')
+            # This converter should exist elsewhere, but its in here for now
```
jessex: Do we want to file an issue and place an explicit TODO on this one?

recidiviz/ingest/vendors/archonix/archonix_scraper.py
```diff
@@ -240,6 +240,27 @@ def get_more_tasks(self, content, params):
...
+
+        Args:
+            content: An lxml html tree.
+            params: dict of parameters passed from the last scrape session.
+            ingest_info: The IngestInfo object to populate
+        """
+        data_extractor = DataExtractor(self.yaml_file)
+        data_extractor.extract_and_populate_data(content, ingest_info)
+        full_name = content.cssselect(
+            '[id=ctl00_ContentPlaceHolder1_spnInmateName]')[
```
jessex: Why can't we extract this directly via `DataExtractor`?

recidiviz/ingest/base_scraper.py
```diff
@@ -0,0 +1,501 @@
...
+            ingest_person.birthdate)
+        person.age = int(ingest_person.age)
+        person.region = self.get_region().region_code
+        person.sex = ingest_person.sex
+        person.race = ingest_person.race
+
+        try:
+            person.put()
+        except (Timeout, TransactionFailedError, InternalError) as e:
+            # Datastore error - fail task to trigger queue retry + backoff
```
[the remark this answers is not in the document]
arian487: The comment is very outdated, this block of code is really just very temporary its going to be ripped out when the real DB is in place.

recidiviz/ingest/data_extractor.py
```diff
@@ -0,0 +1,341 @@
...
+TODO: Add more details here in implementation logic when it is done.
+"""
+
+import itertools
+import yaml
+
+from recidiviz.ingest.models.ingest_info import IngestInfo
+
+
+class DataExtractor(object):
```
jessex: Nit: can we place this into its own `recidiviz.ingest.extractor` package? Not only because we may expect more files and classes related to the logic to spin out of here, but because that will make it easier to extract the extractor out into a stand-alone, open-source library.

recidiviz/ingest/vendors/archonix/archonix_scraper.py
arian487: In this specific example there is no key on the page.  It just says the the name and not 'name:' for example so the data extractor wouldn't know what to extract.

recidiviz/ingest/vendors/archonix/archonix_scraper.py
```diff
@@ -240,6 +240,27 @@ def get_more_tasks(self, content, params):
...
+                0].text_content().strip()
+        surname, given_names = full_name.split(',')
+        ingest_info.people[0].given_names = given_names.strip()
```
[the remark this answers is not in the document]
arian487: Yup, its one person one page one task.

recidiviz/ingest/base_scraper.py
arian487: Done:

https://github.com/Recidiviz/recidiviz-data/issues/179

recidiviz/ingest/data_extractor.py
```diff
@@ -0,0 +1,341 @@
...
+        """Gets the value for a given |cell| by checking the cells to the right
+        and below.
+
+        Args:
+            cell: an Element
+        Returns:
+            A str representing the value from an adjacent cell.
+        """
+        if cell is None:
+            # TODO: fall back to manual search for this value
```
jessex: I'd like for us to get into the habit of tracking TODOs we don't intend to pursue in the current PR with an issue that we reference from the TODO statement itself.

recidiviz/ingest/data_extractor.py
```diff
@@ -0,0 +1,341 @@
...
+
+    def _find_cell(self, key):
+        """
+        Args:
+            key: str
+        Returns:
+            The first cell Element that matches the given |key|
+        """
+        for cell in self.cells:
+            if key.strip() in cell.text_content().strip():
```
jessex: I think we should `strip`/normalize the key when we first initialize the extractor, instead of potentially stripping each key many times as we crawl the page.

recidiviz/ingest/data_extractor.py
```diff
@@ -0,0 +1,341 @@
...
+            The cell below or None.
+        """
+        parent = cell.getparent()
+        index = parent.index(cell)
+        next_row = parent.getnext()
+
+        if parent.getparent().tag == 'thead':
+            next_row = parent.getparent().getnext()[0]
+
+        if next_row is not None and \
```
jessex: I'm sure you've thought of ways to try to generalize between these two `get_below` methods. One idea: extract the core logic of checking the `next_row` and then iterating to the next into a method that returns a generator that `yields` up `next_row` or `None`. In `_get_below`, call it a single time and then return the value it yields. In `_get_all_below`, iterate all the way through until you get `None`. I didn't actually try writing that so let me know if it's nonsense.

recidiviz/tests/ingest/data_extractor_test.py
```diff
@@ -0,0 +1,131 @@
...
+    key_mapping_file = os.path.join(os.path.dirname(__file__), key_mapping_file)
+    extractor = DataExtractor(key_mapping_file)
+
+    expected_info = IngestInfo()
+    person = expected_info.create_person()
+    person.race = 'W'
+    booking = person.create_booking()
+
+    charge1 = booking.create_charge()
+    charge1.charging_entity = 'ROCKWALL CO SO ROCKWALL'
```
jessex: I don't see any PII in here from a glance, but want to make sure that there's definitely no PII and that the combination of specific fields doesn't build up to identifiability, i.e. was not taken directly from a live example.

recidiviz/tests/ingest/data_extractor_test.py
```diff
@@ -0,0 +1,131 @@
...
+    html_file = os.path.join(os.path.dirname(__file__), html_file)
+    with open(html_file, 'r') as f:
+        html_contents = html.fromstring(f.read())
+
+    info = extractor.extract_and_populate_data(html_contents)
+
+    # The scraper scrapes a charge row that just says 'Count=3', we need to
+    # know how to ignore this somehow
+    info.people[0].bookings[0].charges.pop()
+    assert info == expected_info
```
jessex: I'd love to also see test cases covering the errors, such as misconfigured entity or field names in the yaml.

recidiviz/ingest/vendors/archonix/archonix_scraper.py
arian487: Yup, James is working on that functionality right now.

recidiviz/ingest/data_extractor.py
arian487: Done.

recidiviz/ingest/data_extractor.py
```diff
@@ -0,0 +1,341 @@
...
+                result.append(next_row[index])
+                next_row = next_row.getnext()
+            else:
+                return None
+        return result
+
+    def get_value(self, key):
+        """Tries to find a value of a given key.
+
+        Args:
```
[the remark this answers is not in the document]
arian487: Fixed.

recidiviz/ingest/data_extractor.py
arian487: Fair point, we've been in 'get this in as fast as possible' mode and have been neglecting some of the process related work.  I'll make sure we are better with this.

recidiviz/ingest/data_extractor.py
arian487: This can definitely be generalized in some but the logic wouldn't be as simple because its possible to have columns that have empties in them, but a real value below the empty.  We want to preserve that order hence the split up logic, for example:

| Bond |
| $100 |
|    ''     |
| $200 |

In this case important for us to collect the rows as they are because otherwise they'd be assigned to the wrong charge.

recidiviz/ingest/data_extractor.py
```diff
@@ -0,0 +1,341 @@
...
+
+        booking = ingest_info.people[0].bookings[0]
+
+        for charge_dict in charges:
+            charge = booking.create_charge()
+            self._populate_charge(charge, charge_dict)
+
+        return ingest_info
+
+    def _extract_rows(self):
```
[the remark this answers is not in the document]
arian487: I am completely rewriting this logic right now to handle multiple people on a page and it cleans this up and makes it more general (and thus handles multiple people + multiple bookings + multiple charges on a single page)

recidiviz/ingest/data_extractor.py
```diff
@@ -0,0 +1,341 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+# =============================================================================
+
+"""A module which takes any content and extracts from it key/value pairs that
+a user might care about.  The extracted information is put into the ingest data
+model and returned.
+
```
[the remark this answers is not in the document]
arian487: Added an issue here: https://github.com/Recidiviz/recidiviz-data/issues/180

recidiviz/ingest/us_pa_greene/us_pa_greene.yaml
```diff
@@ -0,0 +1,21 @@
...
+  Inmate No: person.person_id
+  Gender: person.sex
+  BirthDate: person.birthdate
+  Age: person.age
+  Race: person.race
+  "Booking #": booking.booking_id
+  Committed By: booking.hold
+  Booking Date-Time: booking.admission_date
+
+keys_to_ignore:
```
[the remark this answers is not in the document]
arian487: This is a bit of a temporary thing until the new DB backend is in place.  I didn't want to spend too much time coding an ingestinfo -> db converter that was going to get completely pulled out when the new db is decided on (in the next few days) so I only wrote conversions for the simplest of keys and for data that actually existed on the pa_greene user page.  The issue to write a proper converter is here: https://github.com/Recidiviz/recidiviz-data/issues/179

recidiviz/tests/ingest/data_extractor_test.py
```diff
@@ -0,0 +1,131 @@
...
+    # Add charge information
+    charge = booking.create_charge()
+    charge.statute = '901'
+    charge.name = 'Criminal Attempt [INCHOATE]'
+    charge.case_number = 'CR-145-2018'
+
+    # Add bond information
+    charge.create_bond().amount = '$150,000.00'
+
+
```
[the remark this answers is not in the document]
arian487: Fixed.

recidiviz/ingest/data_extractor.py
```diff
@@ -0,0 +1,341 @@
...
+        Args:
+            cell: the leaf cell we are analyzing
+        Returns:
+            The cell below or None.
+        """
+        parent = cell.getparent()
+        index = parent.index(cell)
+        next_row = parent.getnext()
+
+        if parent.getparent().tag == 'thead':
```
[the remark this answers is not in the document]
jamwalla: Fixed.

recidiviz/ingest/data_extractor.py
jamwalla: I think the generator is a good idea! Instead of stopping at `None`, we can yield at each row.

recidiviz/ingest/data_extractor.py
```diff
@@ -0,0 +1,341 @@
...
+            next_row = parent.getparent().getnext()[0]
+
+        result = []
+        while next_row is not None:
+            if next_row.tag == 'tr' and index < len(next_row):
+                result.append(next_row[index])
+                next_row = next_row.getnext()
+            else:
+                return None
```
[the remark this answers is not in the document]
jamwalla: I'm not sure - `index < len(next_row)` asks if we can safely get the `index`th element from the row - so `next_row` should always be a `tr` with at least `index` cells inside of it. Does that make sense?

Combining this thought with the generator conversation, though - I'm committing a change that advances `next_row` regardless of this check, and just doesn't yield if the thing we think should be a row isn't a row. Alternatively, we could throw an exception. I'm not sure what the correct behavior would be if a `tr` was followed by a different kind of element - it should always be enclosed in a `table` or `tbody`, right?

recidiviz/ingest/data_extractor.py
jamwalla: good call!

recidiviz/ingest/data_extractor.py
```diff
@@ -0,0 +1,341 @@
...
+        elif class_to_set == 'arrest':
+            if not ingest_info.people:
+                ingest_info.create_person()
+            if not ingest_info.people[0].bookings:
+                ingest_info.people[0].create_booking()
+            if not ingest_info.people[0].bookings[0].arrest:
+                ingest_info.people[0].bookings[0].create_arrest()
+            obj = ingest_info.people[0].bookings[0].arrest
+
+        setattr(obj, ingest_key, value)
```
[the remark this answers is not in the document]
jamwalla: I don't think `setattr` will raise the error, but you're probably right that it's worth checking `hasattr` before setting the field. Added to `_set_value` and `_populate_charge`.

recidiviz/ingest/data_extractor.py
```diff
@@ -0,0 +1,341 @@
...
+                ingest_info.create_person()
+            obj = ingest_info.people[0]
+        elif class_to_set == 'booking':
+            # Create a booking if needed.
+            if not ingest_info.people:
+                ingest_info.create_person()
+            if not ingest_info.people[0].bookings:
+                ingest_info.people[0].create_booking()
+            obj = ingest_info.people[0].bookings[0]
+        elif class_to_set == 'arrest':
```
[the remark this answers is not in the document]
jamwalla: Done.

recidiviz/ingest/data_extractor.py
```diff
@@ -0,0 +1,341 @@
...
+        |extract_and_populate_data| since that method sets up parsing the
+        content.
+
+        Returns:
+            A list of dictionaries, where each dictionary has keys from the
+            schema that map to values.
+        """
+        lists_of_values = []
+        for key, schema_key in self.multi_keys.iteritems():
+            values_below = self._get_values_below(key) or []
```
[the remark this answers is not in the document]
jamwalla: I can't think of any, so let's return the empty list instead of `None`!

recidiviz/ingest/data_extractor.py
jamwalla: Filed #181.

recidiviz/ingest/data_extractor.py
```diff
@@ -0,0 +1,341 @@
...
+            values = [
+                {schema_key: value} for value in values_below
+            ]
+            lists_of_values.append(values)
+
+        charges = []
+        lengths = set(len(l) for l in lists_of_values)
+        # The charges columns should all have the same length
+        if len(lengths) == 1:
```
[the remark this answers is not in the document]
jamwalla: Added a ValueError here, does that sound right?

recidiviz/tests/ingest/data_extractor_test.py
```diff
@@ -0,0 +1,131 @@
...
+    html_file = 'testdata/data_extractor/html/nested_good_table.html'
+    html_file = os.path.join(os.path.dirname(__file__), html_file)
```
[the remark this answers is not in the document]
jamwalla: os.path.join returns a string, wdyt? it's still easy to change!

recidiviz/ingest/data_extractor.py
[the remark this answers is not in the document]
jessex: Yes, `tr` should always be in a `table` or `tbody` I'd think. Not sure if we can expect that basic pattern to ever break. I was thinking about if there's a non-`tr` element as a sibling, e.g. a `div`, but as you point out it shouldn't get picked up regardless.

recidiviz/ingest/data_extractor.py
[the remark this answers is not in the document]
jessex: Looks good!

recidiviz/tests/ingest/data_extractor_test.py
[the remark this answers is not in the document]
jessex: Oh you right, they're the same type. Still don't love variable reuse _generally_ but that's not a hard and fast rule, and I encourage you to follow your stylistic instinct on this one!

recidiviz/tests/ingest/data_extractor_test.py
jessex: I suppose if we wanted to be really slick we could derive which fields in a page are PII we'd want to mask before using it as a fixture from looking at those keys which map to PII fields in our schema, and providing some tool that auto-scrubs any values with those keys, but it feels like a lot of effort... There's still a need to define what testing of scrapers _should_ look like, @arian487.

jamwalla:
thanks @jessex!! I'm going to merge this so Arian can upload his changes with multiple people - let me know if there's anything outstanding here.

jessex:
I think you're good to go, @jamwalla! Feel free to delete the branch unless @arian487 needs it.

PR #184, opened 2018-11-14
Create constructors for IngestInfo.

zacharylawrence:


terinpw:
Looks good - but i'd do a rebase from master as ingest_info.py had a big update with the generic data extractor pr

PR #186, opened 2018-11-14
Restructure the data extractor logic to handle multiple people as well as many objects

arian487:
This changes the logic completely to loop over cells instead and build the object as we read the page.  It will intelligently find the parent and know whether or not to add to the most recent object or an object somewhere else in the tree.  This can handle arbitrarily designed tables.


recidiviz/ingest/extractor/data_extractor.py
```diff
@@ -99,128 +115,105 @@ def extract_and_populate_data(self, content, ingest_info=None):
...
 
+        for cell in self.cells:
+            # This is a tiny hack to avoid an O(n) search over the keys list for
+            # every cell.
+            # An alternative approach is to force the user to give the exact key
+            # with a semi colon in the yaml file, but that might be confusing.
+            # Finally, we could preprocess the keys mapping to include multiple
+            # keys that map to the same value ('hi' and 'hi:' both map to the
+            # same thing) but that is a more expensive preprocessing calculation
+            cell_val = cell.text_content().strip().strip(':').strip()
```
jessex: Do we need the first `strip` in this chain? It looks like you're trying to get a colon out of the way, and then get leading/trailing whitespace out of the way. So I'd think that the first strip is redundant and the third strip is what's key.

recidiviz/ingest/extractor/data_extractor.py
```diff
@@ -99,128 +115,105 @@ def extract_and_populate_data(self, content, ingest_info=None):
...
+                values = [self._get_value_cell(cell)]
+            elif cell_val in self.multi_keys:
+                values = self._get_values_below_cell(cell)
+            if values:
+                key_mapping_val = (self.keys.get(cell_val) or
+                                   self.multi_keys.get(cell_val))
+                value_classes = key_mapping_val.split('.')
+                ingest_key = value_classes[1]
+                class_to_set = value_classes[0]
+                if values:
```
jessex: Wouldn't this always evaluate to `True` since it's nested inside the exact same check and nothing changes the value in the interim?

recidiviz/ingest/models/ingest_info.py
```diff
@@ -17,34 +17,44 @@
...
-from recidiviz.common.constants.booking import CustodyStatus
-from recidiviz.common.constants.booking import JurisdictionStatus
-from recidiviz.common.constants.booking import ReleaseReason
-from recidiviz.common.constants.charge import ChargeClass
-from recidiviz.common.constants.charge import ChargeStatus
-from recidiviz.common.constants.charge import CourtType
-from recidiviz.common.constants.charge import Degree
-from recidiviz.common.constants.person import Ethnicity
-from recidiviz.common.constants.person import Race
+# from recidiviz.common.constants.bond import BondStatus
```
jessex: What's the story here? Use 'em or lose 'em!

recidiviz/ingest/models/ingest_info.py
```diff
@@ -17,34 +17,44 @@
...
     """Class for information about multiple people."""
 
     def __init__(self):
-        self.people = []  # type: List[Person]
+        self.person = []  # type: List[Person]
 
     def __eq__(self, other):
         return self.__dict__ == other.__dict__
 
+    # def __str__(self):
```
jessex: Same here. Do we need this?

recidiviz/ingest/models/ingest_info.py
```diff
@@ -17,34 +17,44 @@
...
+# from recidiviz.common.constants.person import Ethnicity
+# from recidiviz.common.constants.person import Race
 
 
 class IngestInfo(object):
     """Class for information about multiple people."""
 
     def __init__(self):
-        self.people = []  # type: List[Person]
+        self.person = []  # type: List[Person]
```
jessex: Why the singular name for the array type?

recidiviz/tests/ingest/testdata/data_extractor/html/multiple_people_sometimes_charges.html
```diff
@@ -0,0 +1,1079 @@
...
+                    <table>
+                        <tbody><tr>
+                            <td>
+                                <div id="JailInfo">
+						        <table border="0" cellpadding="0" cellspacing="2" width="100%" class="JailView">
+							        
+								        
+									        <tbody><tr style="InmateRecordRow">
+										        <td nowrap="nowrap" width="2%">
+                                                    <img src="ViewImage.aspx?bookno=BCSO09JBN002027"><br><a href="ViewImageFull.aspx?bookno=BCSO09JBN002027" target="blank">Enlarge Photo</a>
```
jessex: Are these booking numbers sensitive/real? There are three `a href`'s in here that I'm curious about.

recidiviz/tests/ingest/testdata/data_extractor/html/multiple_people_sometimes_charges.html
```diff
@@ -0,0 +1,1079 @@
...
+														        </td>
+														        
+														        
+													        </tr>
+													        <tr class="oddNoWidth">
+														        <td class="InmateInfoGridTd">
+															        Booking No:
+														        </td>
+														        <td class="InmateInfoGridTd">
+															        BCSO09JBN002027
```
jessex: What about these id-looking fields that show up in some places in the file?

recidiviz/ingest/extractor/data_extractor.py
```diff
@@ -99,128 +115,105 @@ def extract_and_populate_data(self, content, ingest_info=None):
...
+            parent = self._find_parent_ingest_info(
+                ingest_info, self.hierarchy_map[class_to_set], i)
+
+            get_recent_name = 'get_recent_' + class_to_set
+            class_to_use = getattr(parent, get_recent_name)()
+            # If we are in multikey, we need to pull out the correct index of
+            # class type we want to set
+            if is_multi_key:
+                attr = getattr(parent, class_to_set)
+                if attr != None and isinstance(attr, list) and len(attr) > i:
```
jessex: Would we consider it an error condition at this point for this check to fail? It seems like at least the first two clauses would be considered erroneous, and maybe the third as well.

recidiviz/ingest/extractor/data_extractor.py
```diff
@@ -99,128 +115,105 @@ def extract_and_populate_data(self, content, ingest_info=None):
...
-            elif class_to_set == 'bond':
-                obj = bond
+        is_multi_key = class_to_set in self.multi_key_classes
+        for i, value in enumerate(values):
+            # The first task is to find the parent.
+            parent = self._find_parent_ingest_info(
+                ingest_info, self.hierarchy_map[class_to_set], i)
+
+            get_recent_name = 'get_recent_' + class_to_set
+            class_to_use = getattr(parent, get_recent_name)()
```
jessex: Unless I'm misreading this, it looks like we're actually calling the `get_recent_*` function on the `ingest_info`, which returns an actual instance and not a class itself, right? If so, calling this `class_to_use` is a bit misleading and makes it harder to follow the flow.

recidiviz/ingest/extractor/data_extractor.py
```diff
@@ -99,128 +115,105 @@ def extract_and_populate_data(self, content, ingest_info=None):
...
+                    class_to_use = attr[i]
+
+            create_name = 'create_' + class_to_set
+            create_func = getattr(parent, create_name)
+            # If the object we are trying to operate on is None, or it has
+            # already set the ingest_key then we know we need to create a
+            # new one.
+            if (class_to_use is None or
+                    getattr(class_to_use, ingest_key) != None):
+                obj_to_set = create_func()
```
jessex: I don't love abbreviations for the sake of shorter var names, which is how I think `obj` reads. But that's a matter of style to consider or reject as you wish!

recidiviz/ingest/extractor/data_extractor.py
```diff
@@ -99,128 +115,105 @@ def extract_and_populate_data(self, content, ingest_info=None):
...
-    def _set_value(self, ingest_info, ingest_key, value):
-        """This function intelligently sets the value in the ingest_info
-        object if it can.
-
-        Args:
-            ingest_info: The ingest_info object to set the value on.
-            ingest_key: The key we wish to set.
-            value: The value to set the key to
+    def _set_or_create_object(
+            self, ingest_info, class_to_set, ingest_key, values):
```
jessex: I think given the fairly dynamic logic in here--i.e. we're working with functions and function results, classes and object instances--it would be helpful to have `Args` in this docstring, even if the function itself is private.

recidiviz/ingest/extractor/data_extractor.py
```diff
@@ -99,128 +115,105 @@ def extract_and_populate_data(self, content, ingest_info=None):
...
+
+    def _find_parent_ingest_info(self, ingest_info, hierarchy, val_index):
+        """Find the parent object to set the value on
 
-    def _find_cell(self, key):
-        """
         Args:
-            key: str
+            ingest_info: an IngestInfo object
+            hierarchy: the hierarchy of of the data structure to help us
```
jessex: Extra `of`

recidiviz/ingest/extractor/data_extractor.py
```diff
@@ -63,6 +73,10 @@ def initialize_extractor(self, keys, multi_keys, ignored_keys):
         self.keys = {k.strip():v for k, v in keys.iteritems()}
         self.multi_keys = {k.strip():v for k, v in multi_keys.iteritems()}
         self.all_keys = self.keys.keys() + self.multi_keys.keys() + ignored_keys
+        # We want to know which of the classes are multi keys as this helps
+        # us with behaviour when we set the values.
+        self.multi_key_classes = set(
+            [value.split('.')[0] for value in self.multi_keys.values()])
```
jamwalla: do we want an error here if a class in self.keys is in multi_key_classes?

recidiviz/ingest/extractor/data_extractor.py
```diff
@@ -99,128 +115,105 @@ def extract_and_populate_data(self, content, ingest_info=None):
...
+        """Find the parent object to set the value on
 
-    def _find_cell(self, key):
-        """
         Args:
-            key: str
+            ingest_info: an IngestInfo object
+            hierarchy: the hierarchy of of the data structure to help us
+                walk the data object.
+            val_index: the index of the value we are setting, note that this is
```
jessex: It might just be that I'm not close enough to the code yet, but this highlights something about `IngestInfo` I don't totally understand yet: it has arrays of member variables that we need to iterate through and/or index into in order to find which object to set values on. It feels like that's a bit fragile: a tiny bug (such as an off-by-one) could mean we wind up setting a value on the wrong booking or the wrong arrest, say, which would be very hard to detect and debug.

Do you think my concerns here are misplaced? If not, would it make sense to have a dict from id to entity and update the logic in the data extractor to figure out the id of the entity it should be setting a value on?

recidiviz/ingest/extractor/data_extractor.py
arian487: I've seen sites that have:

'Key : '

So the first strip is needed for leading whitespace, the next strip gets rid of colons, and the last strip gets rid of any whitestace once the colon is removed.

recidiviz/ingest/extractor/data_extractor.py
arian487: Good catch, hold over from older code.  Done.

recidiviz/ingest/models/ingest_info.py
arian487: Lose em!  I decided its best to have the intermediate object reflect exactly what we scrape from the page.  This makes testing a lot easier as the user can just check the website and not care about enums or conversions.

recidiviz/ingest/models/ingest_info.py
arian487: Removed.

recidiviz/ingest/models/ingest_info.py
arian487: If you look at the extractor code, this lets us avoid any annoying mappings required in the data extractor, and lets us dynamically find objects by their parents using the names.  The code would otherwise be a lot more annoying to write as we'd have to keep some type of mapping between whats in the yaml, and what exists internally.

recidiviz/tests/ingest/testdata/data_extractor/html/multiple_people_sometimes_charges.html
arian487: I removed those ids.

recidiviz/tests/ingest/testdata/data_extractor/html/multiple_people_sometimes_charges.html
arian487: removed all ids and used placeholders

recidiviz/ingest/extractor/data_extractor.py
arian487: I'm not sure what you mean exactly, if this fails we could have a bond for example, which is not a type list but just a single object.  It could be set in a previous multi key run (multiple columns for same object), in which case the get_most_recent is what we want.  If it is None, it'll fall back to the logic below which is in charge of creating the object if needed.  And finally if it is of type list, we check the length to again let us know if we want the below logic to create a new one for us (going down a list of charges on the first column).  For example say we have a table like this:

| Statute | | Description | 
| Value    | | Value           |
| Value2  | | Value2         |

If we are in the first column, we need to be creating new charges on the parent booking as we go down the line but once we hit the description column, the parent booking will have the correct number of elements in it and therefore we index on the index of the value to attach the description to the correct booking.  Doing it by columns lets us handle cases where data is displayed in multiple tables for the same person.  For example I saw a pasge that had a charges table and a corresponding bond table of the same length.

recidiviz/ingest/extractor/data_extractor.py
arian487: Fixed

recidiviz/ingest/extractor/data_extractor.py
arian487: Problem is we dont have entity ids to work through at extraction time.  When analyzing the page we simply have cells to work through so if we hit a cell that has charges in it for example, how do we know what booking id that charge belongs to?  On the page, it is inherently tied to the order.  We can have a 'cur_class' object for example which is what get_recent_* accomplishes.  "If I'm not in multi key mode, always operate on the current version of the class" which again is reliant on read order.  Similarly when its a multi key, if I had a list of charges for example, and I read a column of data I can't know which object to attach the data to since most pages don't have html ids attached to their rows (or unique ones at least).

I thought about this a fair bit, and I think my position is that if a page breaks the assumptions made in this code about ordering and web-page 'saneness' then it should be up to the user to custom scrape the data (or update the assumptions).  If for example a page lists a person with two tables, one of which has 5 charges and another has 2 bonds in it, how would a human even interpret that?  Furthermore if its ordered in such a way to cause an off by one (two persons data displayed, and then their charges displayed later in separate tables) that again makes me suspicious about what it means for the data saneness on a webpage.  

Long story short, each user should write unit tests about how they expect the data to be output (like our extractor tests) and its good if those expectations break or cause exceptions.  We want to be notified of weirdly designed sites to either update the assumptions in the extractor or choose not to.

recidiviz/ingest/extractor/data_extractor.py
arian487: Good catch I actually meant to write that I forgot!

recidiviz/ingest/extractor/data_extractor.py
arian487: Done.  Cleaned up the code here a bit.

recidiviz/ingest/extractor/data_extractor.py
arian487: Done.

recidiviz/ingest/models/ingest_info.py
jessex: Ah, I understand. It's working because it's coupling the `get_recent_*` method names to the names of the entity classes themselves. That's clever but let's make sure we've got good test coverage/documentation such that if any names change on either end, we don't silently fail.

recidiviz/ingest/extractor/data_extractor.py
jessex: Oh I see. I was thinking that `.strip(':')` would just remove any colons, but it would only be leading or trailing colons, and we wouldn't even necessarily want to remove any colons anyway. So this makes perfect sense. Thanks for the example!

recidiviz/ingest/extractor/data_extractor.py
jessex: That's the explanation I needed, thanks! I was basically wondering if we should be raising an error from an `else` statement for this block, and your explanation shows why we don't. Because that's an expected case and we want to move onto the below logic.

recidiviz/ingest/models/ingest_info.py
arian487: Agreed.

recidiviz/ingest/extractor/data_extractor.py
arian487: Yea its hard to conceptualize the reasons behind everything until you see all the annoying examples!

recidiviz/ingest/extractor/data_extractor.py
jessex: I think that all sounds correct and proper, and I buy your line of thinking. I think the jury will remain out as we continue to scale outward, though, given that some of the conclusions in the comment boil down to, "the person writing the scraper will need to handle it." That may be unavoidable in some cases, but as we know, the closer we can get to _magic_ with the data extractor just handling everything, the more effective we will be at maintaining a full fleet of scrapers indefinitely. 

Thanks a lot for this comment!

recidiviz/ingest/extractor/data_extractor.py
arian487: Yea sorry about this whole file its fairly intricate logic that is hard to follow, and took me a while to get right.

recidiviz/ingest/extractor/data_extractor.py
arian487: I completely agree, I was more saying that I want the data scraper to be in a place where if assumptions about the structure are broken, it indicates poorly designed site to the point where even a human will have trouble constructing the object.  Any time a human can look and very easily see whats what, those assumptions should be baked into the data extractor and handled.  

I think with this change its in a place where if any of the assumptions about table data break than we can't possibly know what to do (like 5 charges + 2 bonds in separate tables...which I've seen; wouldn't know what to do here as a human either).

Issue #176, opened 2018-11-12
Migrate to Python 3


colincadams:
A lot of the libraries and functionality we want to use require python 3:
* google-cloud-monitoring (for #59)
* enums
* data classes

This would also help reduce the size of /libs (#45)

There are a few blockers:
* app engine task queue -> cloud tasks & pub/sub (discussed some in #88, see [migration docs](https://cloud.google.com/tasks/docs/migrating))
  * or investigate moving docket to a push queue so we don't need pub/sub
* webapp2 (per [these docs](https://cloud.google.com/appengine/docs/standard/python/tools/webapp2)) -> ? (flask?)
* ndb orm -> sqlserver
* others?

colincadams:
We also need to migrate off of users, likely to cloud IAM https://cloud.google.com/appengine/docs/standard/python3/python-differences#users

and replace urlfetch with regular requests
https://cloud.google.com/appengine/docs/standard/python3/python-differences#url_fetch

jessex:
Looks like the only place we use GAE users is in `auth.py`, so that should be a straight-forward migration. Going from `urlfetch` to `requests` shouldn't be hard, either.

colincadams:
Yeah based on my reading of the auth stuff, it looks like we can turn on cloud iap and then we just need to verify the user by checking the jwt header:
https://cloud.google.com/iap/docs/signed-headers-howto

We would create service accounts for things like cloud tasks so that their requests get through the proxy. Not sure how crons work, assuming they run as a service account?

@jessex, Do you see any issues with this?

jessex:
No issues off of the top of my head. Service accounts seems like the right approach! Looking at the IAM console, I don't _think_ Crons work as a service account. I think the `auth.py` code indicates how they work, which is simply that the cron service built into GAE [identifies itself with a particular header](https://cloud.google.com/appengine/docs/standard/python/config/cron#securing_urls_for_cron).

jessex:
It doesn't look like any of those calls are happening at the moment, @colincadams. I searched Stackdriver for instances of "Requester authenticated as app-id" throughout all time and found none.

colincadams:
@rasmi pointed out that we also need to migrate off memcache. Trying to determine whether we can use memorystore (gcp managed redis instance) or need to use one managed by redis labs.

jessex:
[...]

PR #189, opened 2018-11-14
Finish migration to flask, delete webapp2 (part of #176)

colincadams:


colincadams:
Yepp ran it locally and hit all the endpoints as a sanity check!

PR #190, opened 2018-11-14
Create a persistence layer

zacharylawrence:
This PR is the first pass at laying out our persistence layer. It includes:
1. A new persistence layer used to talk to a MS SQL Server.
2. An in-memory fake sqlite3 database for testing.

@jessex I'm new to Python - so please feel free to be heavy handed with any style/design related opinions.

I focused on following the guiding principles under: https://en.wikipedia.org/wiki/Dependency_inversion_principle. Basically, this meant creating a persistence layer blind to the specific database implementation (ie. _ms_sql_server package).

Note: this implementation currently use a sqlite3 in-memory database for testing because an in-memory MS SQL Server implementation doesn't exist. This leads to _override_sqlserver_tables_to_sqlite() directly mutating table constants with near duplicates (eg. change SQL Server "money" type to a sqlite supported "int"). I'm not sure that this is the best design. It means we're using a different database in prod/testing and it leads to a duplicated table spec. I suspect that this can be cleaned up in the future, but decided to leave this for now to move forward quickly. Open to any suggestions on this!


recidiviz/persistence/_ms_sql_server/__init__.py
```diff
@@ -0,0 +1,21 @@
+# Recidiviz - a platform for tracking granular recidivism metrics in real time
```
arian487: Why is this preceded by an underscore?

recidiviz/persistence/_ms_sql_server/ms_sql_server.py
```diff
@@ -0,0 +1,46 @@
...
+class MsSqlServer(object):
+    def __init__(self, engine):
+        """
+        Use the provided MS SQL Server database engine to perform reads/writes.
+
+        :param engine: The MS SQL Server sqlalchemy database engine to use.
+        """
+        self.engine = engine
+
+    def write_name(self, name):
```
arian487: This is definitely going to have to be revamped.  There needs to be something that actually interprets the ingest info object and converts that to the db objects.  A flat write like this will have problems when it comes to understanding what to write.  Unless this is just a temporary thing in place for testing?

arian487: I read a bit further I see you have that in the persistence function.  You'll likely need something that builds a query/object that can be put into the database at once.

recidiviz/persistence/persistence.py
```diff
@@ -0,0 +1,38 @@
+# Recidiviz - a platform for tracking granular recidivism metrics in real time
```
arian487: Since the purpose of this file is to persist data tothe data base and communicate with the database layer, where does conversion come in?  Is this code in charge of calling convert on the ingest_info object which goes and does things like take the age and make it into a birthdate?

recidiviz/persistence/persistence.py
```diff
@@ -0,0 +1,38 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+# =============================================================================
+"""Contains logic for communicating with the persistence layer."""
```
arian487: This is in charge of communicating with the database layer, not the persistence layer correct?

recidiviz/tests/persistence/persistence_test.py
```diff
@@ -0,0 +1,57 @@
...
+class TestPersistence(object):
+    def setup_method(self, _test_method):
+        self.fake_database = fakes.create_in_memory_sqlite_database()
+
+    def test_write_surname(self):
+        # Arrange
+        persistence = Persistence(self.fake_database)
+
+        ingest_info = IngestInfo()
+        ingest_info.people.append(_Person(surname='test_surname'))
```
arian487: use ingest_info.create_person().

arian487: If you want you can add the kwargs to that function so avoid having another line that does person.surname.  Though I think its cleaner that way :)

recidiviz/persistence/_ms_sql_server/__init__.py
zacharylawrence: I wanted to denote that this package shouldn't be used outside of the recidiviz/persistence package. Is this a bad way of doing that? Took this idea from: https://stackoverflow.com/questions/3602110/python-private-module-in-a-package

recidiviz/persistence/_ms_sql_server/ms_sql_server.py
zacharylawrence: Right, I agree. I just set this method up to test with while outlining the file layout/structure I was envisioning. We'll certainly need to built a separate "convert ingest_info"/"write everything" layer.

recidiviz/tests/utils/fakes.py
```diff
@@ -0,0 +1,133 @@
...
+    :return The engine used to communicate with the fake database
+    """
+    _override_sqlserver_tables_to_sqlite()
+    engine = sqlalchemy.create_engine('sqlite:///:memory:')
+    _tables.METADATA.create_all(engine)
+
+    return engine
+
+
+def _override_sqlserver_tables_to_sqlite():
```
zacharylawrence: Do you have thoughts on directly overriding the _tables.* constants like this? Would it make more sense to just pass these into the Persistence object as a dependency?

recidiviz/persistence/persistence.py
zacharylawrence: I was envisioning a separate class/package in the recidiviz/persistance package that would do that conversion. The exact structure of that would be dependent on if we decide to go with an ORM vs Fluent SQL Expression approach. I'll try to outline this next and can hold off on this PR until then if you want.

recidiviz/persistence/persistence.py
```diff
@@ -0,0 +1,38 @@
...
+class Persistence(object):
+    def __init__(self, engine):
+        """
+        Setup persistence communication via MS SQL Server.
+
+        :param engine: The sqlalchemy database engine to use.
+        """
+        self._database = MsSqlServer(engine)
+
+    def write_surname(self, ingest_info):
```
zacharylawrence: @arian487 Just to be clear, at the end of the day I'm imaging this to convert then write the entire ingest_info object. Just using this for testing incremental changes for now.

recidiviz/persistence/persistence.py
zacharylawrence: Oh, I intend for the persistence layer to expose only a single "write ingest_info" method. This layer would involve doing the required conversions and communicating with a database layer that performs all the reads/writes necessary.

recidiviz/tests/persistence/persistence_test.py
zacharylawrence: Like this?

zacharylawrence:
I'm also noticing "ImportError: No module named sqlalchemy" on Travis. I updated the piplock to add this dependency, but should I also update requirements.txt? or something else?

recidiviz/persistence/persistence.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def write(self, ingest_info):
+        """
+        Persist each person in the ingest_info. If a person with the given
+        surname/birthday already exists, then skip that person.
+
+        :param ingest_info: The IngestInfo containing the surname.
+        """
+        for person in ingest_info.people:
+            person_already_exists = self._database.read_people(person.surname,
```
zacharylawrence: @arian487 Updated this with a better test case to show a basic example with logic in the Persistence class.

recidiviz/tests/persistence/persistence_test.py
```diff
@@ -0,0 +1,82 @@
...
+SURNAME_2 = 'test_surname_2'
+BIRTHDATE_1 = datetime(year=1993, month=11, day=15)
+BIRTHDATE_2 = datetime(year=1996, month=2, day=11)
+
+
+class TestPersistence(object):
+    def setup_method(self, _test_method):
+        self.fake_database = fakes.create_in_memory_sqlite_database()
+
+    def test_twoDifferentPeople_persistsBoth(self):
```
zacharylawrence: Playing around with test naming. Thoughts on this?

recidiviz/persistence/_ms_sql_server/__init__.py
arian487: Its definitely not a pattern i've really seen at google.  Its not horrible, but I think its a bit unnecessary.  Up to you.

recidiviz/persistence/_ms_sql_server/ms_sql_server.py
arian487: Sounds good, we can work on that design a bit tomorrow.

recidiviz/persistence/persistence.py
arian487: No this is fine.  For things in active development that 100% won't be used in prod, I am always in favour of getting code in and iterating.  You risk over-designing otherwise :)

recidiviz/persistence/persistence.py
arian487: Gotcha, I misunderstood the comment.

recidiviz/tests/persistence/persistence_test.py
arian487: Yea, but I feel you should still use the create_method.  I'm in favour of creating a separation between what the public sees, and what the object does in the background.  I can envision scenarios where the create_* function does a few more complicated tasks than creating the object and adding it.  I'd prefer to deter users from accessing the people/object attribute directly and appending.

ingest_info.create_person(surname='test_surname')

This allows us to easily migrate to a future where data object creation does more work if needed and is much cleaner IMO.

recidiviz/tests/utils/fakes.py
arian487: I think I'd want this to be restructured so that ms_sql_server has a creator in the init and that exposes some creation method which takes in the engine string and tables.  It doesn't even need to be called ms_sql_server it seems mostly backend agnostic.  That way you can sort of hide away all of this logic, its generally bad practice to override module level constants like you are here (it actually overrides them for everyone).

recidiviz/persistence/persistence.py
arian487: Sounds good, I agree.

recidiviz/tests/persistence/persistence_test.py
arian487: Python doesn't have camel case at all.  test_two_different_* is the way to go.

arian487: I should say, python doesn't use it except for class names.

recidiviz/tests/persistence/persistence_test.py
zacharylawrence: Oh, I think that's what I did. The github preview box just shows a preview of the version when you left the comment. If you click through to the file, it shows: ingest_info.create_person(surname=SURNAME_1)

recidiviz/tests/utils/fakes.py
zacharylawrence: Yeah, I agree I don't like having the modules overridden like this. The ms_sql_server depends on table constants that are specific for the sql server implementation (eg. the Money type).

recidiviz/persistence/_ms_sql_server/ms_sql_server.py
zacharylawrence: Resolved offline

recidiviz/persistence/persistence.py
zacharylawrence: sgtm

recidiviz/tests/utils/fakes.py
zacharylawrence: Updated to use a single schema

recidiviz/persistence/sql_database/sql_database.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def write_person(self, surname=None, birthdate=None):
+        """
+        Write the provided person as a new row to the database.
+
+        :param surname: The name to persist.
+        :param birthdate: The birthdate to persist.
+        """
+        connection = self._engine.connect()
+        connection.execute(tables.PEOPLE.insert(),
```
arian487: You can just use PEOPLE here right?

zacharylawrence: Done

recidiviz/persistence/persistence.py
```diff
@@ -0,0 +1,42 @@
...
+        """
+        Persist each person in the ingest_info. If a person with the given
+        surname/birthday already exists, then skip that person.
+
+        :param ingest_info: The IngestInfo containing the surname.
+        """
+        for person in ingest_info.person:
+            person_already_exists = self._database.read_people(person.surname,
+                                                               person.birthdate)
+            if person_already_exists:
```
jessex: Interface style point: I know this is a first pass, but I'd say that in general I like my data layer to expose boolean-returning methods in cases where you want to know only if some entity exists, instead of sharing methods with callers that want the actual entity, i.e. `does_person_exist(surname, birthdate)` and `get_people(surname, birthdate)` as separate methods. Less because it's a minor performance optimization and more because I think it makes for more imminently readable code.

recidiviz/persistence/sql_database/tables.py
```diff
@@ -0,0 +1,112 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+# =============================================================================
+"""
+Define tables in our SQL Database.
```
jessex: General question: I can go find this out myself, but does sqlalchemy have a database migration/patch utility built-in? Something like Flyway, but for Python.

recidiviz/persistence/sql_database/tables.py
```diff
@@ -0,0 +1,112 @@
...
+                  Column('is_life', Boolean),
+                  Column('is_probation', Boolean),
+                  Column('is_suspended', Boolean),
+                  Column('fine', Integer),
+                  Column('parole_possible', Boolean),
+                  Column('post_release_supervision_length_days',
+                         Integer),
+                  Column('concurrent_with',
+                         ForeignKey('sentences.sentence_id')),
+                  Column('consecutive_wth',
```
jessex: Whoops: `consecutive_with`

recidiviz/persistence/sql_database/tables.py
```diff
@@ -0,0 +1,112 @@
...
+               Column('scraped_person_id', Integer),
+               Column('surname', String),
+               Column('given_names', String),
+               Column('birthdate', DateTime),
+               Column('birthdate_inferred_from_age', Boolean),
+               Column('race', Enum(Dummy)),
+               Column('ethnicity', Enum(Dummy)),
+               Column('place_of_residence', String))
+
+BOOKINGS = Table('bookings', METADATA,
```
jessex: Should this also have a `person_id` foreign key?

jessex:
As discussed offline, thanks for going ahead and merging this! Looks great, @zacharylawrence. I added a few comments which you may have already caught in your follow-up PR.

recidiviz/persistence/sql_database/tables.py
zacharylawrence: Done in #197

recidiviz/persistence/sql_database/tables.py
zacharylawrence: This is updated in #197

recidiviz/persistence/persistence.py
zacharylawrence: #197 removes check specifically and is updated to actually use the value returned from reading the database.

recidiviz/persistence/sql_database/tables.py
jessex: Awesome, those both look like tools I've used in the past, especially Alembic. That's a much better online migration story than what I'd found for Cloud Datastore.

PR #192, opened 2018-11-15
Pretty print ingestion info classes and small cleanup

terinpw:


recidiviz/ingest/models/ingest_info.py
```diff
@@ -16,13 +16,31 @@
...
+
+
+@pretty_print
 class IngestInfo(object):
     """Class for information about multiple people."""
 
-    def __init__(self):
-        self.person = []  # type: List[Person]
+    def __init__(self, person=None):
+        self.person = [] if person is None else person # type: List[Person]
```
arian487: You can just say:

self.person = person or []

Also the actual attribute doesn't need to be person, so you can call it people to be more descriptive to the user.

terinpw: Makes sense, done.

jessex:
Just for my own edification, would you mind pasting an example print output of an `IngestInfo` object (without any PII, of course)?

recidiviz/ingest/models/ingest_info.py
```diff
@@ -16,17 +16,32 @@
...
-    def __init__(self):
-        self.person = []  # type: List[Person]
+    def __init__(self, people=None):
+        self.person = people or [] # type: List[Person]
 
     def __eq__(self, other):
         return self.__dict__ == other.__dict__
 
+    def __str__(self):
+        return to_string
```
jessex: Should this not be `return to_string()`? As is, it looks like it returns an instance of the function itself, not its output.

jessex:
Do we [also want to](https://stackoverflow.com/questions/1436703/difference-between-str-and-repr) implement `__repr__` while we're at it?

terinpw:
> Just for my own edification, would you mind pasting an example print output of an `IngestInfo` object (without any PII, of course)?

Yup, example:

IngestInfo:
   person[0]: _Person:
      age: 100
      sex: M
      race: W
      booking[0]: _Booking:
         admission_date: 1-1-2048- 12:30 am
         total_bond_amount: $1000.00
         booking_id: 123
         charge[0]: _Charge:
            name: Charge 1
         charge[1]: _Charge:
            name: Charge 2
         arrest: _Arrest:
            agency: Agency

terinpw:
Oops, that didn't keep the formatting. here it is

```
IngestInfo:
   person[0]: _Person:
      age: 100
      sex: M
      race: W
      booking[0]: _Booking:
         admission_date: 1-1-2048- 12:30 am
         total_bond_amount: $1000.00
         booking_id: 123
         charge[0]: _Charge:
            name: Charge 1
         charge[1]: _Charge:
            name: Charge 2
         arrest: _Arrest:
            agency: Agency
```

recidiviz/ingest/models/ingest_info.py
```diff
@@ -257,5 +257,4 @@ def __eq__(self, other):
         return self.__dict__ == other.__dict__
 
     def __str__(self):
-        return to_string
-
+        return to_string(self)
```
jessex: Looks like we're missing a newline here, I think?

recidiviz/ingest/models/ingest_info.py
```diff
@@ -16,17 +16,32 @@
 # =============================================================================
 
 """Represents data scraped for a single individual."""
+import types
+
+
+def to_string(self):
```
jessex: Stylistic nit: the first thing people will want to see in this file is the `IngestInfo` class and friends, so I'd move this to the end/bottom.

recidiviz/ingest/models/ingest_info.py
terinpw: Yup updated.

terinpw:
> Do we [also want to](https://stackoverflow.com/questions/1436703/difference-between-str-and-repr) implement `__repr__` while we're at it?

Good point, did not know about `__repr__`. I just made `__repr__` call `to_string()`, since `__str__` defaults to `__repr__` if not implemented. Does that make sense to you or is there something more informative you'd want in `__repr__`?

terinpw:
chatted with @zacharylawrence and lint fail is known. Squashing and merging.

PR #194, opened 2018-11-15
convert unstructured data to table cells

### LINKED ISSUE #181: Data extractor: search for keys outside table cells
the DataExtractor class should manually search for keys in other elements besides table cells. Common patterns include:
- [x] `<span>Key: Value</span>`
- [x] `<label>Key</label><span>Value</span>`
- [x] `<span><label>Key</label>Value</span>`

and there are probably many others.

jamwalla:
Takes unstructured data in the forms
`<foo><bar>key</bar>value</foo>`, `<foo>key</foo><bar>value</bar>` `<foo>key<bar>value</bar></foo>`
and converts the elements into the form `<td>key</td><td>value</td>`

close #181


recidiviz/ingest/scraper_utils.py
```diff
@@ -25,13 +25,67 @@
...
+from lxml.html import HtmlElement
+
 import dateutil.parser as parser
 
 from google.appengine.ext import ndb
 from recidiviz.models import env_vars
 from recidiviz.utils import environment
 
 
+def convert_key_to_cells(content, key):
```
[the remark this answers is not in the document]
jessex: Yeah since it's dealing in concepts I would keep this in a class in the `recidiviz.ingest.extractor` package. If we've gotten to the point that there are several little utilities for understanding/parsing the html on page, I'd consider moving these out into their own utils class inside of the package, and leaving the bulk of `data_extractor` dedicated to the core algorithm.

recidiviz/ingest/scraper_utils.py
```diff
@@ -25,13 +25,80 @@
...
+    """Searches for elements in |content| that match a |key| and converts those
+    elements, along with their adjacent text, as table cells.
+
+    Args:
+        content: (HtmlElement) to be modified
+        key: (string) to search for
+    """
+    matches = content.xpath('.//*[starts-with(normalize-space(text()),"%s")]'
```
jessex: Do we not need to be returning `matches` from this function?

jessex: Oh, I see. I didn't expect results returned by an `xpath` call to be mutable/stateful inside of the page tree. I would assume they returned deep copies you could modify without altering the page. I stand corrected.

jamwalla:
ugh, sorry!! I tried to rebase some new changes under this and ended up pushing it all to master. I just added some comments so it's not a big deal, but sorry for polluting the commit history 😞.

Can we make master a protected branch?

jessex:
Ah, shoot. It's not bad enough that we should consider force-pushing to rewrite history on master (as it would have had it been, say, 20 commits), but yes let's definitely make master a protected branch in the near future.

PR #195, opened 2018-11-15
Vendor scraper for brooks jeffery

terinpw:
- Added vendor scraper for brooks jeffrey using base scraper (and data extractor)
- Added tests for brooks jeffrey scraper
- Added us_co_stone and us_ar_van_buren as children of brooks jeffrey
- fixed bug in data extractor to handle seeing a known key while looping through multi key mappings


recidiviz/ingest/vendors/brooks_jeffrey/brooks_jeffrey_scraper.py
```diff
@@ -0,0 +1,69 @@
...
+from recidiviz.ingest import constants
+from recidiviz.ingest.base_scraper import BaseScraper
+from recidiviz.ingest.extractor.data_extractor import DataExtractor
+
+
+class BrooksJeffreyScraper(BaseScraper):
+    """Scraper for counties using Brooks Jeffrey."""
+
+    def __init__(self, region_name, mapping_filepath=None):
+        if not mapping_filepath:
```
arian487: Neither of the child scrapers use this, I take it this is here just in case a few future brooks jeffrey child scraper needs its own yaml?

recidiviz/ingest/vendors/brooks_jeffrey/brooks_jeffrey_scraper.py
```diff
@@ -0,0 +1,69 @@
...
+        super(BrooksJeffreyScraper, self).__init__(region_name)
+
+    def set_initial_vars(self, content, params):
+        pass
+
+    def get_more_tasks(self, content, params):
+        task_type = params.get('task_type', self.get_initial_task_type())
+        params_list = []
+        if self.is_initial_task(task_type):
+            return BrooksJeffreyScraper._get_person_params(content)
```
arian487: I don't like this pattern of calling a private static method on the class.  It looks weird, and in general I prefer to have static methods when they'd actually be used.  In this case no one will ever call this staticly.  Just make it a private method.

recidiviz/ingest/vendors/brooks_jeffrey/brooks_jeffrey_scraper.py
```diff
@@ -0,0 +1,69 @@
+# Recidiviz - a platform for tracking granular recidivism metrics in real time
```
arian487: I love how short this scraper is!  From > 1000 lines to ~70 lines.  I think we've come a long way here :)

queue.yaml
```diff
@@ -14,19 +25,19 @@ queue:
     max_backoff_seconds: 300
     task_age_limit: 3d
 
-# us_ny - New York (state)
-- name: us-ny-scraper
+# us_mo_stone - Stone County, Missouri
+- name: us-mo-stonescraper
```
jessex: Missing the - before scraper here.

recidiviz/ingest/extractor/data_extractor.py
```diff
@@ -224,7 +225,7 @@ def _find_parent_ingest_info(self, ingest_info, hierarchy, val_index):
         return parent
 
     def _below(self, cell):
-        """Yields the cells below the given cell.
+        """Yields all cells below the given cell and breaks if it finds a key.
```
jessex: It may just be early, but this says we break if we find a key, but `_is_viable` returns True if it finds text which "could be the value for a field. The text should be non-empty and the cell should not contain a key in any of its descendants."

recidiviz/ingest/vendors/brooks_jeffrey/brooks_jeffrey.yaml
```diff
@@ -0,0 +1,11 @@
+key_mappings:
+  "Booking #": booking.booking_id
```
jessex: Does this need quotes because of the `#`?

recidiviz/tests/ingest/vendors/brooks_jeffrey/fixtures/person_page.html
```diff
@@ -0,0 +1,480 @@
...
+
+                        <!-- CMS Content Container -->
+                        <div id="intContentContainer" class="cell">
+
+                            <div class="text2" >
+
+                                		<span class="ptitles">First Middle Last</span>
+		<table width="100%" border="0">
+		<tr>
+				<td width="202" valign="top"><img width="200" src="templates/vbcso.com/images/inmates/201604655F.Jpg"/></td>
```
jessex: Is this a mugshot?

recidiviz/tests/ingest/vendors/brooks_jeffrey/fixtures/home_page.html
```diff
@@ -0,0 +1,556 @@
+
```
jessex: Quick sanity check: any chance of PII in either of these HTML files?

recidiviz/ingest/vendors/brooks_jeffrey/brooks_jeffrey_scraper.py
terinpw: Yep that was the thought.

recidiviz/ingest/vendors/brooks_jeffrey/brooks_jeffrey.yaml
arian487: Yea # is a special thing in yaml, has to be around quotes.

recidiviz/ingest/vendors/brooks_jeffrey/brooks_jeffrey_scraper.py
terinpw: Did it originally bc pycharm was annoyed at me, but undone here and in data_extractor

recidiviz/tests/ingest/vendors/brooks_jeffrey/fixtures/person_page.html
terinpw: Good catch. Had removed all text PII but missed this link.

recidiviz/tests/ingest/vendors/brooks_jeffrey/fixtures/home_page.html
terinpw: I've removed everything i could find about people in jail and any id numbers.

queue.yaml
terinpw: Updated

recidiviz/ingest/extractor/data_extractor.py
terinpw: @jamwalla to also take a glance here as well since you know the code well. 

Looks like `_is_viable` is always called after `_below`. So `_below` will fetch the value cell(s) below a given key, and then `_is_viable` is run on the value cell(s). Does this clarify the comment, or am i missing the inconsistency?

jamwalla: I think we should break on `_element_contains_key_descendant`.

The question of whether a cell  has no text because we haven't found the right cell yet or because the field is actually empty is a tough one, but we don't need to worry about it here, since we know we're moving directly down in a table that could have empty cells. We should make sure we have a test case where a cell that's not the bottom one in a column is empty.

recidiviz/tests/ingest/vendors/brooks_jeffrey/fixtures/home_page.html
jessex: Thanks for that!

recidiviz/ingest/extractor/data_extractor.py
jamwalla: Sorry, there's already a test like this coming in #194 (test_labeled_fields). So just the change to _element_contains_key_descendant!

recidiviz/tests/ingest/vendors/brooks_jeffrey/brooks_jeffrey_scraper_test.py
```diff
@@ -0,0 +1,91 @@
...
+        booking.admission_date = "1-1-2048- 12:30 am"
+        booking.total_bond_amount = "$695.00"
+
+        charge = booking.create_charge()
+        charge.name = "Charge 1"
+
+        arrest = booking.create_arrest()
+        arrest.agency = "Agency"
+
+        assert expected_info == result
```
terinpw: @jamwalla 

This test is now failing (hadn't run since Monday since i was out yesterday), and I think it's because of PR #194. I printed out expected_info and result below:

expected_info:
``` 
IngestInfo:
   person[0]: _Person:
      surname: Last
      given_names: First Middle
      age: 100
      sex: M
      race: W
      booking[0]: _Booking:
         admission_date: 1-1-2048- 12:30 am
         total_bond_amount: $695.00
         booking_id: 123
         charge[0]: _Charge:
            name: Charge 1
         arrest: _Arrest:
            agency: Agency
```

result:
```IngestInfo:
   person[0]: _Person:
      surname: Last
      given_names: First Middle
      age: 100
      sex: M
      race: W
      booking[0]: _Booking:
         booking_id: 123
         arrest: _Arrest:
            agency: Note: The charges and bail amounts may change after court appearances and may not be current. Bond companies and persons wishing to post bail should contact the Detention Center staff at 501-745-2112 for correct bail amount, charges and case numbers.
   person[1]: _Person:
      age: ncy
      booking[0]: _Booking:
         admission_date: 1-1-2048- 12:30 am
         total_bond_amount: $695.00
         charge[0]: _Charge:
            name: Charge 1
```

I think somehow instead of making arrest.agency = "Agency", it see's "Age" in "Agency" and makes a new person with age = "ncy", which then messes with the rest of the parsing.

terinpw:
Also as a side note: when I attempted to run this scraper locally, several rounds of errors arose (all that I found were unrelated to the child scraper itself). I think it's too much to put in this one PR, but I can start that work in a follow up. 

I think for now once the child tests are passing it'll make sense to submit this and do the base scraper fixes afterwords, if there aren't any objections.

jessex:
I agree, @terinpw. Once we have the build passing here, we should be ready to merge.

recidiviz/tests/ingest/vendors/brooks_jeffrey/brooks_jeffrey_scraper_test.py
terinpw: Updated data_extractor w the fix.

PR #197, opened 2018-11-16
 Migrate to sqlalchemy ORM & implement basic entity matching

zacharylawrence:
This PR involves migrating to the sqlalchemy ORM and using it to perform basic entity matching.

Followed the guiding principles outlined by sqlalchemy for handling sessions/transactions: https://docs.sqlalchemy.org/en/latest/orm/session_basics.html#when-do-i-construct-a-session-when-do-i-commit-it-and-when-do-i-close-it


recidiviz/persistence/database/schema.py
```diff
@@ -0,0 +1,136 @@
...
+
+    person_id = Column('person_id', Integer, primary_key=True)
+
+    scraped_person_id = Column('scraped_person_id', Integer)
+    surname = Column('surname', String)
+    given_names = Column('given_names', String)
+    birthdate = Column('birthdate', DateTime)
+    birthdate_inferred_from_age = Column('birthdate_inferred_from_age',
+                                         Boolean)
+    race = Column('race', String)
```
arian487: For everything that is a string but is actually an enum, is there a way to denote that at all?  At the very least add a comment.

recidiviz/persistence/database/schema.py
```diff
@@ -0,0 +1,136 @@
...
+class Bond(Base):
+    __tablename__ = 'bond'
+
+    bond_id = Column('bond_id', Integer, primary_key=True)
+
+    charge_id = Column('charge_id', Integer,
+                       ForeignKey("charge.charge_id"))
+
+    amount = Column('amount', Integer)
+    type = Column('type', String)
```
arian487: Don't use type, its a special keyword in python already.

recidiviz/persistence/entity_matching.py
```diff
@@ -0,0 +1,41 @@
...
+                                           session=session)
+
+    if len(existing_people) > 1:
+        raise PersonMatchingError(
+            'Found multiple people who should have been matched previously')
+
+    return existing_people[0] if existing_people else None
+
+
+class PersonMatchingError(Exception):
```
arian487: Prefer to have module specific errors at the top

zacharylawrence: Done

recidiviz/persistence/database/schema.py
zacharylawrence: Done

recidiviz/persistence/database/schema.py
zacharylawrence: I agree but need to look into how this will be represented in MS SQL Server/sqlite3. I'll leave this as is for now and update it when we learn what our actual database schema will look like.

recidiviz/persistence/database/schema.py
```diff
@@ -0,0 +1,136 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+# ============================================================================
+"""Define the ORM schema objects that map directly the the database."""
```
jessex: `to the`

recidiviz/persistence/database/database.py
```diff
@@ -0,0 +1,40 @@
...
+from recidiviz import Session
+from recidiviz.persistence.database.schema import Person
+
+
+def read_people(surname=None, birthdate=None, session=None):
+    """
+    Read all people matching the optional surname and birthdate. If neither
+    the surname or birthdate are provided, then read all people.
+
+    :param surname: The surname to match against.
```
jessex: I think PyCharm and other IDEs do this, but our style is to use `Args:` and `Returns:` statements in our docstrings. I think all of the new docstrings in here need to be updated.

recidiviz/persistence/persistence.py
```diff
@@ -16,27 +16,41 @@
...
+            existing_person = entity_matching.get_entity_match(person, session)
+
+            if existing_person is None:
+                session.add(person)
+            else:
+                person.person_id = existing_person.person_id
+                session.add(session.merge(person))
+
+            session.commit()
+        except Exception as e:
```
jessex: :+1: I like this idea of letting exceptions bubble up from the database layer into the persistence layer to be handled here, making sure not to let them bubble further into the app layer.

jessex: I think you can do the following shorthand if you don't actually do anything with `e`:
```
except:
    session.rollback()
    raise
```

recidiviz/persistence/database/database.py
```diff
@@ -0,0 +1,40 @@
...
+    :param birthdate: The birthdate to match against.
+    :param session: The transaction to read from.
+    :return: List of people matching the surname and birthdate, if provided.
+    """
+    session = Session() if session is None else session
```
jessex: I wonder if it would be advantageous to enforce that new `session` instances are created in the persistence layer and passed into the database layer, instead of letting them be created directly in the database layer? That way, we can make use of a session across multiple distinct database calls, and we never end up with, say, accidental use of separate sessions when we wanted a single transaction. 

Also, it seems aligned with this advice from the docs you linked in the PR description: "As a general rule, the application should manage the lifecycle of the session externally to functions that deal with specific data. This is a fundamental separation of concerns which keeps data-specific operations agnostic of the context in which they access and manipulate that data."

zacharylawrence: Yeah, that makes sense to me. It certainly seems less error prone this way. Done.

recidiviz/persistence/database/schema.py
zacharylawrence: Done.

recidiviz/persistence/persistence.py
zacharylawrence: Done.

zacharylawrence: As written, this should rethrow/reraise the exception. Do you mean it would make more sense to:
1. Catch the exception here 
2. Maybe log it?
3. Rollback the transaction
4. Just return like normal? Or maybe return a boolean for if the write was successful?

I guess this question really boils down to: How do you imagine the persistence layer being used and if that code has different logic if a ingest_info write fails?

recidiviz/persistence/database/database.py
zacharylawrence: Done.

recidiviz/persistence/persistence.py
jessex: No, I like it as-is here. This was meant only as a compliment for your current approach.  I think the persistence layer is the layer that our ingest app will speak to to persist entities and retrieve persisted entities. I think the database layer should generally not catch its own exceptions but let them bubble up to the persistence layer to be handled here.

I think it's then debatable as to whether the persistence layer should catch exceptions and propagate them (as you have here), or catch exceptions and then return certain error signals. Given that exceptions while communicating with the database are true errors and should be expected to be transient and should be retried, I think propagating the exception is the right move, probably.

An issue you'd likely be interested in is #114: should we continue this return -1 up the stack to the handler where it's converted into an error response to provoke retry? or should we pursue a different approach?

zacharylawrence:
@jessex Thanks for the review! Merging now.

PR #198, opened 2018-11-16
Create scraper for us-pa-daupin

zacharylawrence:
Create a new scraper for us-pa-daupin which only contains names. This includes a sanitized test.

TESTED = Runs locally but fails when writing datastore due to downstream bugs
PAIR = @terinpw


jessex:
💯

PR #199, opened 2018-11-19
automatically extract values from css selectors

jamwalla:
Adds a field css_key_mappings in scraper .yaml files to set fields from css-selected elements


recidiviz/ingest/extractor/data_extractor.py
```diff
@@ -113,11 +117,22 @@ def extract_and_populate_data(self, content, ingest_info=None):
...
+        self._css_selected_cells = []
         if not ingest_info:
             ingest_info = IngestInfo()
         if not ingest_info.person:
             ingest_info.create_person().create_booking()
 
+        for css_key, key_mapping_val in self.css_keys.iteritems():
+            class_to_set, ingest_key = key_mapping_val.split('.')
+            key_elements = content.cssselect(css_key)
+            if key_elements == []:
```
jessex: Since we throw away anything beyond the first element returned, should we raise an error if `len(key_elements) > 1`? Or is that expected but we _know_ the results are redundant?

arian487: I'll make a drive by comment here real quick:  I think this current CSS key method is restrictive right?  It only works for one value and does not handle a case where the same css key appears many times on a page (many people on one page for example).  Expanding it to handle that logic would be a bit cumbersome and somewhat difficult.

My thought is to use the same 'transform to table logic' to handle this case as well.  I like that pattern as it leverages the same logic throughout to actually write data to the ingest_info object.  That also would make it trivial to handle the case where one css key appears throughout the page. 

Just run through the keys finding the elements and wrap the elements around td's like you do for unstructured data, the table key would be just the css key itself and you can update needed_keys in that process as well.  I believe in that case it'll just work in cases where multiple people appear on one page and the same css key is present throughout.

jamwalla:
closed in favor of #207

PR #200, opened 2018-11-19
BaseScraper: support mixed GET and POST requests.

ohinds-vera:
Only pass the initial request data when data is None and it's the initial task.


jessex:
Also I just noticed something you may want to consider: your commits from your machine are coming from `ohinds` but you're logged into and opening PRs from `ohinds-vera`. That may not matter much but it does mean that the credit for your work is being split between your two accounts.

PR #201, opened 2018-11-19
Fixes regions test missing new region.

jessex:


zacharylawrence:
Thanks for the fix! Sorry I missed this one.

PR #203, opened 2018-11-20
Create relationships between ORM classes

zacharylawrence:
Update the ORM objects to include relationships between each table.

This also includes adding a test that reads a person and all of their relationships. Consequently, this PR also includes a dummy _convert_ingest_info() to enable tests that verify that relationships are correctly matched. This logic will need to be completely updated with the data converter logic (ie. ingest_info -> database schema) which is currently being written.

PAIR = @terinpw


recidiviz/tests/persistence/persistence_test.py
```diff
@@ -67,10 +75,13 @@ def test_sameTwoPeople_persistsOne(self):
         assert len(result) == 1
         assert result[0].surname == SURNAME_1
 
+    # TODO: Consider not replace everything with the newly scraped data, instead
```
zacharylawrence: Does it make sense to actually just throw everything away and replace it with the newly scraped data? If we keep only the Primary Key, then the entities will be considered "matched" and we'll maintain a historical view of this data in MS SQL Server. Basically, this would mean that our database is always a representation of what we understand to be the most up-to-date data (ie. whatever's on the website right now).

recidiviz/persistence/persistence.py
```diff
@@ -49,9 +50,43 @@ def write(ingest_info):
...
+
+def _convert_charge(charges):
+    """Converts charge."""
+    return [Charge(name=charge.name,
+                   bond=_convert_bond(charge.bond),
+                   sentence=_convert_sentence(charge.sentence))
+            for charge in charges]
+
+
+# TODO: This needs to use the same object if 2 charges point to the same bond
```
jessex: Is not the same true of `Sentence` below?

recidiviz/tests/persistence/persistence_test.py
jessex: >Basically, this would mean that our database is always a representation of what we understand to be the most up-to-date data (ie. whatever's on the website right now).

Isn't this also true if you just set the fields that have changed? I don't know the full extent of how the temporal table function works yet, but I would imagine that as long as the PK is stable, then whether you set all fields (A,B,C,D) or only the ones that updated (A,C) then either way the database represents the most up-to-date data.

recidiviz/persistence/persistence.py
zacharylawrence: Yup, I'll go ahead and add a comment there too. All this code will end up being thrown away when @arian487 and @jamwalla finish the IngestInfo -> Schema data converter

recidiviz/tests/persistence/persistence_test.py
zacharylawrence: I see - I think I was being unclear. No matter what, if we're using a temporal table and match the PK then the historical view will always be saved. 

MS SQL Server implements this historical view with 2 separate tables (one for what your data actually looks like **right now** and one for the history of changes), which makes the question: How much historical data do we want to keep outside of the historical view (ie. in our "master" table)?

On one hand, it seems reasonable make our master table mirror directly with the most recently scraped data. If you every want historical data, then that necessitates going to the historical table (ie. something like `SELECT COUNT(DISTINCT People) BETWEEN (1 year ago, today)`).

On the other hand, it also seems reasonable to not want to "throw away" data on a new scrape (yes, it's stored historically but that means you'd need to perform a historical query to access it). For example, scraping a website that removed bond information would mean we delete that bond information from our "master" booking.

Important to note here that I think the Person table should be exempt to the above discussion and always persisted across scraping sessions. It's possible we'd want to do the same with bookings (ie. allow looking up against previous bookings without a historical query).

Do you have thoughts on this? Either way, it seems like this decision doesn't need to be made now and should be easy to change in the future.

A final note that should probably go in a separate thread: As we learn more about what our requirements for a database are, I think it's worth taking another look into seeing if the temporal tables of MS SQL Server actual help us. If the types of queries we want to make aren't aided by the temporal functionality of MS SQL Server then we should move away from it. I suspect we'll gain insight on this as the UX team continues interviews and determines the key questions we want to answer with this data. Next week, we plan to write out SQL queries based on a temporal database vs our own temporal schema and compare the flexibility and easy of writing.

jessex: First: love everything in that last paragraph. If you give me a reason to choose an operationally simpler database, I'll gladly do so :). And I'm glad to hear that you're doing direct query comparison after the holiday break!

>Either way, it seems like this decision doesn't need to be made now and should be easy to change in the future.

Also agreed.

But back to the main question, I think it's simpler to ensure that the "master" table always represents our latest reality, so if a scrape _removes_ data from the object graph like in your example, then it should be removed from its respective "master" table. The dominant use case, I imagine, will be "tell me what the state of the world is right now." That also extends towards exporting our data set into other storage, e.g. BigQuery, where we'd probably want to export the current state of the world without having to piece together historical information. I think historical queries will be the minority use case and will accompany more complex functionality like recidivism identification.

zacharylawrence: Yup, that all sounds reasonable to me!

recidiviz/persistence/database/schema.py
```diff
@@ -73,12 +82,18 @@ class Arrest(Base):
 
 
 class Charge(Base):
-    __tablename__ = 'charge'
+    """Represents a Change as defined by our SQL schema"""
```
terinpw: "Charge" instead of "Change"

PR #204, opened 2018-11-20
migrate us_co_mesa to base scraper

colincadams:


recidiviz/ingest/us_co_mesa/us_co_mesa_scraper.py
```diff
@@ -106,90 +94,54 @@ def _get_person_params(self, content):
...
+            if row[0].text_content().startswith('Name'):
+                # TODO: move to scraper_utils or just stop trying to parse name
+                name = row[1].text_content().split()
+                person.given_names = ' '.join(name[:-1])
+                person.surname = name[-1]
+            if row[0].text_content().startswith('DOB'):
+                person.birthdate = row[1].text_content()
             if row[0].text_content().startswith('Booking#'):
-                record.record_id = row[1].text_content()
+                # TODO: should this be booking id?
```
colincadams: @ohinds-vera Originally we went back and forth on whether or not this was actually the booking id or whether it was the person id. This PR switches it to person id as that aligns better with our current data model, but wanted to highlight this change to get your opinion.

recidiviz/ingest/us_co_mesa/us_co_mesa_scraper.py
```diff
@@ -106,90 +94,54 @@ def _get_person_params(self, content):
...
-
-    @staticmethod
-    def _scrape_record(content):
-        record = Record()
+    def populate_data(self, content, params, ingest_info):
+        person = ingest_info.create_person()
 
         table = content.cssselect('table')[0]
         for row in table:
+            if row[0].text_content().startswith('Name'):
```
jessex: Since this is only in here due to the prior lack of css selector support in the data extractor, would it make sense to wait until @jamwalla merges #199 and then delegate to that?

recidiviz/ingest/us_co_mesa/us_co_mesa_scraper.py
```diff
@@ -106,90 +94,54 @@ def _get_person_params(self, content):
...
-
-        return person
-
-    @staticmethod
-    def _scrape_record(content):
-        record = Record()
+    def populate_data(self, content, params, ingest_info):
+        person = ingest_info.create_person()
 
         table = content.cssselect('table')[0]
```
jessex: Since this is only in here due to the prior lack of css selector support in the data extractor, would it make sense to wait until @jamwalla merges #199 and then delegate to that?

arian487:
Hm any reason why the data extractor wouldn't work here. Looking at the site it seems like it should be supported and there is no need to walk the table like you do. If I'm wrong carry on!

colincadams:
I'm not sure I totally understand all the functionality of the data extractor. I attempted to use it for this, and it worked for the DOB and the "Booking#", but I couldn't get it to handle the "bond" cell with all of the charge information.

I can sync up with @jamwalla to see if he can help me get it to work. @arian487 you're supposed to be out eating the foods 🍱 🍶

jamwalla:
I was trying to see if there was an easy way to get this to work but I think the data extractor needs a little more work to handle this case. One issue is the backwards relationship of charges/bonds: since the bond amount is listed above ('contains') the charge, we try to call `get_recent_bond` and error because there's no charge yet (hard-coded into the `hierarchy_map`). The unstructured data section also doesn't yet support key-value pairs in the middle of the text field, like the CASH/SURETY key. This is a really good test case that me and maybe @arian487 should work on more! But the data extractor isn't ready today :/

recidiviz/ingest/us_co_mesa/us_co_mesa_scraper.py
jamwalla: I was thinking that the css selector API would take the description of an element and return the contents of that element as the value, so I'm not sure that would get us far enough for this table. Do you think that feature should have more flexibility?

jessex: It should have the flexibility to handle this case if it appears in more jurisdictions. But I think in the case where a single website has a truly unique layout that is not generalizable to other jurisdictions, then we should hold off on servicing that use case specifically.

recidiviz/ingest/us_co_mesa/us_co_mesa_scraper.py
ohinds-vera: My guess is that this is a booking id, not a person id, but we won't know this until we see someone come back to jail and see if this number changes. My intuition comes from browsing the site and noticing that everyone has a 2018 id (or 2017 every once in a while), and I would expect *some* people to have been in that jail before have been readmitted, which would mean earlier years in this ID. What about the current data model is more consistent with the interpretation of a person ID?

PR #205, opened 2018-11-21
create_scraper script

jamwalla:
Usage: python create_scraper.py region_name
Creates __init__.py, region_name_scraper.py, and region_name.yaml files in
recidiviz/ingest/region_name, and updates queue.yaml and region_manifest.yaml.


jessex:
Would it be possible to move `create_scraper.py` into the `template` package without making it less usable, i.e. you can still call it easily from the root of the directory? I think putting that closer to the ingest package and out of the absolute root makes for better organization and readability.

jessex:
I think it would make a lot of sense to put a readme file at `recidiviz/ingest/scraper_template/README.md`, documenting how to use the template to bootstrap a new scraper, and maybe even documenting nearly everything for new scraper development. What do you think, @jamwalla?

create_scraper.py
```diff
@@ -0,0 +1,121 @@
...
+        target.write(contents)
+
+def create_files(region_name, subs):
+    """Creates __init__.py, region_name_scraper.py, and region_name.yaml files
+    in recidiviz/ingest/region_name
+    """
+    ingest_dir = os.path.join(os.path.dirname(__file__), 'recidiviz/ingest/')
+    template_dir = os.path.join(ingest_dir, 'scraper_template')
+    target_dir = os.path.join(ingest_dir, region_name)
+    os.mkdir(target_dir)
```
jessex: Does this error out if the target_dir already exists? We want to make sure we have a graceful failure here.

create_scraper.py
```diff
@@ -0,0 +1,121 @@
...
+  mode: push
+  rate: 5/m
+  bucket_size: 1
+  max_concurrent_requests: 3
+  retry_parameters:
+    min_backoff_seconds: 5
+    max_backoff_seconds: 300
+    task_age_limit: 3d
+"""
+    with open(os.path.join(top_level_path, 'queue.yaml'), 'a') as queue_file:
```
jessex: Appending is far and away the simplest approach here, but is there any way to sort this alphabetically by region code without much effort? Same for the region manifest updates.

create_scraper.py
```diff
@@ -0,0 +1,121 @@
...
+        queue_file.write(Template(queue_text).safe_substitute(subs))
+
+    region_text = """
+$region:
```
jessex: Testing this out locally, it looks like the region manifest update is not indented far enough. It's outside of the inner `regions:` block, so it won't be picked up by calls in `regions.py`.

create_scraper.py
```diff
@@ -0,0 +1,121 @@
...
+    yaml_target = os.path.join(target_dir, region_name + '.yaml')
+    populate_file(yaml_template, yaml_target, subs)
+
+
+def append_to_config_files(subs):
+    """Updates queue.yaml and region_manifest.yaml with the new region.
+    """
+    top_level_path = os.path.dirname(__file__)
+    queue_text = """
+- name: $region_dashes-scraper
```
jessex: Can we add a comment before this line, matching the descriptive comments in queue.yaml today, e.g. `# us-pa-dauphin-scraper - Dauphin County, Pennsylvania`? We'd need an arg or two for the human-readable description of the county, e.g. `--county Dauphin --state Pennsylvania` to support that.

create_scraper.py
```diff
@@ -0,0 +1,121 @@
...
+from string import Template
+
+
+def populate_file(template_path, target_path, subs):
+    with open(template_path) as template:
+        contents = Template(template.read()).substitute(subs)
+
+    with open(target_path, 'w') as target:
+        target.write(contents)
+
```
zdg2102: Super nit: I think it's supposed to be 2 newlines here

create_scraper.py
```diff
@@ -0,0 +1,121 @@
...
+def create_files(region_name, subs):
+    """Creates __init__.py, region_name_scraper.py, and region_name.yaml files
+    in recidiviz/ingest/region_name
+    """
+    ingest_dir = os.path.join(os.path.dirname(__file__), 'recidiviz/ingest/')
```
zdg2102: Not really familiar with the syntax here, but it looks like this expects the user to be in the top-level recidiviz-data directory when they run the script? Might be worth including some kind of validation to that effect so they don't end up creating the files in some random directory

create_scraper.py
```diff
@@ -0,0 +1,121 @@
...
+"""
+    manifest_path = os.path.join(top_level_path, 'region_manifest.yaml')
+    with open(manifest_path, 'a') as region_file:
+        region_file.write(Template(region_text).safe_substitute(subs))
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser()
+    parser.add_argument('region_name')
+    optional_args = [
```
zdg2102: Can you update the file-level docstring to include the optional arguments as well?

recidiviz/ingest/scraper_template/__init__.py
```diff
@@ -0,0 +1,20 @@
+# Recidiviz - a platform for tracking granular recidivism metrics in real time
```
zdg2102: Weird nit, but should this be a .py? Since it's not actually valid python because of the interpolations, and we're only using it as a text template, it might make more sense as a .txt. Same for the region_scraper template

create_scraper.py
jamwalla: thanks!

create_scraper.py
jamwalla: done

create_scraper.py
jamwalla: added a check

create_scraper.py
jamwalla: done, good catch!

create_scraper.py
jamwalla: It would be great to use python's yaml read/write, but it throws away comments. Is there a better option here than manually parsing the file?

create_scraper.py
jamwalla: thanks!

create_scraper.py
jamwalla: yes, thanks!

recidiviz/ingest/scraper_template/__init__.py
jamwalla: changed all to a txt

jamwalla:
thanks for the comments, Josh and Zach! @jessex -
 - I would love to add the 'how to write a scraper' doc as an .md file in the template dir, I can get started on that now but it'll change once we figure out how to test w/o ingest
- to run this at the top level but have the python file in the templates dir, do we need some other bash script in the top level? I'm not sure what the best way to approach that is.

create_scraper.py
```diff
@@ -0,0 +1,137 @@
...
+    with open(target_path, 'w') as target:
+        target.write(contents)
+
+def create_files(region_name, subs):
+    """Creates __init__.py, region_name_scraper.py, and region_name.yaml files
+    in recidiviz/ingest/region_name
+    """
+    ingest_dir = os.path.join(os.path.dirname(__file__), 'recidiviz/ingest/')
+    if not os.path.exists(ingest_dir):
+        raise OSError('Couldn\'t find directory recidiviz/ingest. Run this' +
```
zdg2102: Nit: missing space, this will read "Run thisscript"

jamwalla: thanks!

jamwalla:
Update: found a `us` library that I think we should use - if we want the state to be a required argument (and I don't see why it shouldn't be) we can (usually) get the timezone from the state, etc.

create_scraper.py
jessex: Yeah I don't see a better idea at the moment. @arian487 has expressed the idea of moving the queue information, alongside the general region manifest information, out of large monolithic files into smaller files that are in individual scraper directories. I approve of that idea, and we have to do something with the queue information once we complete the Cloud Tasks migration with @rasmi's help. So given that, I'd just leave this as-is for now. We're going to be changing it soonish anyway.

recidiviz/ingest/scraper_template/README.md
```diff
@@ -0,0 +1,196 @@
...
+### Usage
+`python create_scraper.py <county> <state>`
+
+For example:
+`python create_scraper.py ny kings`
+
+###### Optional Arguments
+ - `agency`: the name of the agency
+ - `agency_type`: one of `jail`, `prison`, `unified`
+ - `county`: the name of the county
```
jessex: Is this still an optional since we have county and state as required args above?

recidiviz/ingest/scraper_template/README.md
```diff
@@ -0,0 +1,196 @@
...
+it.
+
+###### What is the structure of pages on the website?
+Most website rosters follow a couple of familiar formats. For examples, refer to
+these scrapers:
+ - Data about multiple people on a single page:
+   [TODO(210)](../../../../issues/210)
+ - Multiple results pages with links to individual people:
+   [BrooksJeffreyScraper](vendors/brooks_jeffrey/brooks_jeffrey_scraper.py)
+ - Data about an individual person spread across multiple pages:
```
jessex: The Vermont scraper that @ohinds wrote follows this model. That scraper has not been updated to fit the new hierarchy, though, so it might not be appropriate to link to yet.

recidiviz/ingest/scraper_template/README.md
jamwalla: right, thanks! this was out of date in the script docstring as well.

recidiviz/ingest/scraper_template/README.md
jamwalla: added to the TODO

recidiviz/ingest/scraper_template/README.md
```diff
@@ -0,0 +1,196 @@
+Creating a Scraper
+==================
+From the top level `recidiviz-data` directory, run the `create_scraper.py`
+script to create the relevant files for a new scraper.
+
+### Usage
+`python create_scraper.py <county> <state>`
+
+For example:
+`python create_scraper.py ny kings`
```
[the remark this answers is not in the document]
jamwalla: yep, thanks!

recidiviz/ingest/scraper_template/README.md
```diff
@@ -0,0 +1,196 @@
...
+###### Optional Arguments
+ - `agency`: the name of the agency
+ - `agency_type`: one of `jail`, `prison`, `unified`
+ - `county`: the name of the county
+ - `names_file`: a file with a names list for this scraper
+ - `timezone`: the timezone, e.g. `America/New York`
+ - `url`: the initial url of the roster
+
+For example:
+`python create_scraper.py lake indiana --timezone='America/Chicago'`
```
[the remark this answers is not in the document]
jamwalla: unexpected! it's a `str` where it expects a unicode, so I'm surprised that the abbreviations worked. wrapping the argument in a `unicode()` call fixes this.

jamwalla:
thanks, sorry for the lack of thorough testing! I also fixed an issue where spaces between parts of a county name weren't getting replaced with underscores.

I think the best we can do with `argparse` is to have multi-word county names enclosed in quotes, so I updated the README to reflect that - let me know if you have a better solution!

I'm leaving the script at the top level for now and appending to the manifest files, but I like the idea of splitting those large files up.

thanks!!

PR #207, opened 2018-11-21
convert css keys to table cells

jamwalla:
This should replace #199 but I had trouble updating the old branch - sorry. @arian487's suggestion to add css keys as table cells is really elegant, so here's that attempt!


recidiviz/ingest/extractor/data_extractor.py
```diff
@@ -243,6 +252,17 @@ def _convert_key_to_cells(self, content, key):
         for match in matches:
             self._key_element_to_cell(key, match)
 
+    def _css_key_to_cell(self, content, css_key):
+        print css_key
```
jessex: Remember to remove this prior to merging.

recidiviz/ingest/extractor/data_extractor.py
```diff
@@ -243,6 +252,17 @@ def _convert_key_to_cells(self, content, key):
...
+        matches = content.cssselect(css_key)
+
+        for match in matches:
+            print match
```
jessex: Same here.

jamwalla:
thanks!

Issue #196, opened 2018-11-16
validate IngestInfo objects


jamwalla:
see also #179 
 Before converting to DB, we should validate IngestInfo objects for things like:
- [ ] overlapping fields, e.g. `age` and `birthdate`, are not set
- [ ] values are sensible, e.g. `birthdate` is not in the future
- [ ] the object does not have other/undefined attributes set, e.g. catch `charge.status` instead of `charge.charge_status`

Issue #179, opened 2018-11-13
Write an IngestInfo -> DB converter


arian487:
This standalone class should be the only class that has visibility into the DB and know how to convert the IngestInfo object into whatever database we have.  It should also be in charge of doing conversions to the proper types/enums.

arian487:
Nope, closing this out.

Issue #206, opened 2018-11-21
Decide how to handle unstructured names


colincadams:
Right now ingest info takes two name fields, `given_names` and `surname`. Some websites simply have a name field that isn't broken down into parts. We should decide on either a uniform scheme for parsing this (something simple but consistent, or something more complex like `nameparser` that @arian487 mentioned) or decide not to parse it and store the raw string provided (re: https://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/).

This has some implications on entity matching and depends on the name storage discussion.

jessex:
That's a great article, @colincadams. Thanks for sharing.

arian487:
I actually think we should make it even easier on the person and just concatenate the given_names and surname we get from the ingest_info, throw it in the name parser, and then use whatever is spat out.  Having only a fullname field can be annoying if the site breaks it down (data extractor can't concatenate fields).  By letting them do whatever they want with the name fields, and letting us throw it into the nameparser it'll be consistent and shoud just cleanly work.

arian487:
Fixed in https://github.com/Recidiviz/recidiviz-data/pull/611

PR #212, opened 2018-11-27
recidiviz.persistence.converter

jamwalla:
#179 #196

recidiviz.persistence.converter provides conversion/validation from recidiviz.ingest.models.ingest_info objects to recidiviz.persistence.database.schema objects.

This change assumes that fields in IngestInfo objects should only be `str`s or `bool`s.


jamwalla:
@zacharylawrence - does this mean we should use pennies or rounded dollars in ints?

`SAWarning: Dialect sqlite+pysqlite does *not* support Decimal objects natively, and SQLAlchemy must convert from floating point - rounding errors and other issues may occur. Please consider storing Decimal numbers as strings or integers on this platform for lossless storage.`

jamwalla:
Moving some concerns from earlier from the PR description to a comment:
 - concurrent/consecutive sentences - is this something we calculate ourselves?
 - we have a `scraped_person_id`, `scraped_booking_id` and `scraped_bond_id`. Should all the objects have this field?
It seems like we can address both of these as follow-up changes.

jamwalla:
@jessex, @zacharylawrence and @ohinds-vera -
Zach, can you review the persistence code?
Oliver, can you do a sanity check on the schema as it is now?
+ all three of you for general code review if you have time!

There's plenty more validation we can do, esp. knowing more about the fields, but this is a first step for the actual conversion.

thanks!!

recidiviz/persistence/converter.py
```diff
@@ -0,0 +1,422 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+# ============================================================================
+"""Converts scraped IngestInfo data to the database schema format."""
+
+from datetime import datetime
+from recidiviz.ingest import scraper_utils
+from recidiviz.persistence.database import schema
+
+def normalize(s):
```
jessex: What if `s is None`? Seems like we'll receive that value at some point, and it would produce an `AttributeError` here. Generally in all of these methods, we should explicitly handle None values.

recidiviz/persistence/converter.py
```diff
@@ -0,0 +1,422 @@
...
+def parse_date_or_error(date_string):
+    if date_string == '' or date_string.isspace():
+        return None
+    parsed_date = scraper_utils.parse_date_string(date_string)
+    if parsed_date is None:
+        raise ValueError('cannot parse date: %s' % date_string)
+    return parsed_date
+
+
+def calculate_birthdate_from_age(age):
```
jessex: I think this function deserves a docstring. The behavior is non-obvious. It's worth explaining why we have a function for stating that a person's birthday is today - age.

recidiviz/persistence/converter.py
```diff
@@ -0,0 +1,422 @@
...
+
+def time_string_to_days(time_string):
+    if time_string == '' or time_string.isspace():
+        return 0
+    split_string = time_string.split()
+    if len(split_string) == 2:
+        if split_string[1].upper().startswith('DAY'):
+            return int(split_string[0])
+        if split_string[1].upper().startswith('YEAR'):
+            return int(int(split_string[0]) * 365.24)
```
jessex: `relativedelta` _should_ let you convert from a number of years to a number of days. If you give it an actual range of years instead of an amount, it should even account for leap years automatically.

recidiviz/persistence/converter.py
```diff
@@ -0,0 +1,422 @@
...
+    if parsed_date is None:
+        raise ValueError('cannot parse date: %s' % date_string)
+    return parsed_date
+
+
+def calculate_birthdate_from_age(age):
+    if age == '' or age.isspace():
+        return None
+    try:
+        return datetime.date.today() - datetime.timedelta(days=int(age)*365.24)
```
jessex: Why not use `years` instead of `days` with an imprecise, manual conversion? Python's date and time and delta libraries should be used to handle all manner of date and time conversions on our behalf.

recidiviz/persistence/converter.py
```diff
@@ -0,0 +1,422 @@
...
+def split_full_name(full_name):
+    if full_name == '' or full_name.isspace():
+        return None
+    if ',' in full_name:
+        names = full_name.split(',')
+        if len(names) == 2 and all(names):
+            return tuple(names)
+    names = full_name.split()
+    if len(names) >= 2:
+        return ' '.join(names[:-1]), names[-1]
```
jessex: Why do we want to join in this specific order?

recidiviz/persistence/converter.py
```diff
@@ -0,0 +1,422 @@
...
+    raise ValueError('cannot parse full name: %s' % full_name)
+
+
+def normalize_gender_string(gender):
+    gender = normalize(gender)
```
jessex: Feel free to think this is wildly off-base, but I wonder if we should change the args for this method and the one below to `gender_string` and `race_string`, so that we don't have lines of code that read as "normalize gender" and "normalize race?" That might come off as a bit offensive, like we are trying to normalize gender or racial "outliers," when we're really making efforts to be inclusive. `normalize(gender_string)` is more explicit about what it's trying to accomplish.

Just a thought!

recidiviz/persistence/converter.py
```diff
@@ -0,0 +1,422 @@
...
+        if 'MAN' or 'MALE' in gender:
+            return 'TRANS MALE'
+    elif gender.startswith('F'):
+        return 'CIS FEMALE'
+    elif gender.startswith('M'):
+        return 'CIS MALE'
+
+    return 'OTHER'
+
+def normalize_race_string(race):
```
jessex: I remember @ohinds-vera mentioning that there was a US Census-provided list of rules for how to map different race labels. Is this based on that list? If so, and in general where we are relying on some accepted community or industry standard, our documentation should point to that list.

recidiviz/persistence/converter.py
```diff
@@ -0,0 +1,422 @@
...
+        return 'Other'
+    if race == 'W' or 'WHITE' in race or 'CAUC' in race or 'EURO' in race:
+        return 'White'
+    if race == 'A' or 'ASIA' in race:
+        return 'Asian'
+    if race == 'B' or 'BLACK' in race or 'AFRICA' in race:
+        return 'Black'
+    if 'HAWAII' in race or 'PACIFIC' in race:
+        return 'Native Hawaiian/Pacific Islander'
+    if 'NATIVE' in race or 'INDIAN' in race:
```
jessex: I'm actually a bit surprised that `INDIAN` might not show up in systems where race and ethnicity are smashed together, to refer to someone of Indian descent in the US. Is that a case we think we should handle?

recidiviz/persistence/converter.py
```diff
@@ -0,0 +1,422 @@
...
+
+
+def verify_is_bool(possible_bool):
+    if not isinstance(possible_bool, bool):
+        raise ValueError('expected bool but got %s: %s' %
+                         (type(possible_bool), possible_bool))
+
+def convert_person(ingest_person):
+    """Converts an IngestInfo person into a Schema person.
+        Args:
```
jessex: Nit: this should be dedented one tab.

recidiviz/persistence/converter.py
```diff
@@ -0,0 +1,422 @@
...
+        person.place_of_residence = normalize(ingest_person.place_of_residence)
+
+    if ingest_person.booking is not None:
+        person.bookings = [convert_booking(b) for b in ingest_person.booking]
+
+    return person
+
+def convert_booking(ingest_booking):
+    """Converts an IngestInfo booking into a Schema booking.
+        Args:
```
jessex: Same dedentation here and in methods below.

recidiviz/persistence/persistence.py
```diff
@@ -31,7 +31,7 @@ def write(ingest_info):
          ingest_info: The IngestInfo containing each person
     """
     for ingest_info_person in ingest_info.person:
-        person = _convert_ingest_info(ingest_info_person)
```
jessex: Can we delete the method this is referring to now?

recidiviz/persistence/converter.py
jamwalla: sounds good! `normalize` is a function that I hope only gets called from the converter, with a check that it's not none before that, but it's good to be safe anyways.

recidiviz/persistence/converter.py
jamwalla: added a docstring

recidiviz/persistence/converter.py
```diff
@@ -0,0 +1,422 @@
...
+        return 0
+    try:
+        clean_string = ''.join(
+            dollar_string.replace('$', '').replace(',', '').split())
+        return int(float(clean_string))
+    except:
+        raise ValueError('cannot parse dollar value: %s' % dollar_string)
+
+
+def verify_is_bool(possible_bool):
```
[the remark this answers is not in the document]
jamwalla: done, I like that approach!

recidiviz/persistence/converter.py
```diff
@@ -0,0 +1,422 @@
...
+        booking.scraped_booking_id = normalize(ingest_booking.booking_id)
+
+    if ingest_booking.admission_date is not None:
+        booking.admission_date = \
+            parse_date_or_error(ingest_booking.admission_date)
+
+    if ingest_booking.release_date is not None:
+        booking.release_date = parse_date_or_error(ingest_booking.release_date)
+        booking.release_date_type = 'ACTUAL'
+    elif ingest_booking.projected_release_date is not None:
```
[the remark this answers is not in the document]
jamwalla: Agreed - @terinpw is working on some changes with that field so I think he'll make that change in his current PR.

recidiviz/persistence/converter.py
```diff
@@ -0,0 +1,422 @@
...
+
+    if ingest_booking.charge:
+        booking.charges = [convert_charge(c) for c in ingest_booking.charge]
+
+    if ingest_booking.total_bond_amount is not None:
+        bond_amount = parse_dollar_amount(ingest_booking.total_bond_amount)
+        total_bond = schema.Bond(amount=bond_amount)
+        if not booking.charges:
+            booking.charges = [schema.Charge(bond=total_bond)]
+        elif not any(charge.bond and charge.bond.amount
```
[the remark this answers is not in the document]
jamwalla: added a comment

recidiviz/persistence/persistence.py
jamwalla: yes!

recidiviz/persistence/converter.py
jamwalla: cool, that's way more flexible than `timedelta`!

recidiviz/persistence/converter.py
jamwalla: solved with `relativedelta`

recidiviz/persistence/converter.py
jamwalla: that's a great catch!

recidiviz/persistence/converter.py
jamwalla: ack - he also mentioned a standardized format for the gender field, so we can hold off until I can write that up.

recidiviz/persistence/converter.py
jamwalla: ack

recidiviz/persistence/converter.py
```diff
@@ -0,0 +1,422 @@
...
+
+    if ingest_charge.statute is not None:
+        charge.statute = normalize(ingest_charge.statute)
+        # TODO(215): charge.offense_code = code_to_BJS(ingest_charge.statute)
+
+    if ingest_charge.name is not None:
+        charge.name = normalize(ingest_charge.name)
+
+    if ingest_charge.attempted is not None:
+        verify_is_bool(ingest_charge.attempted)
```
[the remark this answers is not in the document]
jamwalla: Right! When we move to Python 3 we can enforce that the IngestInfo fields are bools, but for now I think this is the best we can do? Let me know if there's a better place to check types.

jamwalla:
thanks, @jessex ! I added a bunch of unittests, ptal.

recidiviz/persistence/converter.py
```diff
@@ -39,6 +39,9 @@ def parse_date_or_error(date_string):
 
 
 def calculate_birthdate_from_age(age):
+    """We estimate a person's birthdate from their age by subtracting their age
```
jessex: I'm sure this has been discussed but we are inferring an estimated birthdate in order to do entity matching by DOB for everyone, regardless of region, then assigning everyone who is scraped on the same date the same birthdate might make entity matching less effective. What do you think? Is that not the goal here?

recidiviz/persistence/converter.py
zacharylawrence: In that case, does it make sense to use `_normalize`?

recidiviz/tests/persistence/converter_test.py
```diff
@@ -0,0 +1,157 @@
...
+
+    def test_parsePersonFields(self):
+        ingest_person = _Person(full_name='LAST, FIRST',
+                                birthdate='12-31-1999',
+                                gender='F', race='W',
+                                ethnicity='HISPANIC',
+                                place_of_residence='NNN\n  STREET \t ZIP')
+
+        schema_person = converter.convert_person(ingest_person)
+        assert schema_person.given_names == 'FIRST'
```
zacharylawrence: This is sad for now, but I think in the future we'll move to a Person object with a working __eq__ method.

recidiviz/tests/persistence/converter_test.py
```diff
@@ -0,0 +1,157 @@
...
+        schema_charge = schema_booking.charges[0]
+        assert schema_charge.bond is not None
+        assert schema_charge.bond.amount == 100
+
+    def test_updateChargeWithTotalBondAmount(self):
+        ingest_booking = _Booking(total_bond_amount='$100.00')
+        ingest_booking.create_charge().create_bond(bond_id='ID')
+
+        schema_booking = converter.convert_booking(ingest_booking)
+        assert len(schema_booking.charges) == 1
```
zacharylawrence: Is it worth making one of these tests for when there are >1 bonds?

recidiviz/persistence/converter.py
```diff
@@ -0,0 +1,445 @@
...
+    """
+    if age == '' or age.isspace():
+        return None
+    try:
+        return datetime.date.today() - relativedelta(years=int(age))
+    except:
+        raise ValueError('cannot parse age: %s' % age)
+
+
+def time_string_to_days(time_string):
```
zacharylawrence: I suspect that we may be able to find supported existing packages to help us with some of these conversions. For example: https://dateparser.readthedocs.io/en/latest/. Maybe the flexibility from rolling our own logic here is important, but it seems like a lot of work may already be done for us.

recidiviz/persistence/converter.py
```diff
@@ -0,0 +1,445 @@
...
+        if split_string[1].upper().startswith('YEAR'):
+            future_date = datetime.date.today() + \
+                relativedelta(years=int(split_string[0]))
+            return (future_date - datetime.date.today()).days
+    try:
+        return int(time_string)
+    except:
+        raise ValueError('cannot parse time duration: %s' % time_string)
+
+def split_full_name(full_name):
```
zacharylawrence: Another example of a library that may be able to help us: https://nameparser.readthedocs.io/en/latest/

recidiviz/persistence/converter.py
```diff
@@ -72,7 +72,7 @@ def split_full_name(full_name):
             return tuple(names)
     names = full_name.split()
     if len(names) >= 2:
-        return ' '.join(names[:-1]), names[-1]
+        return names[-1], ' '.join(names[:-1])
```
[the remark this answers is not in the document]
jamwalla: done! re: #206, if names are only used for matching within a single county, we don't have to sweat how they're parsed, since they'll be consistent across scrapes.

recidiviz/persistence/converter.py
jamwalla: Would we ever entity match based on inferred birthdate? I think depending on the results of name storing/entity matching discussion, we could also separate out age and DOB fields (and not set DOB unless we actually scrape it) with the understanding that DOB will be purged eventually.

Does that address this? Or is it still worth aligning to the same date on birth year (jan 1, e.g.)?

recidiviz/tests/persistence/converter_test.py
jamwalla: yes! done

recidiviz/persistence/converter.py
jamwalla: yep

recidiviz/persistence/converter.py
jamwalla: dateparser is great, + i imagine we'll have to revisit this anyways as we see more ways that sentences are represented.

recidiviz/persistence/converter.py
jamwalla: I'm tempted to hold off on this one for now. We have #206 filed and the decision we make depends on the amount of entity matching we plan to do.

see also https://github.com/Recidiviz/recidiviz-data/pull/212#discussion_r237920972 on this PR

recidiviz/persistence/converter.py
jessex: Well I am not 100% privy to what the goals are for an inferred birthdate, is more what I'm getting at. Why do we want to infer a DOB, and given that, do we want to pick an arbitrary constant day, use the scrape day, or pick a random day?

jamwalla: I see! I talked to Oliver and Zach L about this just now - it seems like we should eventually move to store birth year separately, but in any case it makes more sense to infer the birthdate to a set date to keep open the possibility of entity matching. I changed this to set birthdays to Jan. 1.

jamwalla:
even when testing locally, let's run the data converter as a form of validation

jessex:
What's the current state here, @jamwalla? It's looking nice! But I want to make sure there's not more big change coming before I take a final pass.

jamwalla:
Yeah, sorry for the under-communication! I got a little tangled up in other
work. I discussed with Arian today and we decided to implement enum values
with mappings to be more restrictive (he has a global mapping in another
branch). If we want to separate that work into another PR I think this is
ready for another review. We can use it for basic validation as is since
Terin hooked up the persistence layer to the scrapers.

PR #213, opened 2018-11-27
Remove special retry logic from us_ny

colincadams:
Stumbled upon this when removing uses of memcache. Using this to start a discussion of how to go forward from here.

To me it doesn't make sense to have this special retry logic live in the scraper task itself. If we want special retry logic that ensures a "resume" occurs if the scrape fails partway through, then this should live at a higher abstraction. One potential solution to this is to replace cron job with a continually running "controller" job that handles starting and stopping scrapers. This could resume scrapes if a session fails partway through, and it could also handle kicking off the process to release people who weren't found after a scrape session has completed. This avoids messy communication from tasks running on the queue triggering higher level operations.

WDYT? For now I've just removed this retry logic from the child altogether. I assume this would break us_ny, as it won't resume scrape sessions. Do you have thoughts on the best way to handle this? Should the scrape controller see if the existing session finished and if not resume it instead?


recidiviz/ingest/us_ny/us_ny_scraper.py
```diff
@@ -1091,105 +1077,6 @@ def link_person(record_list):
...
-
-        fail_count = 0 if not fail_count else fail_count
-
-        if fail_count < 3:
-            logging.warning("Couldn't parse next page of results (attempt %d)."
-                            " Failing task to force retry.", fail_count)
-            fail_count += 1
-            memcache.set(key=self.fail_counter, value=fail_count, time=600)
-            return False
-        else:
```
jessex: I agree with removing the failure count and I 100% agree this should be managed at a higher level of abstraction. That's clear. But I think we can remove that for now without also removing this hacky end-of-alphabet check. Removing the former does not break the NY scraper, as far as I can tell, but removing the latter does. I think if we just drop the if/else here and check if we're at the end of the alphabet with failure, then that's a reasonable enough state to leave this scraper in until I get a chance to move it over to the new scraper hierarchy.

colincadams: Done. This means that parsing failures won't be retried automatically by the task queue, but a resume will always be scheduled if there is an open session and the last name is less than 'ZYT'.

PR #214, opened 2018-11-27
Cache env vars locally instead of in memcache

colincadams:


recidiviz/models/env_vars.py
```diff
@@ -48,40 +47,34 @@ class EnvironmentVariable(ndb.Model):
...
-    region_code = "all" if not region_code else region_code
-    memcache_name = region_code + "_" + var_name
-
-    value = memcache.get(memcache_name)
+    value = LOCAL_VARS.get(var_name)
 
     if not value:
-
         datastore_result = EnvironmentVariable.query(ndb.AND(
-            EnvironmentVariable.region == region_code,
```
jessex: Why did we drop the region-specificity in this change?

jessex: No, you're right. I don't have a great scenario off of the top of my head. This is a smart move.

PR #216, opened 2018-11-28
Update booking schema w fields necessary for inferring release.

terinpw:


recidiviz/persistence/database/schema.py
```diff
@@ -53,6 +53,7 @@ class Booking(Base):
 
     admission_date = Column('admission_date', DateTime)
     release_date = Column('release_date', DateTime)
+    release_date_type = Column('release_date_type', String)
```
terinpw: Will have values inferred, projected, actual -- updated in schema doc as well.

PR #217, opened 2018-11-28
Remove memcache usage from regions.py (and refactor some other methods)

colincadams:
Part of #176

The main change here is that we don't cache the scraper package and
scraper class name in memcache. Instead, we only parse the region manifest
into a dict once, keep it in memory indefinitely, and get the scraper
package and class name from there.

This has the side effect of making Region object construction much cheaper
as we don't have to parse the full yaml file each time. Because of this,
I moved all of the region specific methods to be Region class methods.
This required some changes throughout the code base, and I may have gotten
a little carried away, so if you prefer any (or all) of them as non-class
methods I am happy to revert them.

In the future where each region has it's own manifest, instead of one large
`region_manifest`, I imagine we will only store the parsed config for a
region once it is needed. If this grows too large, we could put a
replacement policy in place.


recidiviz/utils/regions.py
```diff
@@ -77,17 +73,44 @@ def __init__(self, region_code):
...
+        ingest_module = getattr(top_level, "ingest")
+        scraper_module = getattr(ingest_module, self.scraper_package)
+
+        return scraper_module
+
+    def get_scraper(self):
+        """Retrieve a scraper object for a particular region
+
+        Returns:
+            Region's fully resolved scraper class (e.g.,
```
jessex: This and the previous method are both documented as returning the same thing, though it's not true. I think this one is inaccurate?

PR #219, opened 2018-11-29
Persistence method to mark inferred release date on bookings

terinpw:
Adds implementation to mark bookings with an inferred release date. Will still need to follow up and trigger this method after a background scrape has completed.


recidiviz/persistence/database/database.py
```diff
@@ -37,3 +38,35 @@ def read_people(session, surname=None, birthdate=None):
...
+
+    Args:
+        session: The transaction to read from
+    Return:
+        List of all bookings
+    """
+    return session.query(Booking).all()
+
+
+def read_open_bookings_scraped_before_date(session, region, date):
```
jessex: Did we consider doing all of this in SQL? Instead of reading in all relevant bookings and then updating each one-by-one, we could do something like `update booking set release_date = :date and release_date_type = 'INFERRED' where region = :region and last_scraped_date < :date and release_date is None and release_date_type = 'PROJECTED'`

terinpw: That's a great point. I chatted with @ohinds and when we mark people as released, we actually also want to update any associated charges to have a 'RESOLVED_UNKNOWN_REASON' status. This proves a little more cumbersome in SQL, so I think it now makes sense to rely on the query + ORM relationships to fetch the rows and then have a separate update method. 

Otherwise, I agree updating in one swoop is cleaner.

recidiviz/persistence/database/schema.py
```diff
@@ -53,8 +53,9 @@ class Booking(Base):
 
     admission_date = Column('admission_date', DateTime)
     release_date = Column('release_date', DateTime)
-    release_date_type = Column('release_date_type', String)
+    release_date_inferred = Column('release_date_inferred', Boolean)
```
terinpw: Per conversations w @ohinds and @jamwalla, we need a separate field for projected_release_date. In this case, the release_date_type can just be a boolean as there are only 2 options (inferred or actual).

recidiviz/persistence/persistence.py
```diff
@@ -18,10 +18,53 @@
...
+    finally:
+        session.close()
+
+
+def _infer_release_date_for_bookings(bookings, date):
+    """Marks the provided bookings with an inferred release date equal to the
+    provided date. Also resolves any charges associated with the provided
+    bookings as 'RESOLVED_UNKNOWN_REASON'"""
+    for booking in bookings:
+        assert booking.release_date is None
```
zacharylawrence: I'm not sure if this holds true in Python, but I'd normally discourage using assert statements in prod code. Is raising an Assertion Error here really what we want to be doing? Does it make sense to log an error and skip this booking? Or set it anyway?

Alternatively if you think this condition is worth verifying, is it possible to write a test that verifies the data eventually used by this method is valid? Or would an `infer_release_on_open_bookings` test including empty and set release_date bookings already in the database cover this case?

Also, feel free to ignore this comment if you think an AssertionError actually makes the most sense here or disagree in general.

recidiviz/persistence/database/database.py
```diff
@@ -37,3 +37,34 @@ def read_people(session, surname=None, birthdate=None):
...
+    """
+    Reads all open bookings in the given region that have a last_scraped_date
+    set to a time earlier than the provided date.
+
+    Args:
+        session: The transaction to read from
+        region: The region to match against
+        date: The exclusive upper bound on last_scrape_date to match against
+        Returns:
```
zacharylawrence: Nit: Spacing

recidiviz/tests/persistence/database/database_test.py
```diff
@@ -0,0 +1,66 @@
...
+class TestDatabase(object):
+    """Test that the methods in database.py correctly read from the SQL
+    database """
+
+    def setup_method(self, _test_method):
+        fakes.use_in_memory_sqlite_database()
+
+    def test_readOpenBookingsBeforeDate(self):
+        # Arrange
+        region = 'region'
```
zacharylawrence: Are these constants?

recidiviz/tests/persistence/database/database_test.py
```diff
@@ -0,0 +1,66 @@
...
+        incorrect_region = 'other_region'
+        release_date = datetime.datetime(2018, 7, 20)
+        most_recent_scrape_date = datetime.datetime(2018, 6, 20)
+        date_in_past = most_recent_scrape_date - datetime.timedelta(days=1)
+
+        # Bookings that should be returned
```
zacharylawrence: Could you remove these comments but using more descriptive variable names & expanding the test method name?

recidiviz/tests/persistence/persistence_test.py
```diff
@@ -14,14 +14,15 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 # =============================================================================
-"""Tests for persistence."""
+"""Tests for persistence.py."""
```
zacharylawrence: I'm not sure the style convention we're going for here, but technically this also tests calls being made to the database layer right (ie. all of persistence)?

recidiviz/tests/persistence/persistence_test.py
```diff
@@ -159,3 +160,34 @@ def test_readPersonAndAllRelationships(self):
...
+        b1.release_date_inferred = True
+        b1.charges[0].status = 'RESOLVED_UNKNOWN_REASON'
+
+        # Act
+        persistence.infer_release_on_open_bookings(
+            region, most_recent_scrape_date)
+
+        # Assert
+        bookings = database.read_bookings(session)
+        assert set(bookings) == set([b1, b2])
```
zacharylawrence: `set([b1, b2])`

Does this need to be wrapped in a List?

recidiviz/tests/persistence/persistence_test.py
```diff
@@ -159,3 +160,34 @@ def test_readPersonAndAllRelationships(self):
...
+                     charges=[charge_1])
+        b2 = Booking(region=incorrect_region, last_scraped_date=date_in_past,
+                     charges=[charge_2])
+
+        session = Session()
+        session.add(b1)
+        session.add(b2)
+        session.commit()
+
+        # Update open booking with inferred scrape date
```
zacharylawrence: Consider extracting to a method to give a method name instead of a comment?

recidiviz/tests/persistence/persistence_test.py
```diff
@@ -159,3 +160,34 @@ def test_readPersonAndAllRelationships(self):
...
         assert sentence_1.is_life == IS_LIFE_1
         assert sentence_2.is_life == IS_LIFE_2
+
+    def test_infer_release_date_on_open_bookings(self):
+        # Arrange
+        region = 'region'
+        incorrect_region = 'other_region'
+        most_recent_scrape_date = datetime(2018, 6, 20)
+        date_in_past = most_recent_scrape_date - timedelta(days=1)
+        charge_1 = Charge(name='Charge1', status='Charged')
```
zacharylawrence: Why do we need charges on these bookings?

recidiviz/persistence/persistence.py
zdg2102: Is AssertionError an Exception? I.e. will it be caught by the except clause triggering the transaction rollback in the previous method?

Generally I'm also in favor of avoiding assertions in prod code, but in this case I think we should roll back the transaction if child entities are in an invalid state, so throwing an exception here seems reasonable if that's the effect it will have.

zacharylawrence: Yeah, that's a good point. Maybe it makes sense to create a new DatabaseError exception type and throw that?

zdg2102: Yeah, throwing some kind of semantic exception seems better to me than an assertion

terinpw: Agreed that we should rollback if this happens, especially because it would mean that our something about our DB query is no longer correct. AssertionError is an Exception so it should be caught by the except clause (https://docs.python.org/2/library/exceptions.html#exception-hierarchy)

I'm new to python as well, but from digging around it looks like assert is a) ok to use in prod code and b) might be what we want here. From https://wiki.python.org/moin/UsingAssertionsEffectively it seems like this would be a 'cant happen' situation, as we should only ever use this method after reading open bookings from our db. 

What do y'all think?

recidiviz/persistence/database/database.py
terinpw: Done.

recidiviz/tests/persistence/database/database_test.py
terinpw: Updated!

recidiviz/persistence/persistence.py
zdg2102: This is admittedly pretty nitty, but in my understanding at least, assertions mean "this is a fundamental invariant and if this is violated we have no option but to let execution die". Recovering from an assertion error seems unsemantic to me, which is why I'd be in favor of some kind of explicit exception

recidiviz/tests/persistence/database/database_test.py
terinpw: Updated the variable names to be more descriptive. I do think comments grouping them into bookings to return and not return makes some sense, unless we'd rather put that in the name as well (which might make the names a bit unwieldy imo). Take a look and let me know if you think it could be clearer!

recidiviz/tests/persistence/persistence_test.py
terinpw: part of releasing an open booking is updating the charge status to 'RESOLVED_UNKNOWN_REASON', so wanted to test that as well.

recidiviz/tests/persistence/persistence_test.py
terinpw: Nope, good point, done.

recidiviz/tests/persistence/persistence_test.py
terinpw: I think the difference is in these tests we are thoroughly testing persistence.py. We are using the database layer as well, but in these tests I assume those calls work, and am not validating any edge cases, etc. 

That being said, doesn't matter to me if we say 'persistence' or 'persistence.py' so happy to update if you think it's clearer

recidiviz/persistence/persistence.py
terinpw: Chatted offline, and added a persistence level error.

recidiviz/tests/persistence/persistence_test.py
jessex: I think if we wanted to be super duper thorough, we'd consider unit tests for the persistence layer with the calls to the database layer mocked, unit tests for the database layer, and integration tests that run from the persistence layer through the database layer. But I think that would be overkill here so I'm fine with one set of tests that runs through both persistence and database, as long as we're covering all of the cases we can come up with.

recidiviz/tests/persistence/database/database_test.py
zacharylawrence: Oh, I was actually thinking of putting them in ALL_CAPS at the top of the file but maybe that's not the right way to do that?

recidiviz/tests/persistence/database/database_test.py
zacharylawrence: Looks good to me! Thanks!

recidiviz/tests/persistence/persistence_test.py
zacharylawrence: sgtm

PR #226, opened 2018-11-29
connect base_scraper to persistence layer

terinpw:


jessex:
Now that we are officially no longer writing to Datastore from descendants of the `BaseScraper`, can we update the README to note how to verify data that is being ingested locally?

recidiviz/persistence/persistence.py
```diff
@@ -20,16 +20,23 @@
...
-    surname/birthday already exists, then update that person.
+    If in prod, persist each person in the ingest_info. If a person with the
+    given surname/birthday already exists, then update that person.
+
+    If not in prod, simply print the given ingest_infos for debugging
 
     Args:
          ingest_info: The IngestInfo containing each person
     """
+    if not environment.in_prod():
```
jessex: Would you mind making this toggle-able via an environment variable or flag of some sort? Some of us would be able to persist data locally without issue.

recidiviz/persistence/persistence.py
```diff
@@ -20,16 +20,23 @@
...
+        print ingest_info
```
jessex: I think we should change this to an actual logging call. Even for local usage, there are some [advantages](https://www.loggly.com/blog/4-reasons-a-python-logging-library-is-much-better-than-putting-print-statements-everywhere/).

recidiviz/utils/environment.py
```diff
@@ -46,6 +46,7 @@ def in_prod():
         False if not
     """
     current_environment = os.getenv('SERVER_SOFTWARE', '')
+    print "TERINPW: current_environ = " + str(current_environment)
```
jessex: Remember sure to drop this or change it to a log statement before merging!

recidiviz/persistence/persistence.py
terinpw: Yup that makes sense, added it to the environment variables.

recidiviz/persistence/persistence.py
terinpw: Good point, done.

recidiviz/utils/environment.py
terinpw: Yup, realized that was there and removed!

PR #228, opened 2018-11-29
error checks in data extractor

jamwalla:
handle some cases where a cell is None


recidiviz/tests/ingest/extractor/data_extractor_test.py
```diff
@@ -389,3 +389,24 @@ def test_text_label():
...
+
+    expected_info = IngestInfo()
+    person = expected_info.create_person()
+    person.race = 'WHITE'
+    person.sex = 'M'
+    person.create_booking()
+
+    html_file = '../testdata/data_extractor/html/th_rows.html'
+    html_file = os.path.join(os.path.dirname(__file__), html_file)
+    with open(html_file, 'r') as f:
```
zacharylawrence: Could you use the `html.fromstring` helper util?

recidiviz/tests/ingest/extractor/data_extractor_test.py
```diff
@@ -383,9 +370,27 @@ def test_text_label():
     bond2 = charge2.create_bond()
     bond2.amount = '$100'
 
-    html_file = '../testdata/data_extractor/html/text_label.html'
-    html_file = os.path.join(os.path.dirname(__file__), html_file)
-    with open(html_file, 'r') as f:
-        html_contents = html.fromstring(f.read())
+    html_contents = html.fromstring(
```
zacharylawrence: In a future change, consider making these constants.

PR #232, opened 2018-11-30
Remove status on the Person object

zacharylawrence:


terinpw:
This change looks fine, but looks like you'll have to update the data_extractor test as well.

zacharylawrence:
> This change looks fine, but looks like you'll have to update the data_extractor test as well.

Updated!

PR #233, opened 2018-11-30
DataExtractor: make a copy of content

### LINKED ISSUE #230: data extractor should make a copy of content
Data Extractor directly modifies content which will confound the person using it.

jamwalla:
close #230


jessex:
This should do it, though I think it might be worth adding a test case that shows that it's a deep copy and not a shallow one, i.e. let the extract call make changes to the underlying content and then show that the content that was passed into the function remains unchanged.

jamwalla:
thanks!!

PR #234, opened 2018-11-30
Add SQL script to create schema

zdg2102:


jessex:
Instead of dropping this right in the root, can we put this in a `db/` or `bin/` directory? And _perhaps_ place that inside of `recidiviz/`?

recidiviz/create_schema.sql
```diff
@@ -0,0 +1,283 @@
...
+  birthdate date,
+  birthdate_inferred_from_age boolean,
+  gender gender,
+  race race,
+  ethnicity ethnicity,
+  county_of_residence varchar(100)
+);
+
+CREATE TABLE PersonHistory (
+  person_id integer NOT NULL, -- not shared
```
jessex: I might have read in that ANSI 2011 standards doc you shared that in the historical tables, the PK is a triple of the original PK, the valid_from, and the valid_to. Am I mistaken or are we intentionally eschewing that idea? It would seem to be useful in that it a) provides an index for the dominant historical query pattern (get the historical records for this entity between these dates), and b) provides protection against absolute duplicates of PK and valid timestamps.

jessex:
This might be off-base, but I wonder if it would be best to structure this as the first migration of more migration patches to come? I presume that the way we'll handle schema changes is with migrations via Sqlalchemy, and structuring the initial creation of the schema as migration 1 has nice benefits like consistency and standards-setting and permanence. Otherwise, schema changes we want to make need to both produce a migration for Sqlalchemy and presumably alter this script in the same way. What do you think?

zdg2102:
Going to close this PR since we're using alembic instead.

PR #236, opened 2018-12-03
move sessions to new datastore api

colincadams:
Feel free to pick this part. I took some liberties with making Session a wrapper around Entity, let me know if you have a better idea for how to handle that.

I also don't love the mocking that we have to do without the testbed support. We could instead use the datastore emulator, but I couldn't find a good way to cleanly launch that from within the test.


jessex:
Yeah I'm not sure if Datastore Emulator can help here. I don't think there's any viable way to run it, for example, from Travis CI during a build. Did you give any thought to making a stub Datastore for sessions, which would hold objects in memory and implement calls to write, update, and get sessions and session records? It would be kind of like reimplementing a minimal version of the testbed for our own purposes, which might be even more trouble than mocking. I liked the testbed, unfortunately.

recidiviz/ingest/sessions.py
```diff
@@ -40,35 +53,46 @@ class ScrapeSession(ndb.Model):
...
-    scrape_type = ndb.StringProperty(choices=(constants.BACKGROUND_SCRAPE,
-                                              constants.SNAPSHOT_SCRAPE))
+    @classmethod
+    def from_entity(cls, entity):
+        return cls(entity)
 
+    @classmethod
+    def new(cls, key, start=None, end=None, docket_item=None, last_scraped=None,
+            region=None, scrape_type=None):
+        entity = datastore.Entity(key)
```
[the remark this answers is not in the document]
colincadams: Yepp, it's not available 🙁 it seems like there is hope that sometime in the future it will be supported, but likely not for a while. The [github issue](https://github.com/GoogleCloudPlatform/datastore-ndb-python/issues/272) is still open, but hasn't received much love recently.

colincadams:
I did think about writing a thin in memory datastore, it would be pretty straightforward for get and set. But then I started thinking about queries and those become more complicated, it seemed like a decent amount of effort to build support for all filter operators, ordering, etc. and I wasn't sure that it was worth it at this point.

However the tradeoff is that the way in which we call datastore from methods within `sessions` aren't really tested. There isn't *that* much logic here, but a bug could slip through in the way we craft the query or something like that.

jessex:
>However the tradeoff is that the way in which we call datastore from methods within sessions aren't really tested. There isn't that much logic here, but a bug could slip through in the way we craft the query or something like that.

Yeah, and it's the kind of thing that we won't trivially catch in manual end-to-end testing. You'd need to not just run scrapers, but run various scrapers in different successions. It's arguable that we'd want that level of testing anyway at a certain point, but I'd feel better about deferring that indefinitely if we were confident in our unit test coverage.

I looked through the second-gen runtime docs and the [google-cloud-python](https://github.com/googleapis/google-cloud-python/issues) and couldn't find any great leads on a drop-in solution. This has been my only painpoint with GCP so far, the relative lack of local testing support. Have you taken a look [here](https://github.com/googleapis/google-cloud-python/blob/master/datastore/tests/system/test_system.py#L55), @colincadams? I think they're making some sort of emulator use.

recidiviz/tests/ingest/sessions_test.py
```diff
@@ -157,3 +160,57 @@ def wire_sessions_to_query(mock_client, mock_query, session_list):
     query = mock_query.return_value
     client.query.return_value = query
     query.fetch.return_value = (session.to_entity() for session in session_list)
+
+def fix_dt(dt):
+    return dt.replace(tzinfo=pytz.UTC)
+
+@pytest.mark.requires_emulator
```
jessex: This looks really nice! I'll hold off on a final pass until the Travis errors have been sorted out. But this is really, really promising.

jessex: That sounds like a great approach to me. Not just because I'd generally like to have a clean separation between unit and integration tests, but because the emulator is kind of an odd duck dependency, as you point out. It's a different process we're using as a sort of sidecar with its own set of dependencies.

colincadams:
Alright I got the rest of the tests converted and implemented skipping these tests by default. PTAL @jessex!

recidiviz/tests/ingest/sessions_test.py
```diff
@@ -97,322 +125,414 @@ class TestAddDocketItemToCurrentSession(object):
...
+        # These environment variables can only be set while these tests are
+        # running as they also impact ndb/testbed behavior
+        cls.prior_environs = write_datastore_emulator_environs()
+        sessions.clear_ds()
 
-class TestGetOpenSessions(object):
-    """Tests for the get_open_sessions method in the module."""
+    @classmethod
+    def teardown_class(cls):
+        restore_environs(cls.prior_environs)
```
jessex: Two questions:
* Does teardown_class get called even if there are tests errors or failures?
* Does the environment variable shifting matter from invocation to invocation of `pytest`? That is, do these variables get recreated fresh with each new run anyway, or do we risk bleeding changes to the environment across an entire machine?

jessex:
Only the one comment from me, @colincadams! If we've got good/safe answers to both, then feel free to wrap this up.

colincadams:
I made a small change to how we setup/teardown the env variables. Let me know what you think!

recidiviz/tests/ingest/sessions_test.py
jessex: With the second one, I was getting at the lifecycle here and wondering if we, say, failed to invoke the teardown method would we be leaving behind altered environment variables on the local machine? Basically, are these changes persistent in some way such that changes to our `env.yaml` that get left behind would cause tests on a future run to fail, or even issues with the dev appserver itself?

PR #241, opened 2018-12-03
migrate from users to cloud iap

colincadams:


colincadams:
An interesting side effect that I learned from testing this is that hitting 'http://metadata/computeMetadata/v1/project/project-id' returns an error locally (as expected) but on travis it will return travis' project id because it runs on app engine 😬

PR #243, opened 2018-12-04
Use the return value from populate_data

zacharylawrence:


recidiviz/ingest/base_scraper.py
```diff
@@ -130,7 +130,7 @@ def _generic_scrape(self, params):
             # If we want to scrape data, we should either create an ingest_info
             # object or get the one that already exists.
             ingest_info = params.get('ingest_info', IngestInfo())
```
colincadams: I still think this line should go away, but that can be a separate PR 😃

zacharylawrence: I agree and also wanted to seperate it in a different PR

PR #247, opened 2018-12-04
Rename data to post_data

zacharylawrence:


recidiviz/ingest/base_scraper.py
```diff
@@ -107,28 +107,31 @@ def _generic_scrape(self, params):
...
             return -1
+
         # For an initial task we just make a call to get initial variables.
         # This may be a noop for some scrapers.
         if self.is_initial_task(task_type):
             self.set_initial_vars(content, params)
         if self.should_scrape_data(task_type):
-            # If we want to scrape data, we should either create an ingest_info
-            # object or get the one that already exists.
+            # If we want to scrape post_data, we should either create an
```
jessex: I think this one should have stayed as "data"?

zacharylawrence: Good catch, thanks!

Issue #215, opened 2018-11-27
define enum values for database fields


jamwalla:
The following fields need to have enums defined:
 - `person.race`
 - `person.ethnicity`
 - `booking.release_reason`
 - `booking.custody_status`
 - `booking.classification`
 - `charge.offense_code` (`charge.statute` -> BJS code)
 - `charge.degree`
 - `charge.charge_class`
 - `charge.status`
 - `charge.court_type`
 - `bond.type`
 - `bond.status`

In addition, as we write these enums we'll also want conversions from common ways of expressing these values (i.e. 'caucasian' -> 'white')

jamwalla:
@ohinds-vera - I started writing some of this in #212, but I wouldn't want to write too much without seeing the data on a lot of different pages. Any thoughts on how we should go forward here?

zdg2102:
Enums are now defined for both application and database layers.

PR #248, opened 2018-12-04
define enum values via recidiviz/common/global_map.py

jamwalla:
#215

We'll expect exact matches to map strings to schema enum values. As we do research, we should add mappings from common string values to the schema values we interpret them to represent.


recidiviz/common/global_map.py
```diff
@@ -0,0 +1,70 @@
...
+
+def convert_value_to_enum(enum_type, value):
+    """This function will take a value and convert it into a standard value
+    to be used in the database and frontned.
+
+    Args:
+        enum_type: The enum type we are trying to convert.
+        value: The value we are converting.
+    Returns:
+        The converted value (string) or raises an error if it can't find a
```
jessex: The converted value is an enum, not a string, right?

jamwalla:
thanks!! yep, that's the plan (#191).

PR #250, opened 2018-12-04
Remove ingest_info.py in favour of a proto

arian487:


recidiviz/ingest/extractor/data_extractor.py
```diff
@@ -196,31 +197,35 @@ def _set_or_create_object(
             parent = self._find_parent_ingest_info(
                 ingest_info, self.hierarchy_map[class_to_set], i)
 
-            get_recent_name = 'get_recent_' + class_to_set
-            object_to_set = getattr(parent, get_recent_name)()
+            #get_recent_name = 'get_recent_' + class_to_set
```
zacharylawrence: Can we remove this?

recidiviz/ingest/extractor/data_extractor.py
```diff
@@ -196,31 +197,35 @@ def _set_or_create_object(
...
 
-            create_name = 'create_' + class_to_set
-            create_func = getattr(parent, create_name)
             # If the object we are trying to operate on is None, or it has
             # already set the ingest_key then we know we need to create a
             # new one.
-            if (object_to_set is None or
-                    getattr(object_to_set, ingest_key) != None):
-                object_to_set = create_func()
+            #print object_to_set
```
zacharylawrence: We can remove this too?

recidiviz/ingest/extractor/data_extractor.py
```diff
@@ -196,31 +197,35 @@ def _set_or_create_object(
...
-            get_recent_name = 'get_recent_' + class_to_set
-            object_to_set = getattr(parent, get_recent_name)()
+            #get_recent_name = 'get_recent_' + class_to_set
+            object_to_set = getattr(parent, class_to_set)
+            is_list = (parent.DESCRIPTOR.fields_by_name[class_to_set].label ==
+                FieldDescriptor.LABEL_REPEATED)
+            if is_list:
+                if not is_multi_key and len(object_to_set) != 0:
+                    object_to_set = object_to_set[-1]
             # If we are in multikey, we need to pull out the correct index of
```
zacharylawrence: Should these comments be at the same indent level?

recidiviz/ingest/extractor/data_extractor.py
```diff
@@ -196,31 +197,35 @@ def _set_or_create_object(
...
-            create_name = 'create_' + class_to_set
-            create_func = getattr(parent, create_name)
             # If the object we are trying to operate on is None, or it has
             # already set the ingest_key then we know we need to create a
             # new one.
-            if (object_to_set is None or
-                    getattr(object_to_set, ingest_key) != None):
-                object_to_set = create_func()
+            #print object_to_set
+            if (not object_to_set or
```
zacharylawrence: This could fit on one line?

recidiviz/ingest/models/ingest_info.proto
```diff
@@ -0,0 +1,88 @@
+syntax = "proto2";
+
+package recidiviz.ingest.models;
+
+message IngestInfo {
+	repeated Person person = 1;
```
zacharylawrence: I remember you saying before but I don't remember why we can't call this people?

recidiviz/ingest/us_co_mesa/us_co_mesa_scraper.py
```diff
@@ -143,10 +143,13 @@ def populate_data(self, content, params, ingest_info):
...
                             charge_meta) > 1 else None
 
-                        booking.create_charge(
+                        charge = booking.charge.add(
                             name=charge_name.strip(),
                             statute=charge_statute,
                             charge_class=charge_class,
-                            bond=bond,
                         )
+                        charge.bond.bond_id = bond_id.strip()
```
zacharylawrence: Is this caused from trying to put the unicode string into the proto? I noticed that before, but wasn't sure if there was any way we could do this automatically.

.pylintrc
```diff
@@ -20,7 +20,7 @@ ignore=CVS,lib
 
 # Add files or directories matching the regex patterns to the blacklist. The
 # regex matches against base names, not paths.
-ignore-patterns=.*node_modules.*
+ignore-patterns=.*node_modules.*, *_pb2.py
```
zacharylawrence: Like we were talking earlier, this means that we'd need to generate & upload a new version of the pb2 file every time we make a change to the proto. I think ideally, we'd have travis/appengine generate those files for us. Maybe @jessex has more context on how that could be done?

If that seems like too much trouble then it probably seems worth tracking that future change in a bug.

recidiviz/ingest/base_scraper.py
```diff
@@ -52,7 +52,7 @@
 
 from recidiviz.ingest import constants
 from recidiviz.ingest import scraper_utils
-from recidiviz.ingest.models.ingest_info import IngestInfo
+from recidiviz.ingest.models import ingest_info_pb2
```
jessex: Generally, I'd like client code to not have pb2 references everywhere we are interfacing with the class, and instead read as though it's working with Plain Old Python Objects. If we consistently do `import ingest_info_pb2 as ingest_info` then we help reach that goal.

But given that this was previously importing the `IngestInfo` class directly, can we not just do the same here?
`from recidiviz.ingest.models.ingest_info_pb2 import IngestInfo`

.pylintrc
jessex: You all have more Protobuf experience than I do, but it seems to me like we'd not want to defer the creation of that file to some _remote_ process, since this is the file that we rely on locally to make anything work. What I think would be useful would be baking the execution of that generation command into the local build.

Right now, devs are in the habit of running pylint and pytest independently, but sometimes we forget, and running each command separately doesn't scale. Maybe it's time to have a single build command that incorporates everything we need to run, and generation of that pb2.py file could potentially be part of that?

recidiviz/ingest/extractor/data_extractor.py
```diff
@@ -196,31 +197,35 @@ def _set_or_create_object(
             parent = self._find_parent_ingest_info(
                 ingest_info, self.hierarchy_map[class_to_set], i)
 
-            get_recent_name = 'get_recent_' + class_to_set
-            object_to_set = getattr(parent, get_recent_name)()
+            #get_recent_name = 'get_recent_' + class_to_set
+            object_to_set = getattr(parent, class_to_set)
+            is_list = (parent.DESCRIPTOR.fields_by_name[class_to_set].label ==
```
jessex: This is a little hard to read. Can we shift anything that absolutely relies on Protobuf-specific functionality (like this) behind functions in some Protobuf utils class? I'd love if this read as something like `is_list = is_repeated(parent)`

recidiviz/ingest/extractor/data_extractor.py
```diff
@@ -196,31 +197,35 @@ def _set_or_create_object(
...
-            create_func = getattr(parent, create_name)
             # If the object we are trying to operate on is None, or it has
             # already set the ingest_key then we know we need to create a
             # new one.
-            if (object_to_set is None or
-                    getattr(object_to_set, ingest_key) != None):
-                object_to_set = create_func()
+            #print object_to_set
+            if (not object_to_set or
+                    object_to_set.HasField(ingest_key)):
```
jessex: Another benefit of putting these Protobuf-specific fields behind a lil' interface is that if we have to switch away from Protobuf in the future for any reason, or if Protobuf 3 breaks everything, then we won't need a long PR like this making surgical incisions everywhere. We would just change all of that logic in one place.

recidiviz/ingest/models/ingest_info.proto
```diff
@@ -0,0 +1,88 @@
+syntax = "proto2";
```
jessex: Can we make sure there's a license at the top of this file before we merge, assuming there's a way to place comments in `.proto`?

recidiviz/ingest/models/ingest_info_pb2.py
```diff
@@ -0,0 +1,687 @@
...
+)
+
+
+
+
+_INGESTINFO = _descriptor.Descriptor(
+  name='IngestInfo',
+  full_name='recidiviz.ingest.models.IngestInfo',
```
jessex: Was this auto-generated, too? If so does it imply some magical way we can still reference that package directly? I'm guessing no since you're not using it elsewhere.

recidiviz/ingest/us_co_mesa/us_co_mesa_scraper.py
```diff
@@ -143,10 +143,13 @@ def populate_data(self, content, params, ingest_info):
                         charge_class = charge_meta[1] if len(
                             charge_meta) > 1 else None
 
-                        booking.create_charge(
+                        charge = booking.charge.add(
```
jessex: This call doesn't fail if `booking` has no charges yet? What does `booking.charge` refer to if not an instance of a `charge`?

jessex: Apply this same question to all instances of `x.y.add()` in here.

recidiviz/tests/ingest/extractor/data_extractor_test.py
```diff
@@ -310,18 +316,18 @@ def test_bad_labels():
...
     person.person_id = '046573'
-    person.sex = 'Male'
-    booking = person.create_booking()
+    person.gender = 'Male'
+    booking = person.booking.add()
     booking.booking_id = '00119283'
     booking.admission_date = '07/19/2018 14:24'
     booking.facility = 'MCSO'
-    arrest = booking.create_arrest()
+    arrest = booking.arrest
```
jessex: Unless I'm missing something, this `booking` instance doesn't have any arrests yet. What's the difference here between, say, `booking.arrest` and `booking.arrest.add()`?

Oh is it that the latter wouldn't make sense since `booking.arrest` is not repeated? So with a Protobuf-generated Python class, `booking.arrest` refers to a single instance, whereas `booking.charge` refers to a repeated type, hence the call to `booking.charge.add()` to instantiate a new charge on the booking. But even if `booking` has no actual arrest yet, `booking.arrest` returns a non-None value?

Thanks for coaching me up on Protobuf!

recidiviz/ingest/models/ingest_info.py
```diff
@@ -1,262 +0,0 @@
...
-        self.post_release_supervision_length = post_release_supervision_length
-
-    def __eq__(self, other):
-        return self.__dict__ == other.__dict__
-
-    def __repr__(self):
-        return to_string(self)
-
-
-def to_string(obj):
```
jessex: Do the Protobuf-generated classes have pretty printing built in? How do they log?

recidiviz/ingest/models/ingest_info_pb2.py
```diff
@@ -0,0 +1,687 @@
+# Generated by the protocol buffer compiler.  DO NOT EDIT!
```
jessex: What sort of test conventions exist for you all as regards working with Protobuf files? Do you write any unit tests to ensure that the generated pb2.py file stays in sync with the .proto file, for example?

recidiviz/ingest/extractor/data_extractor.py
arian487: Done.

recidiviz/ingest/extractor/data_extractor.py
arian487: Done.

recidiviz/ingest/extractor/data_extractor.py
arian487: Done.

recidiviz/ingest/extractor/data_extractor.py
arian487: Done.

recidiviz/ingest/models/ingest_info.proto
arian487: Because it makes the data extractor logic a lot cleaner.  We'd otherwise have to maintain a mapping in the data extractor because the data extractor dynamically creates the objects.

recidiviz/ingest/us_co_mesa/us_co_mesa_scraper.py
arian487: I'm not sure what you mean exactly?

.pylintrc
arian487: I agree a process like that would be really nice to have.  I'll talk to oliver about creating and designing that he was looking for some engineering work.  

At google, a lot of this proto generation is abstracted away from us.  We don't ever see pb2 files  checked in.  Our build files define it as a dependency and whenever we run unit tests, or any package any code it is built and packaged as part of that single run.  

I'm not sure I necessarily agree with with pylint and pytest being a part of a singular build command, thought it can/should be part of a presubmit.  That being said, pb2 generation also needs to be a part of every pytest invocation locally.  I frequently will run `pytest path/to/file`.  If that test relies on a fresh pb2 then it should rebuild the proto.  What I really want is kind of what we have with Blaze at google (Bazel externally).  The unit tests define their dependencies and all of that is build/compiled at test run time.  

I'll open an issue into investigating a general local presubmit, for now I just did it this way for simplicity to get it in (others are relying on the proto).

recidiviz/ingest/base_scraper.py
arian487: We can, its general practice at google (at least in ads) to import the base proto only.  It makes no difference to me though, happy to make this change.

recidiviz/ingest/us_co_mesa/us_co_mesa_scraper.py
arian487: Nope, so the actual underlying proto logic is quite complex.  booking.charge does not refer to charge, rather it refers to an internal proto type which handles charge.  In this case charge is a repeated field which has .add implemented.  What that does is create a charge proto type, append it to the underlying list which it maintains, and then returns the object.  Similarly if you do: booking.charge[x], that doesn't refer to the actual charge object but again the internal proto type which has dict accessor overridden to handle this call.

recidiviz/tests/ingest/extractor/data_extractor_test.py
arian487: Yup, you're mostly correct.  All fields are just FieldDescriptors, not the real types we expect.  The logic happens under the hood to create the type if needed, and access indices (if its a repeated field).  

A good example is if you try:  len(booking.charge) you'll get 0, but if you do isinstance(booking.charge, list) you get false.  Len has been overridden in the complex internal type for a repeated field but NOT for non repeated fields.  len(booking.arrest) will return an error.  

So booking.arrest IS some weird none value interestingly, but for simplicity you don't need to create the object to set things on it.  booking.arrest is some internal proto type that has field accessors overridden so if you do booking.arrest.arrest_id it'll create the arrest AND set the idea in one fell swoop.  If you notice this helped me simplify the code a bit in data extractor.

recidiviz/ingest/models/ingest_info.py
arian487: Pretty printing is built in, and even better when you do x==y and its false, it'll print out the offending lines :)

Lots of benefits from moving to proto here.  In fact Seralization to json is also built in which is great!

recidiviz/ingest/us_co_mesa/us_co_mesa_scraper.py
colincadams: This changes the relationship from [many charges to one bond] to [one charge to one bond]. This means the bond object will get duplicated for every charge, which is something we wanted to avoid.

One way we talked about handling this was by inverting the relationship, so that charges live inside bonds but it's much less intuitive in the general case. Other ideas for how to handle this?

recidiviz/ingest/models/ingest_info_pb2.py
arian487: At google we never write tests that analyze the generated proto.  This is because a pb2 file is never checked in and only ever created during a test run, or a package deploy, meaning it has to be generated every single time (so drift issues cannot possibly occur).  Each unit test file defines its depencies, and if you rely on importing a pb2 file, you need to specify it as a dependency (and under the hood it'll generate the pb2 file but for that test run only).  If you forget to specify it, your code will fail when it tries to import the file.  

Its generally very idiot proof.  You should read up on bazel, I think its a worthy thing to maybe try to move towards as it'll make our testing + presubmit story a lot nicer.

.pylintrc
jessex: >Blaze at google (Bazel externally)

2 cute 4 me

recidiviz/ingest/base_scraper.py
jessex: Well I'm just some person potentially spouting nonsense. Why is it practice to only import the base proto? I doubt I'm smarter than all of Google Ads eng somehow.

recidiviz/ingest/models/ingest_info_pb2.py
jessex: Got it, cool. If there's anything Proto-related that we merge now but want to change later, let's just make sure we've got the issues filed and the TODOs placed.

recidiviz/ingest/us_co_mesa/us_co_mesa_scraper.py
jessex: Awesome. Thanks a lot for the info!

recidiviz/ingest/us_co_mesa/us_co_mesa_scraper.py
arian487: Discussed offline.

PR #251, opened 2018-12-04
Update SQLAlchemy schema

zdg2102:
- Remove redundant column name strings
- Update column types to include enums and fixed string lengths
- Add additional relationship tables to avoid use of postgres-only ARRAY type

This is to prepare the declarative_base schema to be used to generate the initial alembic migration to create the DB tables


recidiviz/persistence/database/schema.py
```diff
@@ -14,141 +14,415 @@
...
+The below schema uses only generic SQLAlchemy types, and therefore should be
+portable between database implementations.
+
+NOTE: Many of the tables in the below schema are historical tables. Historical
+tables have no primary key. Their equivalent of the primary key from the
+corresponding master table is non-unique. A composite primary key could
+technically be generated from the ID and the period columns, but there would be
+no benefit from it, as that would require any foreign key relationship to
+directly use the period columns, which is contrary to the purpose of the
+temporal table design. Because of this non-uniqueness, any foreign key
```
zacharylawrence: Are there any foreign keys pointed to the historical table?

recidiviz/persistence/database/schema.py
```diff
@@ -14,141 +14,415 @@
...
+from sqlalchemy import Boolean, Column, Date, DateTime, Enum, ForeignKey, \
+    Integer, String
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.orm import relationship
 
+
+# Enum values. Exposed separately from the SQLAlchemy Enum that owns them for
+# easier access to the values.
+#
+# These values should not be used directly by application code, but should be
```
zacharylawrence: In that case, should we use a name like _GENDER or _RACE?

recidiviz/persistence/database/schema.py
```diff
@@ -14,141 +14,415 @@
...
-                                         Boolean)
-    race = Column('race', String)
-    ethnicity = Column('ethnicity', String)
-    place_of_residence = Column('place_of_residence', String)
-
-    bookings = relationship('Booking')
+    """Represents a person in the SQL schema"""
+    __tablename__ = 'Person'
+
+    person_id = Column(Integer, primary_key=True)
```
zacharylawrence: We still want a scraped_person_id for a JMS uuid

recidiviz/persistence/database/schema.py
```diff
@@ -14,141 +14,415 @@
...
+    surname = Column(String(255))
+    given_names = Column(String(255))
+    birthdate = Column(Date)
+    birthdate_inferred_from_age = Column(Boolean)
+    gender = Column(gender_enum)
+    race = Column(race_enum)
+    ethnicity = Column(ethnicity_enum)
+    place_of_residence = Column(String(255))
+
+    bookings = relationship('Booking', back_populates='person')
```
zacharylawrence: @terinpw and I were talking earlier about leaving off back_populate until we encounter a case where we'd actually need to do that. Thoughts?

recidiviz/persistence/database/schema.py
```diff
@@ -14,141 +14,415 @@
...
+
+
+class PersonHistory(Base):
+    """Represents the historical state of a person"""
+    __tablename__ = 'PersonHistory'
+
+    person_id = Column(Integer, nullable=False, index=True)
```
zacharylawrence: Is the Master table's PK all we need to index on?

recidiviz/persistence/database/schema.py
```diff
@@ -14,141 +14,415 @@
...
+    surname = Column(String(255))
+    given_names = Column(String(255))
+    birthdate = Column(Date)
+    birthdate_inferred_from_age = Column(Boolean)
+    gender = Column(gender_enum)
+    race = Column(race_enum)
+    ethnicity = Column(ethnicity_enum)
+    place_of_residence = Column(String(255))
+    valid_from = Column(DateTime, nullable=False)
```
zacharylawrence: Nit: Consider moving to the valid_to/from top of the class? I don't have strong feelings.

recidiviz/persistence/database/schema.py
```diff
@@ -14,141 +14,415 @@
...
-    race = Column('race', String)
-    ethnicity = Column('ethnicity', String)
-    place_of_residence = Column('place_of_residence', String)
-
-    bookings = relationship('Booking')
+    """Represents a person in the SQL schema"""
+    __tablename__ = 'Person'
+
+    person_id = Column(Integer, primary_key=True)
+    surname = Column(String(255))
```
zacharylawrence: Should we also index on county/scraped_person_id/name/dob to help improve entity matching?

zacharylawrence: Do we have to give lengths to the strings?

recidiviz/persistence/database/schema.py
```diff
@@ -14,141 +14,415 @@
...
 
+
+# Enum values. Exposed separately from the SQLAlchemy Enum that owns them for
+# easier access to the values.
+#
+# These values should not be used directly by application code, but should be
+# mapped to application layer enums in order to maintain decoupling between
+# the application and database.
+#
+# TODO: replace with Python enums after converting to Python 3
```
jessex: Can we add the issue number after this TODO, `TODO (176): replace with...`

recidiviz/persistence/database/schema.py
```diff
@@ -14,141 +14,415 @@
...
+        'Asian',
+        'Black',
+        'Native Hawaiian/Pacific Islander',
+        'Other',
+        'White')
+ethnicity = ('Hispanic',
+             'Not Hispanic')
+
+# Booking
+release_reason = ('Bond',
```
jessex: How do all of these relate to the enums @jamwalla just added in #248? If we can avoid having two separate enum types that need to have all of their values held in sync then that would be ideal.

recidiviz/persistence/database/schema.py
jessex: Default variable length/size can vary by provider and underlying engine so I think it's good practice to be explicit.

recidiviz/persistence/database/schema.py
jessex: I'm new to Sqlalchemy but in this case the `back_populates` means there will be a `Booking.person` accessor that lets you trace from a booking to a person, right? I could see waiting until needing to trace in that direction is helpful, but how does it work under the hood? Is it app-layer sugar that we can enable at any time with a code change, or does it instantiate/require some sort of database-layer relationship change?

recidiviz/persistence/database/schema.py
```diff
@@ -14,141 +14,415 @@
...
+    holds = relationship('Hold', back_populates='booking')
+    arrest = relationship('Arrest', uselist=False, back_populates='booking')
+    charges = relationship('Charge', back_populates='booking')
+
+
+class BookingHistory(Base):
+    """Represents the historical state of a booking"""
+    __tablename__ = 'BookingHistory'
+
+    # NOTE: BookingHistory does not contain last_scraped_date column
```
jessex: Why not? What's the harm?

recidiviz/persistence/database/schema.py
```diff
@@ -14,141 +14,415 @@
...
+    post_release_supervision_length_days = Column(Integer)
+    valid_from = Column(DateTime, nullable=False)
+    valid_to = Column(DateTime)
+
+
+class SentenceRelationship(Base):
+    """Represents the relationship between two sentences"""
+    __tablename__ = 'SentenceRelationship'
+
+    # NOTE: (A,B) is equal to (B,A). There should only be one
```
jessex: Do we have any safeguards in place to ensure we don't wind up with two equivalent pairs in opposite order?

I thought about ways to try to enforce this with keys alone but this late at night I have nothing. We could try to enforce this with the access patterns, e.g. `select distinct least(sentence_a_id, sentence_b_id), greatest(sentence_a_id, sentence_b_id) from sentence_relationship` would get you all distinct pairs regardless of order, so maybe something similar on write would let the app layer avert duplicates.

recidiviz/persistence/database/schema.py
zdg2102: Depends on what you mean. Since the master and historical tables are only linked by persistence of shared IDs, rather than actual foreign keys, the only foreign keys that would point to historical tables would be from other historical tables. And the historical tables have "foreign keys" to each other, in the sense that you can find a historical booking from a historical charge, but they're just int columns, they have no enforced foreign key constraints.

recidiviz/persistence/database/schema.py
zdg2102: I don't think so, because they technically need to be exposed for the mapper to the persistence layer enums. Although since privacy isn't enforced anyway, I'm fine with changing it if you think it would be more clear for a reader.

recidiviz/persistence/database/schema.py
zdg2102: @zacharylawrence so it's optional? All the SQLAlchemy docs I was following use it for every call to relationship(), so I just assumed it was required.

@jessex good question. I was planning on running this in offline mode to generate the output SQL before I actually run it against the DB, so we'll be able to check then and see exactly what it's doing.

recidiviz/persistence/database/schema.py
zdg2102: @zacharylawrence and I were discussing this yesterday and landed on the opposite approach. Keeping the enums separate allows us to keep the DB layer decoupled from the application layer (in the same way as having our own persistence models instead of using the ORM entities through everything), so we could, e.g., add a new enum value to the DB without exposing it yet to the application layer. If you think that's overkill though, I'm fine just using one set.

recidiviz/persistence/database/schema.py
zdg2102: None of the other columns really jumped out at me as needing an index. I looked into indexing the period columns, but saw a few places saying that indexing on date columns is tricky and can sometimes lead to worse performance than just a table scan, so I figured I'd leave those for now and we can look into that separately if needed.

recidiviz/persistence/database/schema.py
zdg2102: It lets us avoid making a ton of updates to the historical table without violating the historical table model. If someone is in the system for four months without anything related to their booking status changing, but we scrape them every day, we either have to add 100+ historical snapshots that differ only in last_scraped_date, or we have to alter the last_scraped_date of the top snapshot, which breaks the model that the historical tables should be append-only.

recidiviz/persistence/database/schema.py
zacharylawrence: Oh right, I guess I'd want something package-private but that doesn't exist. I think this lgtm.

recidiviz/persistence/database/schema.py
zacharylawrence: @jessex Right, back_populate allows you to travel the reverse direction (eg. Booking to Person). My understanding is that we can enable or disable this at any time since it just modifies the read SQL query being sent to the database, but it would be good to verify this.

@zdg2102 Right, you should be able to leave it off and everything will work fine.

recidiviz/persistence/database/schema.py
zacharylawrence: sgtm

recidiviz/persistence/database/schema.py
jessex: That's reasonable! Especially so if we only expect the enums to change very rarely. It would be nice to have a test case that asserts that the same set of values is present for each enum pair, though.

recidiviz/persistence/database/schema.py
jessex: Oh! Excellent point. Carry on!

recidiviz/persistence/database/schema.py
jessex: Mm, yeah, slightly annoying. I imagine this is an entity that we'll create with most every booking, too? Maybe leave this as-is until we know we're running into issues otherwise.

recidiviz/persistence/database/schema.py
zdg2102: Based on some changes we discussed this morning around how we expect data to be provided to the persistence layer, we now might end up dropping having two models and just do everything via the ORM entities. I'm not really sure yet how that's going to play out, so I'll leave this as-is for now until we've made a decision. I will sync with @jamwalla though about making sure the enums line up, so that if we leave them or switch, it won't cause any changes to the actual database.

zdg2102:
@zacharylawrence @jessex I added primary keys to the historical tables. They shouldn't be used and nothing should ever point to them, but it looks like SQLAlchemy requires every table to have a unique primary key.

recidiviz/persistence/database/schema.py
```diff
@@ -198,9 +204,17 @@ class BookingHistory(Base):
...
     __tablename__ = 'BookingHistory'
 
-    # NOTE: BookingHistory does not contain last_scraped_date column
+    # NOTE: BookingHistory does not contain last_scraped_date column. This is to
+    # avoid needing to create a new BookingHistory entity when a booking is
+    # re-scraped and no values have changed except last_scraped_date.
+
+    # This primary key should NOT be used. It only exists because SQLAlchemy
+    # requires every table to have a unique primary key.
+    booking_history_id = Column(Integer, primary_key=True)
```
jessex: Are this and the other history ids going to be an auto-increment by default?

zdg2102: primary_key=True gives autoincrement by default in SQLAlchemy

recidiviz/persistence/database/schema.py
zacharylawrence: SGTM. As you explained offline, we have to do this because foreign keys must point to a unique PK.

recidiviz/persistence/database/schema.py
zacharylawrence: As discussed offline, external_id sgtm!

recidiviz/persistence/database/schema.py
zacharylawrence: @zdg2102 I don't have strong feelings on this but I'd prefer to minimize clutter in the ORM object's namespace until we have a concrete use case for adding this (then we just update this one parameter).

zdg2102:
Updated the persistence tests to include fields I added non-null checks for in this PR

recidiviz/persistence/database/schema.py
zdg2102: Removing it caused a whole bunch of mapper errors. I'm going to leave it for now, and try to mess around with getting the links one-way only in a separate PR.

zacharylawrence: sgtm. Don't think it's worth worrying about if it's too much trouble

PR #253, opened 2018-12-05
Update region_manifest.yaml documentation

jamwalla:


region_manifest.yaml
```diff
@@ -15,6 +15,20 @@
...
+# Much of this file can be filled in automatically from the create_scraper.py
+# script. For each region, the following fields are required:
+#   agency_name: the official name of the facility
+#   agency_type: one of 'jail', 'prison', 'unified'
+#   base_url: the base url of the roster or search page for the facility
+#   names_file: a list of names the scraper should search for
+#     (use us_ny_names.csv as a default)
+#   queue: the scraper name in queue.yaml
+#   region_code: the region code, formatted as <country>_<state>_<county>
+#   region_name: the official name of the region
```
jamwalla: this field is not consistent currently - @jessex, do you know what the correct use is?

jamwalla:
@jessex - this was one piece of feedback from the 20%ers. One question was the procedure for vendor scrapers - is it any different for this file?

thanks!!

region_manifest.yaml
jessex: So I don't think this field is even needed any more, actually. I think it's outdated. If you look for usages in the code, there are places that refer to "region_name" but are actually using the "region_code" and there are no more places that actually reference `region.region_name`. I _think_ we can remove this now!

jessex:
@jamwalla: no update to the region manifest required for vendor scrapers. You only need to make sure your region scraper extends the vendor scraper, but otherwise it doesn't need to be listed anywhere. We should make that clear in the docs.

jamwalla:
thanks @jessex !

Filed #256 to remove `region_name`.

PR #254, opened 2018-12-05
Add an ingest_info_proto

arian487:
As per some discussions, we decided to keep the in memory object which is what the scraper reads into, this keeps relationships good.  The proto is changed to be a flatter list of data that relates to each table, and maintains relationships using IDs (either generated or provided).  The pervious model lost some relational data, and it would otherwise be difficult to allow for this relationship to exist on a direct ingest model.

The idea here is that we can then get table dumps from their data base significantly more easily, and is likely closer to what might exist in their own database.


arian487:
To add to this, the single point of entry to the persistence layer will still be this proto.  I'm going to add a convert to proto function in ingest_info.py that is invoked by the base scraper before a call to the persistence layer happensl;

recidiviz/ingest/models/ingest_info.proto
```diff
@@ -0,0 +1,113 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+// =============================================================================
+
+syntax = "proto2";
+
+package recidiviz.ingest.models;
```
zacharylawrence: Should this be updated to live outside the ingest package now?

recidiviz/ingest/models/ingest_info.proto
```diff
@@ -0,0 +1,113 @@
...
+
+message Booking {
+	optional string booking_id = 1;
+	optional string admission_date = 2;
+	optional string project_release_date = 3;
+	optional string release_date = 4;
+	optional string release_reason = 5;
+	optional string custody_status = 6;
+	optional string jurisdiction_status = 7;
+	optional string hold = 8;
```
zacharylawrence: The schema currently supports Bookings with multiple holds

recidiviz/ingest/models/ingest_info.proto
```diff
@@ -0,0 +1,113 @@
...
+	optional string ethnicity = 9;
+	optional string place_of_residence = 10;
+
+	repeated string booking_ids = 11;
+}
+
+message Booking {
+	optional string booking_id = 1;
+	optional string admission_date = 2;
+	optional string project_release_date = 3;
```
zacharylawrence: Typo: projected

recidiviz/ingest/models/ingest_info.proto
arian487: Looking at it, I'm not sure I agree.  I think ingest.models is a good place because it is a data model after all, and represents data that we would ingest.

recidiviz/ingest/models/ingest_info.proto
arian487: This will have to be updated in ingest_info too, will save for another PR

recidiviz/ingest/models/ingest_info.proto
arian487: Fixed

recidiviz/ingest/models/ingest_info.proto
jessex: The main question, I think, is whether this model will be used directly in calculation code? If not, is that because the calculation code won't live in this repository or because it will read data out of Cloud SQL into a different model? I think it's an okay answer to say we don't know yet so it should continue to live inside of ingest, but let's keep an eye on it.

recidiviz/ingest/models/ingest_info.proto
jessex: This comment highlights one thing I just want to echo again: let's not get in a position where we have 2 or 3 different version of `IngestInfo` which can easily drift apart from one another. Zach won't always be around to point out that two models don't match :)

recidiviz/ingest/models/ingest_info.proto
zacharylawrence: I don't think there is any intention for this model to be used in calculate code since this model exclusively represents data as we expect to receive it from various jurisdictions (eg. scraped data, direct ingest, etc.)

I think the tricky part about this is that the `ingest` directory should probably be renamed to something like `ingest.scraper` now that we plan to support multiple forms of ingesting data. Since this proto will be used to communicate between each of these ingest layers and the persistence layer (eg. over network calls) then I think it makes the most sense to live outside this scraper specific package and in a common directory between the Persistence Layer and all the Ingest Layers.

Issue #252, opened 2018-12-05
Investigate a presubmit script to run locally which runs pytest + pylint and also does pb2 generation


arian487:
Write now the local story around testing is difficult with the introduction of protos especially.  We can consider using bazel to do this,  but we want to get to a place where users running tests locally shouldn't know or care about complex dependencies like building protos.

PR #258, opened 2018-12-05
Add a converter to go from ingest_info to ingest_info_proto

arian487:
This adds a util to go from IngestInfo python to an ingest info proto.  It uses python memory address as the ID to maintain relationships (or the provided id).


arian487:
So it looks like the way other people have handled protos is by having a script that gets and installs protoc:

https://github.com/ga4gh/ga4gh-schemas/blob/master/tools/travis-install-protoc.sh

They then add this to the travis yaml:

https://github.com/ga4gh/ga4gh-schemas/blob/master/.travis.yml#L16

We can then write a wrapper around pytest which first finds all of the protos in the subtree, generates them, and then runs the tests.  Locally, users would use the small wrapper script to run tests instead of pytest but I don't necessarily love that idea either.  Can we bake this into pytest itself?  Thoughts?

recidiviz/ingest/ingest_info_utils.py
```diff
@@ -0,0 +1,86 @@
...
+    proto = IngestInfo()
+
+    person_map = {}
+    booking_map = {}
+    charge_map = {}
+    arrest_map = {}
+    bond_map = {}
+    sentence_map = {}
+
+    def _populate_proto(proto_name, ingest_info_source, id_name, proto_map):
```
jessex: I'd love to have a docstring on this even though it's internal, since it's nearly impossible for me to identify what these arguments mean on sight.

recidiviz/ingest/ingest_info_utils.py
```diff
@@ -0,0 +1,86 @@
...
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+# =============================================================================
+
+"""Convenience code for an ingest_info object."""
+
+from recidiviz.ingest.models.ingest_info_pb2 import IngestInfo
+
+
+def convert_ingest_info_to_proto(ingest_info):
+    """Converts an ingest_info python object to an ingest info proto.
```
jessex: Can we add a little bit more about the exact operation it's doing (dynamically pulling field names from the proto, looking for those field names on the IngestInfo, setting fields, caching them in memory in case they appear in multiple parts of the object graph, etc.)?

jessex:
Since there's no longer a direct reference to `ingest_info_pb2` anywhere (hence "No module named ingest_info_pb2" failures) are you suggesting the script to run protoc to generate `ingest_info_pb2` so that it's available? I haven't worked with Protobuf in any serious project before and you all have but I'm curious about how far the complexity of working with it will go versus something more vanilla.

recidiviz/ingest/ingest_info_utils.py
arian487: Added docstring plus more details clarifying logic.

recidiviz/ingest/ingest_info_utils.py
arian487: Added all of those comments in the internal object which handles much of that logic.

arian487:
I'm not sure what you mean, there is a direct reference to ingest_info_pb2 and it needs to exist.  I think likely to move this CL along, I'm going to just submit the pb2 file so travis can be happy and that'll unblock zach in connecting this bit to the persistence layer.  I added https://github.com/Recidiviz/recidiviz-data/issues/252 from the previuous PR.  

In general, the complexity should stop as soon as we figure out automatic generation which shouldn't be too hard.  I think proto has a lot of benefits and built in features.  The fact that it is language agnostic is a huge win for us and especially the direct ingest side.  We could for example decide to have a different language server which handles direct ingest side or for example some vendors might have written their code in something other than python.  Giving them a gold standard proto to work with makes it extremely easy for them to interface with us.

arian487:
I'm not sure what you mean, there is a direct reference to ingest_info_pb2 and it needs to exist.  I think likely to move this CL along, I'm going to just submit the pb2 file so travis can be happy and that'll unblock zach in connecting this bit to the persistence layer.  I added https://github.com/Recidiviz/recidiviz-data/issues/252 from the previuous PR.  

In general, the complexity should stop as soon as we figure out automatic generation which shouldn't be too hard.  I think proto has a lot of benefits and built in features.  The fact that it is language agnostic is a huge win for us and especially the direct ingest side.  We could for example decide to have a different language server which handles direct ingest side or for example some vendors might have written their code in something other than python.  Giving them a gold standard proto to work with makes it extremely easy for them to interface with us.

arian487:
I'm not sure what you mean, there is a direct reference to ingest_info_pb2 and it needs to exist.  I think likely to move this CL along, I'm going to just submit the pb2 file so travis can be happy and that'll unblock zach in connecting this bit to the persistence layer.  I added https://github.com/Recidiviz/recidiviz-data/issues/252 from the previuous PR.  

In general, the complexity should stop as soon as we figure out automatic generation which shouldn't be too hard.  I think proto has a lot of benefits and built in features.  The fact that it is language agnostic is a huge win for us and especially the direct ingest side.  We could for example decide to have a different language server which handles direct ingest side or for example some vendors might have written their code in something other than python.  Giving them a gold standard proto to work with makes it extremely easy for them to interface with us.

arian487:
I'm not sure what you mean, there is a direct reference to ingest_info_pb2 and it needs to exist.  I think likely to move this CL along, I'm going to just submit the pb2 file so travis can be happy and that'll unblock zach in connecting this bit to the persistence layer.  I added https://github.com/Recidiviz/recidiviz-data/issues/252 from the previuous PR.  

In general, the complexity should stop as soon as we figure out automatic generation which shouldn't be too hard.  I think proto has a lot of benefits and built in features.  The fact that it is language agnostic is a huge win for us and especially the direct ingest side.  We could for example decide to have a different language server which handles direct ingest side or for example some vendors might have written their code in something other than python.  Giving them a gold standard proto to work with makes it extremely easy for them to interface with us.

arian487:
I'm not sure what you mean, there is a direct reference to ingest_info_pb2 and it needs to exist.  I think likely to move this CL along, I'm going to just submit the pb2 file so travis can be happy and that'll unblock zach in connecting this bit to the persistence layer.  I added https://github.com/Recidiviz/recidiviz-data/issues/252 from the previuous PR.  

In general, the complexity should stop as soon as we figure out automatic generation which shouldn't be too hard.  I think proto has a lot of benefits and built in features.  The fact that it is language agnostic is a huge win for us and especially the direct ingest side.  We could for example decide to have a different language server which handles direct ingest side or for example some vendors might have written their code in something other than python.  Giving them a gold standard proto to work with makes it extremely easy for them to interface with us.

recidiviz/tests/ingest/ingest_info_utils_test.py
```diff
@@ -0,0 +1,136 @@
...
+
+from recidiviz.ingest import ingest_info_utils
+from recidiviz.ingest.models import ingest_info
+from recidiviz.ingest.models import ingest_info_pb2
+
+
+def test_id_is_generated():
+    info = ingest_info.IngestInfo()
+    person = info.create_person()
+    person.surname = 'testname'
```
zacharylawrence: Does it make sense to pull these strings/etc. into file constants to avoid the possibility of typos?

arian487: Eh, don't think its too necessary for unit tests that'll likely never change.

arian487: Yea I saw, in this case the unit tests are independent, I could use different names in each and it'll just work, furthermore the cost of 'messing' up is extremely tiny.  I don't think its a big deal at all and more than happy to skip this.

zacharylawrence: sgtm - doesn't seem like a big deal

jessex:
>On our app's side, protobuf seems like a clear win because of all the things Arian said and the strictly enforced fields (I'm hesitating to overload the word "schema" in this discussion).

>Yesterday, ohinds seemed firmly in the camp of proto not being too complex to ask jurisdictions to use.

Okay, I feel good about this then! Excited to see it develop.

jessex:
Thanks for the thoughtful, thrice duplicated response, @arian487. I was being dumb with my words because I was tired and stuck on a train. I meant that there was no such file matching that reference directly inside of the repository, but I see that, for now, we are checking the compiled Proto in, so I retract the comment. I think we're on the same page on the need to figure out tooling and workflow here, and I am bought into the value of using Protobuf for the API.

PR #259, opened 2018-12-06
Update data converter to expose all errors at once

terinpw:
Data converter becomes a class which raises a single ValueError whose message contains the error strings for all conversion issues for that ingest_info object.


Pipfile
```diff
@@ -64,6 +64,7 @@ flask = "==1.0.2"
 sqlalchemy = "*"
 us = "*"
 dateparser = "*"
+ipdb = "*"
```
jamwalla: do we need this? I don't see it used in this change.

terinpw: Is a debugger that I used, but installing it in my pipenv makes it appear in the pipfile. 

However, after discussions, I moved it into to be the dev packages, so nobody else has to install unless the explicitly call pipenv install --dev

terinpw:
Chatted w @jamwalla and @zacharylawrence, and the proto conversion is going to happen as a part of this class, so I think we can leave the static methods.

terinpw:
After conversation with @jamwalla  and @zacharylawrence, created converter_utils class for previously static methods.

PR #265, opened 2018-12-06
Pass scraper_start_date in scraper params

terinpw:
Any task that commits to our db needs to know the scraper_start_date in order to fill it on the booking. 

Pass this in as a param from scraper_start/resume. The entrypoint may change depending on our decisions re: docket, but until new implementation exists, I think it makes sense to leave it as is.


recidiviz/ingest/base_scraper.py
```diff
@@ -152,7 +154,10 @@ def _generic_scrape(self, params):
             if not ingest_info:
                 raise ValueError(
                     'IngestInfo must be populated if there are no more tasks')
-            persistence.write(ingest_info)
+            scraper_start_date = datetime.strptime(
```
jessex: I thought we had a utility somewhere for parsing of generic date strings... ah yes in `scraper_utils`. Would it make sense to reuse that here?

recidiviz/ingest/scraper.py
```diff
@@ -130,7 +139,9 @@ def start_scrape(self, scrape_type):
                                            scrape_type))
             return
 
-        params = {"scrape_type": scrape_type, "content": docket_item}
+        params = {"scrape_type": scrape_type,
+                  "content": docket_item,
+                  'scraper_start_date': self.get_now_as_str()}
```
jessex: We're mixing single and double quotes in this dict. Which should we be using generally?

recidiviz/ingest/scraper.py
```diff
@@ -228,7 +239,9 @@ def resume_scrape(self, scrape_type):
                     ScrapeKey(self.get_region().region_code, scrape_type))
                 return
 
-        params = {'scrape_type': scrape_type, 'content': content}
+        params = {'scrape_type': scrape_type,
+                  'content': content,
+                  'scraper_start_date': self.get_now_as_str()}
```
jessex: I wonder if there's any use for allowing API calls to pass a `scraper_start_date` in as a query parameter? To let people pretend that the scraper started at a specific time for testing or operational use cases?

recidiviz/ingest/base_scraper.py
terinpw: Yup, sure would. Good catch.

recidiviz/ingest/scraper.py
terinpw: I updated this to use double quotes (to match the rest of the dict), but we're a bit all over the place. The dictionary later in this file uses single quotes. 

We should probably just pick one as a repo style and try to maintain it.

recidiviz/ingest/scraper.py
terinpw: Oh interesting I hadn't originally thought of that. 

My first inclination is to not expose this via API. I think that since we're persisting this datetime with the data from live websites, we always will want the datetime to match the time the data was scraped. Even if we wanted to manually queue off scraper runs it seems like we would want the current datetime to be recorded. 

I cant think of any where we wouldn't want this; do you have any specific ones in mind?

terinpw:
Also adjusted last_scraped_date to be last_scraped_time and a datetime (instead of a date object). This way we can do multiple runs for a region in a single day and record the difference in timing.

recidiviz/ingest/scraper.py
jessex: Yes we should! Filed #270.

recidiviz/ingest/scraper.py
jessex: I don't have a specific use case in mind. And now that I think of it more, I realize that it would actually be fairly confusing since we'd have inconsistency between our self-reported scraper start date and what the time was in the real world when the attributes we scraped were stored. We can't actually go back and scrape the website at an earlier time. So ignore this!

PR #266, opened 2018-12-06
Unify flask apps into single app that delegates to blueprints

colincadams:
In the python 3 environment our application can only have a single entrypoint, not one per handler. This modifies our application to be run from a single entrypoint that creates a single flask app. It also means that all routing lives in flask instead of in appengine.

For further context, here is [a description of application startup](https://cloud.google.com/appengine/docs/standard/python3/runtime#application_startup) and [the `app.yaml` reference](https://cloud.google.com/appengine/docs/standard/python3/config/appref) as well as [an example `app.yaml`](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/appengine/standard_python37/building-an-app/building-an-app-1/app.yaml) for the python 3 environment.

This is compatible with python 2 so this pull request is targeting `master`.


cloud_datastore_admin.py
```diff
@@ -10,16 +10,16 @@
...
 
 from recidiviz.utils.auth import authenticate_request
 
-app = Flask(__name__)
+datastore_export = Blueprint('datastore_export', __name__)
 
-@app.route('/datastore_export')
+@datastore_export.route('/')
 @authenticate_request
-def datastore_export():
```
colincadams: Would be a separate PR but @jessex can this go away entirely?

base.yaml
```diff
@@ -6,25 +6,13 @@ builtins:
...
 - url: /_ah/pipeline.*
   script: recidiviz.mapreduce.lib.pipeline.handlers._APP
   login: admin
 
-- url: /calculator_pipeline
-  script: recidiviz.calculator.recidivism.pipeline.app
-
-- url: /datastore_export
-  script: cloud_datastore_admin.app
-  login: admin
```
colincadams: I removed the `login: admin` because as far as I can tell it is covered by the `@authenticate_request` decorator. Any concerns with this?

jessex: Sounds right, after your recent auth changes.

cloud_datastore_admin.py
jessex: Yeah I don't think we need Datastore export anymore. Even if we continue using it, it will be for information we don't require automatic (or any) export for. Feel free to remove.

main.py
```diff
@@ -0,0 +1,37 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+# =============================================================================
+
```
jessex: Does this file need to be in the root of the repository, or can we shift it into the `recidiviz` package?

base.yaml
```diff
@@ -6,25 +6,13 @@ builtins:
 - deferred: on
 
 handlers:
-- url: /scraper/start
```
jessex: I think with the change from `/scraper` to `/scraper_control`, we need to update `cron.yml`.

main.py
colincadams: I moved it into the `recidiviz` package and called it `server.py`. The only difference this way is that we will have to name the `entrypoint` explicitly in `app.yaml`.

https://cloud.google.com/appengine/docs/standard/python3/runtime#application_startup

base.yaml
colincadams: This still should be `/scraper/start` as defined in the url_prefix to `/scraper` in `server.py`. Not sure if this is what tripped you up, but I was very confused by this at first because the flask docs say "Additionally it will prefix the endpoint of the function with the name of the blueprint which was given to the Blueprint constructor (in this case also simple_page)." However this is referring to the endpoint name in flask, i.e. `scraper_control.start` that can be passed into the `url_for` function, not the url itself.

I opted to keep all the url's consistent in this PR. LMK if there are any worth updating here.

jessex: Yep, that is exactly what tripped me up. Thanks for the clarification, Colin!

PR #267, opened 2018-12-06
Set up alembic config to allow generating migrations from schema

zdg2102:


recidiviz/persistence/database/migrations/env.py
```diff
@@ -11,16 +14,18 @@
...
+    user = os.getenv('SQLALCHEMY_DB_USER')
+    password = os.getenv('SQLALCHEMY_DB_PASSWORD')
+    host = os.getenv('SQLALCHEMY_DB_HOST')
+    db_name = os.getenv('SQLALCHEMY_DB_NAME')
 
-# other values from the config, defined by the needs of env.py,
-# can be acquired:
-# my_important_option = config.get_main_option("my_important_option")
-# ... etc.
+    return 'postgresql://%s:%s@%s/%s' % (user, password, host, db_name)
```
jessex: All in on postgres!

recidiviz/persistence/database/migrations/env.py
```diff
@@ -11,16 +14,18 @@
...
-# from myapp import mymodel
-# target_metadata = mymodel.Base.metadata
-target_metadata = None
+# Metadata from schema
+target_metadata = Base.metadata
+
+
+def get_sqlalchemy_url():
+    """Returns string needed to connect to database"""
+    user = os.getenv('SQLALCHEMY_DB_USER')
```
jessex: Have you thought about where we'll store these in deployed environments? KMS?

zdg2102: Yeah, KMS seems to be the recommended way to do it. I'll have to look into exactly how it will work though.

colincadams: I'm currently looking at moving what we currently call `EnvironmentVariable` (where the proxy user and password are stored) from datastore to KMS. Assuming that goes smoothly you could store this using that class as well?

colincadams: Actually, KMS just manages keys and allows you to use them to encrypt/decrypt data, you can't store the data in KMS afaict. So it looks like we could either (1) keep storing secrets in datastore  or move them to gcs or (2) use KMS to encrypt them, set them as environment variables via the yaml file, and then have the server decrypt them via KMS at start up.

I couldn't find any best practices for this and no advantages jump out at me so I would say keep using datastore. Any other thoughts on this @jessex or @zdg2102?

jessex: What I am accustomed to from past experience would be to store the secrets at certain paths in a dedicated secret management service like Vault, and then read those into environment variables at startup with a library like Summon. I naively thought KMS was a secrets service, but I recognize it's just keys. But that's still useful!

[I read this article](https://cloud.google.com/kms/docs/secret-management) on secrets management with KMS, and I think what @colincadams is suggesting is basically approach 2 in that article, but with Datastore instead of Object Storage. I think that's a reasonable starting approach since we already have the infrastructure set up for it, and I think we could use that approach for @zdg2102's database secrets here, but it might be worth a later design sprint.

zdg2102: Looks like @colincadams has #275 in the works to handle this. I'll hold off on this till that change is finished and then merge it in here.

recidiviz/persistence/database/migrations/env.py
zacharylawrence: Does there exist a way to connect to the database without needing to concat strings to build the url? For example, I think I remember `pyodbc` letting us pass in a kwargs.

zdg2102: For whatever reason, using a string like this seems to be the standard method for creating an engine: https://docs.sqlalchemy.org/en/latest/core/engines.html

I don't think it's too unmanageable, since it's a stable format that we shouldn't need to update again after this.

zacharylawrence: SGTM - I think I was remembering this, but it seems like you're right that we don't need this level of customization: https://docs.sqlalchemy.org/en/latest/core/engines.html#custom-dbapi-args.

recidiviz/persistence/database/migrations/env.py
zdg2102: Updated to read from secrets.yaml instead

PR #271, opened 2018-12-07
Add a test to make sure there is no drift between proto and object

arian487:


recidiviz/tests/ingest/models/ingest_info_test.py
```diff
@@ -0,0 +1,65 @@
...
+    pass
+
+
+def test_proto_fields_match():
+
+    def _verify_fields(proto, ingest_info_source, ignore=None):
+        ignore = ignore or []
+        proto_fields = [field.name for field in proto.DESCRIPTOR.fields]
+        source_fields = vars(ingest_info_source)
+        for field in proto_fields:
```
jessex: I guess the alternative implementation is to create a `set` for each array of fields and then do a set difference, but getting that approach to the ability to raise errors for either side of the equation would end up with just about the same amount of code.

recidiviz/tests/ingest/models/ingest_info_test.py
```diff
@@ -0,0 +1,65 @@
...
+
+        for field in source_fields:
+            if field not in proto_fields and field not in ignore:
+                raise FieldsDontMatchError(
+                	   "Field '%s' exists in '%s'"
+                    " IngestInfo object but not in the proto object" % (
+                        field, proto.__name__))
+
+
+    person_fields_ignore = ['booking_ids', 'booking']
```
jessex: Do you think it's obvious why these are ignored?

recidiviz/tests/ingest/models/ingest_info_test.py
arian487: I actually wrote it out this way the first time and came to the same conclusion :)

recidiviz/tests/ingest/models/ingest_info_test.py
arian487: This current set is all related to fields in either object that help with object linking (id references, or the objects themselves).  But I added an ignore option because I casn see other fields that way may want to ignore in this test in the future.

PR #272, opened 2018-12-07
Ingest info __setattr__

jamwalla:
override __setattr__ in IngestInfo objects to disallow setting new fields that don't already exist.


recidiviz/ingest/models/ingest_info.py
```diff
@@ -31,6 +31,12 @@ def __eq__(self, other):
     def __repr__(self):
         return to_string(self)
 
+    def __setattr__(self, name, value):
+        if hasattr(self, 'person') and not hasattr(self, name):
```
jessex: Since they're all identical except for a single var, can we extract this out to a helper?

jamwalla: good idea!

PR #273, opened 2018-12-07
Add validator for ingest_info protos

zacharylawrence:
This logic will allow validating the structure of ingest_info protos sent to the persistence API.

In a future PR once the persistence layer is restructured to use the new ingest_info proto, the result of this validator will be used to create a JSON error response when a malformed ingest_info proto is sent to the persistence API.


recidiviz/persistence/validator.py
```diff
@@ -0,0 +1,140 @@
...
+
+    Returns:
+        non_existing_ids: ids that have been referenced but don't exist
+        extra_ids: ids for rows that exist in ingest_info but aren't referenced
+    """
+    non_existing_ids = {}
+    extra_ids = {}
+
+    # Bookings
+    booking_ids = {booking.booking_id for booking in ingest_info.bookings}
```
jessex: Minus differences in `flatten`, these blocks are all performing the same set difference operation across the same conceptual fields. Would it be clean to extract that out to a helper method that takes in the object and the id attribute name and return the non-existing ids and extra ids for that sub-entity? Then we just invoke that a small handful of times, placing the results of each call into a map?

recidiviz/persistence/validator.py
```diff
@@ -0,0 +1,140 @@
...
+        'charges': _get_duplicates(
+            [charge.charge_id for charge in ingest_info.charges]),
+        'bonds': _get_duplicates(
+            [bond.bond_id for bond in ingest_info.bonds]),
+        'sentences': _get_duplicates(
+            [sentence.sentence_id for sentence in ingest_info.sentences]),
+    }
+
+
+def _get_duplicates(collection):
```
jessex: I kind of expect `duplicates` and `flatten` operations to be in some functional package somewhere in the standard library or a quasi-standard library.

recidiviz/persistence/validator.py
```diff
@@ -0,0 +1,140 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+# =============================================================================
+"""Validates that ingest_info protos conform to the schema requirements."""
+
+
+def get_errors(ingest_info):
```
colincadams: nit: I would call this `validate` instead of `get_errors`

recidiviz/persistence/validator.py
```diff
@@ -0,0 +1,140 @@
...
+
+    Returns:
+        A dictionary of all validation errors
+    """
+    non_existing_ids, extra_ids = _get_malformed_ids(ingest_info)
+    errors = {
+        'duplicate_ids': _get_duplicate_ids(ingest_info),
+        'ids_referenced_that_do_not_exist': non_existing_ids,
+        'ids_never_referenced': extra_ids
+    }
```
colincadams: It's hard to know which structure is more useful without knowing how it is called, but I could also see this structure getting flipped like this:
from
```
errors:
  duplicate_ids: 
    people: [x,y]
    bookings: []
    ...
  ids_referenced: ...
```

to
```
errors:
  people:
    duplicate_ids: [x,y]
    ids_never_referenced: [z]
    ...
  bookings: 
  ...
```

Is there a reason you decided to do it this way? The only reason I present the other way is I think the code structure might be more readable as you could fill all the errors for a particular type at once, but they really aren't that different.

```
def validate(ingest_info):
  return {
    'bookings': validate_bookings(ingest_info),
  }

def validate_bookings(ingest_info):
  ...
  return Errors(duplicate, dne, not_referenced)
```

recidiviz/persistence/validator.py
```diff
@@ -0,0 +1,140 @@
...
+                           ingest_info.charges if
+                           charge.HasField('bond_id')}
+
+    non_existing_ids['bonds'] = referenced_bond_ids - bond_ids
+    extra_ids['bonds'] = bond_ids - referenced_bond_ids
+
+    return non_existing_ids, extra_ids
+
+
+def _flatten(collection):
```
colincadams: This can be replaced with [`itertools.chain`](https://docs.python.org/2/library/itertools.html#itertools.chain).

recidiviz/persistence/validator.py
zacharylawrence: I think how the response structure is going to be used is flexible since we're currently planning to just return this as JSON to the user. I tried swapping it around like described, what do you think of this? I like that it makes the methods shorter and separates logic based on table type (eg. when to flatten).

recidiviz/persistence/validator.py
colincadams: I've never used `iteration_utilities` before. `itertools` is a builtin module so that's why I recommended it, but it looks like `iteration_utilities.flatten` just calls `itertools.chain` [under the hood](https://github.com/MSeifert04/iteration_utilities/blob/master/iteration_utilities/_recipes.py#L137), and flatten is a better name. It also looks to implement duplicate finding in C so it seems like a good way to go for that.

zacharylawrence: SGTM - I'll leave it as `iteration_utilities` then. Thanks for the deep dive explanation!

recidiviz/persistence/validator.py
jamwalla: drive-by:
 - `iteration_utilities.flatten` looks like `itertools.chain`
 - `iteration_utilities.unique_everseen` shouldn't be necessary with `set`
 - in general you can skip the square brackets around comprehensions inside function calls - i.e. `flatten([x for x in y])` == `flatten(x for x in y)` and saves an iteration

but I'm surprised there's not a simple standard library approach for duplicates!

jamwalla: or better, `set(itertools.chain.from_iterable(ingest_info.people))`!

recidiviz/persistence/validator.py
jessex: Yeah that's right, it gets ugly fast. DRY is a neat guideline but not an unambiguous mandate. :+1:

recidiviz/persistence/validator.py
zacharylawrence: Yeah, good point with `iteration_utilities.unique_everseen` - removed. Also removed the extraneous `[` `]`. Thanks for the tips!

PR #274, opened 2018-12-10
converter: parse ethnicity when set as race

jamwalla:
in the data converter, check if the `person.race` field is set with an ethnicity instead of a race (as in #257) and set the appropriate field.


recidiviz/persistence/converter.py
```diff
@@ -116,8 +116,16 @@ def _convert_person(self, ingest_person):
                                           ingest_person.gender)
 
         if ingest_person.race is not None:
-            person.race = self._convert(converter_utils.string_to_enum, 'race',
-                                        ingest_person.race)
+            if converter_utils.enum_contains_value('ethnicity',
+                                                   ingest_person.race) and \
+                not converter_utils.enum_contains_value('race',
+                                                        ingest_person.race):
+                person.ethnicity = self._convert(converter_utils.string_to_enum,
```
jessex: What do we think of the case where `ingest_person.race` and `ingest_person.ethnicity` are both set, and there is an ethnicity in both fields? We'd set `person.ethnicity` right here and then override it again at line 131. If the values are the same, i.e. if the ingest source duplicate ethnicity values across their ethnicity and race fields, then no big deal. But if they are different, then we might be losing information. It's also conceivable, though really rare, I'd guess, that the opposite of this condition could happen, where a `race` value is set on `ingest_person.ethnicity`. 

Are those cases we need to handle?

jamwalla: Good call. I think we should special-case only when race is set to an ethnicity and ethnicity is not set. In other words, we should always handle the pair together, and let the error get raised if the race field is an ethnicity and the ethnicity field is set.

I'm inclined to not think about ethnicity set as race until it comes up on a webpage - more likely that would be a mistake the scraper writer should correct.

Let me know what you think!

recidiviz/persistence/converter.py
```diff
@@ -111,21 +112,26 @@ def _convert_person(self, ingest_person):
...
-                                                 'ethnicity',
-                                                 ingest_person.race)
-            else:
-                person.race = self._convert(converter_utils.string_to_enum,
-                                            'race', ingest_person.race)
-
-        person.ethnicity = self._convert(converter_utils.string_to_enum,
-                                         'ethnicity',
-                                         ingest_person.ethnicity)
+        # If |ingest_person.race| is really the ethnicity field *and* the
```
jessex: This logic makes sense to me! But I'd love if someone with more direct experience with the data converter took a quick look, like Zipper Zach or Terin or Arian.

recidiviz/tests/persistence/converter_test.py
```diff
@@ -48,6 +48,12 @@ def test_parsePersonFields(self):
         assert schema_person.ethnicity == constants.person.Ethnicity.HISPANIC
         assert schema_person.place_of_residence == 'NNN STREET ZIP'
 
+    def test_parseEthnicityAsRace(self):
```
zacharylawrence: This is backwards right? You're parsing the field labeled `race` as `ethnicity`?

recidiviz/persistence/converter_utils.py
```diff
@@ -41,6 +41,13 @@ def normalize(s):
     return ' '.join(s.split()).upper()
 
 
+def enum_contains_value(enum_type, value):
```
zacharylawrence: Nit: It would be nice to do a `contains` check instead of looking for an exception here. Although I don't think it's a huge deal and we can come back to it in a future PR if you want.

recidiviz/persistence/converter.py
zacharylawrence: Logic here looks good to me!

Quick comment on readability: This comment feels like it's just explaining what the code is doing (instead of explaining WHY the code is doing something). Maybe we can improve readability by just refactoring the code instead? For example, you could extract a method like:
```
def race_is_actually_ethnicity(ingest_person):
            return (converter_utils.enum_contains_value('ethnicity', ingest_person.race) 
                    and not converter_utils.enum_contains_value('race', ingest_person.race))
```
then this check would become:
```
if (ingest_person.race is not None
    and ingest_person.ethnicity is None
    and race_is_actually_ethnicity(ingest_person):
```

To me, this feels like it conveys everything that this comment currently does. If you still think this section needs a comment, maybe something along the lines of `If the race field is actually an ethnicity then assume this is a mistake in the provided data` (or something else more gracefully worded than that). This lets us explain to the reader why there's a need for swapping these fields.

Thoughts on this?

recidiviz/persistence/converter_utils.py
jamwalla: this is way more correct, since it doesn't return false when `enum_type` isn't right. thanks!

jamwalla:
thanks, Zach! I like this a lot better. ptal!

recidiviz/persistence/converter_utils.py
zacharylawrence: LGTM - thanks!

jessex:
Great recommendations, @zacharylawrence!

PR #276, opened 2018-12-10
 Move secrets to new datastore api

colincadams:


colincadams:
Note that right now this PR is based on the `python3` branch, which doesn't include changes from #275 yet, so I also cherry-picked the change from that (f088216) onto this branch. Once #275 is merged I will rebase `python3` from `master` so that this only has the changes from the most recent commit.

recidiviz/utils/secrets.py
```diff
@@ -19,49 +19,61 @@
...
-    Entity keys are expected to be in the format region_varname, to enforce
-    uniqueness.
 
-    Attributes:
-        name: (string) Variable name
-        value: (string) Variable value, set by admins in gcloud console
-    """
-    name = ndb.StringProperty()
-    value = ndb.StringProperty()
+def clear_ds():
```
jessex: Idle question: are we using `clear_ds` anywhere outside of tests? If not, should we create an annotation like `@environment.test_only`? I've used similar annotations in Java-land, and also used just marking methods as package-private and documenting them as test-only, but we don't so much have access to the latter here.

colincadams: I like the idea of creating that annotation. I'll create it in a separate pr off of `master` so that others can use it if needed and then create another pr to use it for these `clear_ds` methods on `python3`.

PR #280, opened 2018-12-11
Create canonical string values for enums

zdg2102:
Also rename enums in schema.py


zdg2102:
Note: This change only updates schema.py. I'll be making a separate followup PR to update the application layer enums and tests to follow the same pattern.

recidiviz/common/constants/bond.py
```diff
@@ -28,7 +28,7 @@ class BondType(object):
 
 class BondStatus(object):
     # TODO: resolve inconsistency between Python and DB enum values
-    ACTIVE = 'Active'
+    ACTIVE = 'ACTIVE'
```
jessex: I think I may be misunderstanding the note in the canonical strings class, but I don't yet get why these classes don't refer to those same constants, e.g. `ACTIVE = enum_strings.bond_status_active`?

zdg2102: Sorry, that was what I meant by my previous comment. I'm currently working on a separate PR to make that change across all the enums and tests. These minor string changes are just to temporarily get the tests passing on this PR to clear the way for the followup.

jessex: Oh, duh. Sorry I stopped reading in the 90s.

zdg2102:
@jamwalla I'll add you as a reviewer once it's done, but on the followup PR I'm removing `UNKNOWN` everywhere. I'm going to have another followup PR after that where you and I can figure out exactly which enums should and shouldn't have `UNKNOWN` or `NOT_PROVIDED`

PR #285, opened 2018-12-11
Add a base scraper test for all tests to inherit

arian487:
This adds a util base class that all scraper writers can/should inherit.  It runs some extra validations on the outputs, as well as verifies that the ingest_info can be properly converted to the proto and its output validated.  

Any nice utilities for testing scrapers should go in here in the future.


recidiviz/tests/utils/base_scraper_test.py
```diff
@@ -0,0 +1,138 @@
...
+
+        # Attempt to convert the result to the final proto and validate its
+        # output.
+        result_proto = ingest_info_utils.convert_ingest_info_to_proto(result)
+        validate(result_proto)
+
+        assert result == expected_result
+        return result
+
+
```
jamwalla: nit, extra newlines

recidiviz/tests/utils/base_scraper_test.py
```diff
@@ -0,0 +1,135 @@
...
+            with open(self.yaml, 'r') as ymlfile:
+                manifest = yaml.load(ymlfile)
+
+            person = ingest_info.IngestInfo().create_person()
+            booking = person.create_booking()
+            charge = booking.create_charge()
+            arrest = booking.create_arrest()
+            sentence = charge.create_sentence()
+            bond = charge.create_bond()
+            object_verification_map = {
```
jamwalla: should this map live somewhere else? How much does this overlap with errors that would get thrown from the data extractor?

recidiviz/tests/utils/base_scraper_test.py
```diff
@@ -0,0 +1,135 @@
...
+
+"""Base test class for scrapers."""
+import yaml
+from recidiviz.ingest import ingest_info_utils
+from recidiviz.ingest.models import ingest_info
+from recidiviz.persistence.validator import validate
+
+
+class BaseScraperTest(object):
+    """A base class for scrapers which does extra validations."""
```
jessex: Nit: "A base class for scraper tests..." ?

recidiviz/tests/utils/base_scraper_test.py
```diff
@@ -0,0 +1,135 @@
...
+                    raise AttributeError(
+                        "Attribute %s is unknown on %s, found in key_mappings"
+                        % (attr, class_to_set))
+
+            if 'multi_key_mappings' in manifest:
+                for value in manifest['multi_key_mappings'].values():
+                    class_to_set, attr = value.split('.')
+                    if attr not in vars(object_verification_map[class_to_set]):
+                        raise AttributeError(
+                            "Attribute %s is unknown on %s, found in"
```
jessex: Nit: need another space at the end of this line.

recidiviz/tests/utils/base_scraper_test.py
```diff
@@ -0,0 +1,135 @@
...
+            extra validations on the output.
+        """
+        result = self.scraper.get_more_tasks(content, params)
+
+        # Make sure at minimum endpoint and tasktype are returned.
+        for key in ['endpoint', 'task_type']:
+            for param in result:
+                if key not in param:
+                    raise AttributeError(
+                        "No value %s exists in the"
```
jessex: Another nit: maybe reword this slightly? I think "No value for 'endpoint' exists" is more sensible than "No value endpoint exists", for example.

recidiviz/tests/utils/base_scraper_test.py
arian487: So I think some of the errors will be thrown from data extractor but definitely not all.  This catches nonsense keys and also very clearly lets the user know if some values are nonsense.  In my experience its good to have some overlapping redundancy in unit tests because its hard to predict how a user might break things down the line.  For example they could write a unit test and not actually call populate_data in the test (but this way at least some validation is done on their yaml before things ship to prod).

I think lets leave this here for now, and if someone else uses this map (to my knowledge none do), we can move it somewhere else.

recidiviz/tests/utils/base_scraper_test.py
arian487: Fixed

recidiviz/tests/utils/base_scraper_test.py
arian487: Fixed.

recidiviz/tests/utils/base_scraper_test.py
arian487: Fixed.

PR #290, opened 2018-12-11
remove calculate dependency on mapreduce, ndb

colincadams:


recidiviz/calculator/recidivism/identifier.py
```diff
@@ -62,6 +59,9 @@ def find_recidivism(person, include_conditional_violations=False):
 
     Args:
         person: a person to determine recidivism for.
```
jessex: This line can be removed now.

recidiviz/calculator/recidivism/pipeline.py
```diff
@@ -23,21 +23,24 @@ class that initiates pipelines on web requests.
...
+    'total_people_mapped': 0,
+    'total_metric_combinations_mapped': 0,
+    'unique_metric_keys_reduced': 0,
+    'total_records_reduced': 0,
+    'total_recidivisms_reduced': 0,
+}
+
 
-def map_person(person):
+# placholder - should be mapreduce execution id
```
jessex: `placeholder`

jessex:
I think there's more configuration we can remove now, including but not limited to:
* A ton of library code under `/lib`
* The /_ah/* config from base.yaml
* The module and version config from local_app.yaml
* The mapreduce.yaml file

Take a look at this [commit](https://github.com/Recidiviz/recidiviz-data/commit/641a9e792e69eaaf889e460892e31e2fdfcce734) from when calculation was first added to the repository for context.

colincadams:
alright I removed what I believe is everything except for the `libs/` changes. It's hard to tell if any other parts of the system have started to depend on those packages since they were added or not and we need to just delete the whole `libs/` directory eventually.

colincadams:
Just kidding, I removed the packages from `libs/` as well. Looks like the only one that was still depended on is `pbr`.

PR #293, opened 2018-12-12
create_scraper script use BaseScraperTest

jamwalla:
updates the create_scaper script to use the new base test class (#285)


jamwalla:
added a pair of comments, ptal thanks!

PR #296, opened 2018-12-12
Update alembic URL to support SSL connections

zdg2102:
Also update env.py strings to use format() instead of %s


recidiviz/persistence/database/migrations/env.py
```diff
@@ -17,15 +17,22 @@
...
     """Returns string needed to connect to database"""
-    user = secrets.get_secret('SQLALCHEMY_DB_USER')
-    password = secrets.get_secret('SQLALCHEMY_DB_PASSWORD')
-    host = secrets.get_secret('SQLALCHEMY_DB_HOST')
-    db_name = secrets.get_secret('SQLALCHEMY_DB_NAME')
 
-    return 'postgresql://%s:%s@%s/%s' % (user, password, host, db_name)
+    use_ssl = secrets.get_secret('sqlalchemy_use_ssl')
+
+    if use_ssl == True:
```
zacharylawrence: `use_ssl` has to be either True or False, right?

zdg2102: It's yaml, so I'm fairly confident you could write "use_ssl: pigeon" and it would still run, which is why I set up the check this way.

zacharylawrence: Does the yaml value get parsed as a String? If you're comparing that to a boolean `True` or `False` then won't everything evaluate to False?

```
In [1]: "True" == True
Out[1]: False
```

```
In [2]: "False" == False
Out[2]: False
```

zdg2102: Good question. I haven't been able to run this on local yet because my environment is messed up. I'm trying to fix that, then I'm going to run this and update depending on how it behaves.

Issue #302, opened 2018-12-13
Consider breaking up the data converter


zacharylawrence:
Look into the feasibility of breaking up the `ingest_info proto` -> `persistence layer entities` converter which will help isolate code (prevent lots of logic in one file) & make it easier to write more in depth tests (eg. comprehensive tests for parsing a booking). We should also write those more detailed tests.

PR #303, opened 2018-12-13
Update data converter to use ingest_info.proto

zacharylawrence:
This PR updates the data converter to convert `ingest_info.proto` -> `persistence layer entities` instead of `ingest_info.py` -> `schema ORM object`.

As explained in #302, we should consider breaking apart this converter to further isolate code (this file is starting to contain a lot of logic). This has the additional benefit of allowing us to write more detailed tests against each compontent (eg. when booking has `age` and `DOB`). We should write those tests in the future. Note that these changes are not included now in the interest of testing an end-to-end prod run.


recidiviz/persistence/converter.py
```diff
@@ -15,192 +15,166 @@
...
 
-        person.gender = self._convert(converter_utils.string_to_enum,
-                                      'gender',
-                                      ingest_person.gender)
+        if ingest_person.HasField('gender'):
+            person.gender = self._convert(converter_utils.string_to_enum,
+                                          'gender',
+                                          ingest_person.gender)
 
-        if ingest_person.race is not None and ingest_person.ethnicity is None \
```
jamwalla: did we mean to lose this way of handling ethnicity in the race field?

recidiviz/tests/persistence/converter_test.py
```diff
@@ -16,154 +16,315 @@
 # =============================================================================
 """Tests for data converter."""
 
-from datetime import datetime
 import pytest
+import unittest
```
jamwalla: can these all be pytests?

recidiviz/persistence/converter.py
zacharylawrence: Great catch, updated here & in the test.

recidiviz/tests/persistence/converter_test.py
zacharylawrence: Talking with @rasmi it seems like we should prefer `unittest`. I updated below to use `self.assertThrows()` which allows us to avoid remove the dependency here on `pytest`.

recidiviz/persistence/converter.py
```diff
@@ -15,192 +15,174 @@
...
-            and converter_utils.race_is_actually_ethnicity(ingest_person):
+        if ingest_person.HasField('race') \
+                and not ingest_person.HasField('ethnicity') \
+                and converter_utils.race_is_actually_ethnicity(ingest_person):
             # TODO(289): determine how to properly set race here once UNKNOWN
```
jamwalla: doesn't have to be part of this change, but since we're leaving things blank instead of using `UNKNOWN` we can delete this todo

recidiviz/persistence/converter.py
```diff
@@ -211,111 +193,138 @@ def _convert_booking(self, ingest_booking):
...
-
-        if ingest_charge.sentence:
-            charge.sentence = self._convert_sentence(ingest_charge.sentence)
+        if ingest_charge.HasField('name'):
+            charge.name = self._convert(converter_utils.normalize,
+                                        ingest_charge.name)
+
+        if ingest_charge.HasField('attempted'):
+            charge.attempted = self._convert(
+                converter_utils.verify_is_bool,
```
jamwalla: can you move `bool(strtobool(...))` inside `converter_utils.verify_is_bool` - and maybe rename to `string_to_bool` since we no longer expect a bool?

recidiviz/tests/persistence/converter_test.py
```diff
@@ -16,154 +16,331 @@
...
+        # Arrange
+        ingest_info = IngestInfo()
+        ingest_info.people.add(race='HISPANIC')
 
-        schema_person = self.converter._convert_person(ingest_person)
+        # Act
+        result = converter.convert(ingest_info)
+
+        # Assert
         # TODO(289): check race once the proper behavior here has been
```
jamwalla: you solved this todo!

recidiviz/persistence/converter.py
zacharylawrence: Done.

recidiviz/persistence/converter.py
zacharylawrence: Good plan - I went ahead and renamed this to `parse_bool` too.

recidiviz/tests/persistence/converter_test.py
zacharylawrence: Done!

jamwalla:
🙌🏽

PR #304, opened 2018-12-13
Exposes API to infer release of people

terinpw:
- Created infer_release.py which contains an API to infer releases of people in the supplied region
- Moved helper methods from scraper_control to ingest_utils that are needed by infer_release
- Moved helper methods from scraper_utils into common_utils to prevent circular dependency in persistence (was throwing errors with my new test).


recidiviz/ingest/infer_release.py
```diff
@@ -0,0 +1,48 @@
...
+from recidiviz.utils.params import get_values
+from recidiviz.persistence import persistence
+
+infer_release_blueprint = Blueprint('infer_release', __name__)
+
+
+@infer_release_blueprint.route('/release')
+@authenticate_request
+def infer_release():
+    log = logging.getLogger()
```
arian487: This is unnecessary, you can just do logging.error which will use the default logger like you're getting here.

recidiviz/ingest/ingest_utils.py
```diff
@@ -0,0 +1,77 @@
+# Recidiviz - a platform for tracking granular recidivism metrics in real time
```
arian487: Can you merge the ingest_info_utils with this.  Doesn't make sense having an ingest_utils and an ingest_info_utils.  I think you can take all of the code in ingest_info_utils and put it in here (the name makes more sense).

recidiviz/ingest/infer_release.py
terinpw: Cool done.

recidiviz/ingest/ingest_utils.py
terinpw: Ah didn't realize that file existed. Moved ingestion_info_utils code into ingest_utils (and same with tests).

PR #305, opened 2018-12-13
Add a converter to convert from entity to the database object

arian487:
quick converter to go from entity object to database object


recidiviz/persistence/database/database_utils.py
```diff
@@ -0,0 +1,72 @@
...
+
+from recidiviz.persistence.database import schema
+
+
+def convert_person_to_database(person):
+    """Converts the given person to a database person."""
+    person_db = schema.Person()
+    for k, v in vars(person).items():
+        if k == 'bookings':
+            person_db.bookings = [convert_booking_to_database(b) for b in
```
zacharylawrence: Do you know if you can set the bookings on the ORM object to a `List` like this. You might need to use the `.extend()` method.

recidiviz/persistence/database/database_utils.py
```diff
@@ -0,0 +1,72 @@
...
+                                  person.bookings]
+        else:
+            setattr(person_db, k, v)
+    return person_db
+
+
+def convert_booking_to_database(booking):
```
zacharylawrence: Does it make sense to make these private? Would we need to expose anything other than converting the top level `Person`?

recidiviz/persistence/database/database_utils.py
arian487: Discussed offline, it works

recidiviz/persistence/database/database_utils.py
arian487: Done.

PR #309, opened 2018-12-13
Allow for two way conversion

arian487:


recidiviz/persistence/database/database_utils.py
```diff
@@ -16,59 +16,88 @@
...
+from recidiviz.persistence import entities
 from recidiviz.persistence.database import schema
 
 
-def convert_person_to_database(person):
-    """Converts the given person to a database person."""
-    person_db = schema.Person()
-    for k, v in vars(person).items():
+def convert_person(person_src):
+    """Converts the given person."""
```
zacharylawrence: Cool idea! This change LGTM.

Could you update the pydoc to explain that this method works on both types of objects? You could also consider 2 "public" functions that both just call into this one using a `bool` instead of referencing the type - but if you update the pydoc well enough that level of clarity might not be necessary? Thoughts?

arian487: I thought about doing the two public functions and i think its not too necessary, I've updated the pydoc.

PR #312, opened 2018-12-13
Add from_str to each enum

zacharylawrence:
This follows closely to the example explained in: https://stackoverflow.com/a/49060635


recidiviz/persistence/converter_utils.py
```diff
@@ -42,28 +42,20 @@ def normalize(s):
...
-        enum_type: The enum type to convert to.
-        value: The string representation of an enum of the provided enum_type
-
-    Return:
-        (enum) Converted enum of type enum_type.
-    """
-    return global_map.convert_value_to_enum(
-        enum_type, normalize(value))
+def race_is_actually_ethnicity(ingest_person):
+    try:
```
zacharylawrence: @jamwalla If we don't want to expose the internal `dict`, I'm not sure if there's a better way to do this? Any thoughts?

jamwalla: i don't mind catching the KeyError, I think it's a pattern that gets used

zacharylawrence: Great SGTM, thanks!

PR #313, opened 2018-12-14
Update SQLAlchemy URL to use correct SSL params

zdg2102:
Also convert secrets.yaml variables in env.py back to regular environment variables. As migrations will never be triggered remotely, these values can be set on a local machine rather than needing to rely on secrets.yaml.


zacharylawrence:
This LGTM, but you could still use a yaml file that just gets passed around and read all locally (ie. doesn't use the recidiviz.utils.secrets and instead just reads the yaml on local disk)? This would just let Josh or someone send out a `database_secrets.yaml` that contains all the secrets instead of having to set multiple env variables. I'm not sure exactly your goal with this, but thoughts?

zdg2102:
This change is partially motivated by the fact that importing google.appengine in secrets util is causing namespace problems with google.protobuf, so keeping it as simple as possible with regular environment variables is what I'm aiming for at least for the current moment.

I could see sometime down the line consolidating these secrets just for convenience, but they'll have to be partially set manually no matter what because the key and cert paths will be local to a given machine. So I think it should be fine to leave as-is for now, and we can make it more convenient once we've finished the Python 3 migration, which should fix the import problem.

PR #315, opened 2018-12-14
Minor schema updates in preparation for initial migration

zdg2102:
- Remove `UNKNOWN` and `NOT_PROVIDED` enum strings

- Add `RESOLVED_UNKNOWN_REASON` to `ReleaseReason` in both application and DB enums and remove `release_date_inferred` column

- Remove PII columns from `PersonHistory` table

- Remove `offense_code` and `number_of_counts` columns from `Charge` and `ChargeHistory`


zdg2102:
Renamed `RESOLVED_UNKNOWN_REASON` to `INFERRED_RELEASE`

recidiviz/persistence/database/schema.py
```diff
@@ -183,6 +184,11 @@ class PersonHistory(Base):
     """Represents the historical state of a person"""
     __tablename__ = 'PersonHistory'
 
+    # NOTE: PersonHistory does not contain surname, given_names, or birthdate
```
jessex: How do we handle a person changing names?

zdg2102: The fields will be updated on master and there will be no history that anything was ever different. Since at least for now name essentially acts as an internal-only field used exclusively for entity matching, it's only ever relevant on the most recent snapshot of an entity, which is all we compare against to see if a new scrape matches an existing entity.

jessex: Mm, fair enough!

PR #317, opened 2018-12-14
Move region into person instead of on booking

terinpw:
- moved region from booking schema to person schema
- updated persistence.write to take in a region name 
- removed region from ingest_info

Co-Authored-By: jamwalla


jamwalla:
close #306 close #314

PR #319, opened 2018-12-14
Add initial generated migration to create tables

zdg2102:


recidiviz/persistence/database/migrations/versions/5b129798a95d_create_tables.py
```diff
@@ -0,0 +1,329 @@
+"""create_tables
+
+Revision ID: 5b129798a95d
+Revises:
```
jessex: Is this meant to be manually filled in, i.e. "here are the tables/structures being revised by this migration"?

recidiviz/persistence/database/migrations/versions/5b129798a95d_create_tables.py
```diff
@@ -0,0 +1,329 @@
...
+    sa.ForeignKeyConstraint(['booking_id'], ['Booking.booking_id'], ),
+    sa.PrimaryKeyConstraint('hold_id')
+    )
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_table('Hold')
+    op.drop_index(op.f('ix_Charge_external_id'), table_name='Charge')
```
jessex: I'm vaguely surprised we have to manually drop indices before dropping their respective tables.

jessex:
If we don't already, can we add docs to one of our readme's describing how to implement/generate new database migrations? Or file away an issue for such if we have more of the process still to hammer out?

recidiviz/persistence/database/migrations/versions/5b129798a95d_create_tables.py
zdg2102: "Revises" is the ID of the previous migration, so it's blank since this is the first one.

recidiviz/persistence/database/migrations/versions/5b129798a95d_create_tables.py
zdg2102: Since this is auto-generated code, could be that it's not strictly required, but just easier for consistently handling some other edge cases? Not sure though.

zdg2102:
@jessex yep, that's coming in a followup PR

PR #322, opened 2018-12-14
UsNy --> BaseScraper

ohinds:
- Use BaseScraper as the base class for the UsNy region

- Fully tested using BaseScraperTest

- Had to add some special-case logic in BaseScraper to allow passing
  the person details page through to grab data in the case we don't
  get a disambiguation page. This is because we aren't sure when we
  make the request whether we'll need to SCRAPE_DATA or
  GET_MORE_TASKS.

- The UsNyPerson and UsNyRecord remain for now, until we rip out all
  ndb stuff.


recidiviz/ingest/scraper_utils.py
```diff
@@ -156,17 +156,18 @@ def calculate_age(birthdate, check_date=None):
         ((check_date.month, check_date.day) < (birthdate.month, birthdate.day))
 
 
-def get_id_value_from_html_tree(html_tree, html_id):
-    """Retrieves the value of the given id from the given html tree.
+def get_value_from_html_tree(html_tree, html_id, tag='id'):
```
jamwalla: can we change `tag` to like `id_attribute_name`? We should document that we expect `tag`=`html_id` attribute to uniquely identify an element.

recidiviz/ingest/us_ny/us_ny_scraper.py
```diff
@@ -41,1226 +37,286 @@
...
-        return last_scraped
+            # Find the particular DIN we were after on this click.
+            result_list = [
+                res for res in
+                content.xpath('//table/tr/td[@headers="din"]/form')
+                if res.xpath('div/input[@type="submit"]')[0].value == din
+            ]
 
+        params_list = []
+        for row in result_list:
```
arian487: I see this pattern a few times (setting the data by traversing the data fields), can you pull it out into a separate helper?

recidiviz/ingest/us_ny/us_ny_scraper.py
```diff
@@ -41,1226 +37,286 @@
...
+
+            # Get whether the charge was an attempt
+            attempted = 'False'
+            if charge_name.lower().startswith('att'):
+                attempted = 'True'
+                charge_name = ' '.join(charge_name.split()[1:])
+
+            ingest_info.person[0].booking[0].create_charge(
+                attempted=attempted,
+                charge_class='FELONY',
```
arian487: Can we use the actual enum here instead of the string?  ChargeClass.FELONY

recidiviz/ingest/us_ny/us_ny_scraper.py
```diff
@@ -41,1226 +37,286 @@
...
+                degree=charge_degree,
+                level=charge_level,
+                name=charge_name.strip(),
+                status='SENTENCED',
```
arian487: Use the enum here as well

recidiviz/ingest/us_ny/us_ny_scraper.py
```diff
@@ -41,1226 +37,286 @@
...
+            charge_name = row.xpath('td[1]')[0].text.strip().rstrip()
+
+            if not charge_name:
+                break
+
+            # Change 'class' into 'level'
+            charge_class_str = row.xpath('td[2]')[0].text
+            charge_level = None
+            if charge_class_str.startswith('A'):
+                charge_level = 'ONE'
```
arian487: Hm, I see we don't have an enum for charge level.  Should we have one?  Doesn't have to be a part of this PR just posing the question.

recidiviz/ingest/scraper_utils.py
ohinds: Totally! Good point.

recidiviz/ingest/scraper.py
```diff
@@ -296,6 +296,10 @@ def fetch_page(url, post_data=None, json_data=None):
                             "retry. \n\n%s", log_error)
             return -1
 
+        # Proxy errors don't raise and exception. : (
```
colincadams: nit: 'and' -> 'an'

recidiviz/ingest/us_ny/__init__.py
```diff
@@ -21,3 +21,5 @@
 
 
 from recidiviz.ingest.us_ny import us_ny_scraper
+from recidiviz.ingest.us_ny.us_ny_person import UsNyPerson
```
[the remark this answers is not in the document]
ohinds: Yes, but @colincadams and I talked about this. He's going to rip them all out at once when he does the Python 3 PR. Removing the entities breaks a bunch of docket and region tests because they all use the UsNyPerson and Record as the example entities for the test. And his stuff will remove those tests.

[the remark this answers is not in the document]
arian487: Gotcha, thanks!

recidiviz/ingest/scraper_utils.py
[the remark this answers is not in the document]
ohinds: We should probably make sure that part of archonix is tested :)

recidiviz/ingest/us_ny/us_ny_scraper.py
```diff
@@ -41,1226 +37,286 @@
...
+                    self._get_person_passthrough_params(content))
+            params_list.extend(self._get_next_page_params(content))
 
-        return person_details
+        # Add session variables to the post data of each params dict.
+        session_vars = self._get_session_vars(content)
 
-    def create_person(self, person_details):
-        """Creates a Person from the details scraped from the person page.
+        for one_params in params_list:
```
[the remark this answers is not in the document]
ohinds: That shadows a parameter to the function. But you are right, that's a terrible name.

recidiviz/ingest/us_ny/us_ny_scraper.py
```diff
@@ -41,1226 +37,286 @@
...
+            elif charge_class_str.startswith('C'):
+                charge_level = 'THREE'
+            elif charge_class_str.startswith('D'):
+                charge_level = 'FOUR'
+            elif charge_class_str.startswith('E'):
+                charge_level = 'FIVE'
+
+            # Get the degree
+            charge_degree = None
+            if charge_name.lower().endswith('1st'):
```
jamwalla: optionally - we could stick the last word of `charge_name` into `charge_degree` and add '1st', '2nd' etc. to the ChargeDegree map. Maybe also `charge_level`? This is more explicit, though, so up to you!

recidiviz/ingest/us_ny/us_ny_scraper.py
ohinds: Yes! I will file an issue.

ohinds: Filed #333.

recidiviz/ingest/us_ny/us_ny_scraper.py
ohinds: I like it.

recidiviz/ingest/scraper_utils.py
```diff
@@ -125,17 +125,23 @@ def calculate_age(birthdate, check_date=None):
         ((check_date.month, check_date.day) < (birthdate.month, birthdate.day))
 
 
-def get_id_value_from_html_tree(html_tree, html_id):
-    """Retrieves the value of the given id from the given html tree.
+def get_value_from_html_tree(html_tree, attribute_value, attribute_name='id'):
+    """Retrieves the value of an html tag from the given html tree. The
+    tag is chosen. If there are multiple matches, only the first is returned.
```
jessex: What do we mean by "The tag is chosen" exactly?

recidiviz/ingest/us_ny/us_ny.yaml
```diff
@@ -0,0 +1,15 @@
...
+  Inmate Name: person.surname
+  Sex: person.gender
+  Date of Birth: person.birthdate
+  Race / Ethnicity: person.race
+  Custody Status: booking.custody_status
+  Housing / Releasing Facility: booking.facility
+  Date Received (Current): booking.admission_date
+  Earliest Release Date: booking.projected_release_date
+
+  # TODO: string format is blocked on Python 3 migration, so uncomment
```
jessex: I'm confused. What's different about "Aggregate Minimum Sentence" and, say, "Custody Status"?

recidiviz/ingest/us_ny/us_ny_scraper.py
```diff
@@ -21,14 +21,10 @@
 Region-specific notes:
     - DOCCS includes historical data (back to ~1920s, reliably back to ~1970s)
     - DOCCS allows for surname-only search
-    - DOCCS returns the nearest match alphabetically, and then all subsequent
```
jessex: Why is this no longer relevant? (If this is revealed late in the code, sorry!)

recidiviz/ingest/us_ny/us_ny_scraper.py
```diff
@@ -41,1226 +37,276 @@
...
-            person_details: (dict) all of the scraped details to save
-
+        session_vars = {
+            UsNyScraper.TOKEN_NAME: scraper_utils.get_value_from_html_tree(
+                content, UsNyScraper.TOKEN_NAME, attribute_name='name'),
+        }
+
+        return session_vars
+
+    def _get_post_data(self, html_tree):
```
jessex: I think this is worth of its own docstring, too.

recidiviz/ingest/us_ny/us_ny_scraper.py
```diff
@@ -41,1226 +37,276 @@
...
-        This method does not actually save the updated entity to the database.
-        It is transient upon return.
-
-        Args:
-            person_details: (dict) all of the scraped details to save
+        return params_list
 
+    def _get_session_vars(self, content):
+        """Returns the session variables contained in a webpage.
+        Arguments:
```
jessex: Did I black out and miss our docstrings switching from `Args` to `Arguments`?

recidiviz/ingest/us_ny/us_ny_scraper.py
```diff
@@ -41,1226 +37,276 @@
...
+                charge_level = 'FOUR'
+            elif charge_class_str.startswith('E'):
+                charge_level = 'FIVE'
+
+            # Get the degree
+            charge_degree = charge_name.split()[-1]
+            try:
+                _ = ChargeDegree.from_str(charge_degree)
+                charge_name = ' '.join(charge_name.split()[:-1])
+            except KeyError:
```
jessex: How exceptional is this? I don't love use exceptions for non-exceptional control flow if it can be avoided.

recidiviz/ingest/us_ny/us_ny_scraper.py
```diff
@@ -41,1226 +37,276 @@
...
-            sex=record.sex,
-            surname=record.surname,
-            given_names=record.given_names)
-
-        return snapshot
+        data_extractor = DataExtractor(self.mapping_filepath)
+        ingest_info = data_extractor.extract_and_populate_data(content,
+                                                               ingest_info)
+
+        # Get release date, if released
```
jessex: Generally speaking, why can't we rely on the data extractor for this?

recidiviz/ingest/scraper_utils.py
ohinds: Typo, sorry.

recidiviz/ingest/us_ny/us_ny.yaml
ohinds: The format of the Aggregate Minimum Sentence field will be parsed by the dateparser we're going to use in Python 3 (now built in), but is unsupported by the data extractor currently.

recidiviz/ingest/us_ny/us_ny_scraper.py
```diff
@@ -41,1226 +37,276 @@
...
+        if ingest_info.person[0].booking[0].custody_status == 'RELEASED':
+            release_date_string = content.xpath(
```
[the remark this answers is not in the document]
ohinds: I added a check that there is exactly 1 element in the person list and that there is exactly one booking in the booking list on that person.

recidiviz/ingest/us_ny/us_ny_scraper.py
```diff
@@ -41,1226 +37,276 @@
...
+                '//*[@headers="t1l"]')[0].text.split()[0]
+            ingest_info.person[0].booking[0].release_date = release_date_string
+
+        # Parse charge information
```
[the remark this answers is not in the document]
ohinds: Because I need to extract the 'attempted' field and the 'degree' information from the charge name itself.

recidiviz/ingest/us_ny/us_ny_scraper.py
ohinds: This is, as of now, the @zacharylawrence approved method for checking whether a value is in the map and will be mapped properly. We talked about adding a 'check_valid' method to constant classes if this pattern has to repeat often.

recidiviz/ingest/us_ny/us_ny_scraper.py
ohinds: Because the release reason / type is embedded in the same text field, so needs to be split out manually.

recidiviz/ingest/us_ny/us_ny_scraper.py
ohinds: Yes, I've changed the search to just start at 'a', which is alphabetically before 'aaardvark' anyhow. I'm sad to see the aaardvark go, though.

recidiviz/ingest/us_ny/us_ny_scraper.py
zacharylawrence: Yeah, this is a good point @jessex. What alternatives would you prefer? Should `ChargeDegree.from_str()` just return `None` in that case? Forcing users to specifically check if a valid value is returned?

I think it's reasonable consider a failed `from_str()` to throw an exception but we should probably be using our own exception type instead of KeyError?

One case where `from_str()` failing with an exception is nice is during the data_conversion step where failure to parse a string into an enum means that exception should bubble up and prevent us from writing to the database.

recidiviz/ingest/base_scraper.py
```diff
@@ -108,24 +108,34 @@ def _generic_scrape(self, params):
...
 
-        # Let the child transform the post_data if it wants before sending the
-        # requests.  This hook is in here in case the child did something like
-        # compress the post_data before it put it on the queue.
-        post_data = self.transform_post_data(post_data)
+        # Here we handle a special case where we weren't really sure
+        # we were going to get data when we submitted a task, but then
+        # we ended up with data, so no more requests are required,
+        # just the content we already have.
+        if endpoint is None:
```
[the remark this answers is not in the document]
colincadams: This raises a bigger question in my mind around the `task_type` concept, outside of this PR but wanted to raise it here. It seems that `task_type` can't always describe the desired behavior, what if instead we allowed the child scraper to define it's own task / page types that described it's page structure better and generic scrape just always called `get_more_tasks` and `populate_data` on it? Then the child would just return an empty list from `get_more_tasks` if it is a leaf node, and could return data from it if it hit a leaf node earlier than expected?

I don't think this would add much overhead for the simplest scrapers, and it would help with ones like `us_pa` where right now we have to do extra checks on `endpoint` to figure out where in the page tree we are. I think it also removes the burden of knowing how to configure the behavior of generic scrape, it would just always fetch content and always allow you to return more tasks or data.

What do you think? I can also file an issue and move the discussion there.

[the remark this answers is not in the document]
ohinds: Yes, that's a viable alternative that would simplify some situations. Definitely file the issue and we can talk there.

recidiviz/ingest/us_ny/us_ny_scraper.py
ohinds: I think this last point is important. I like the idea of throwing our own exception to bubble up instead of writing to the DB but silently nulling fields that don't conform. Filed #361.

PR #323, opened 2018-12-14
Migrated push queues from TaskQueues to Cloud Tasks.

rasmi:


requirements.txt
```diff
@@ -42,24 +50,29 @@ mock==2.0.0
 modulegraph==0.10.4
 more-itertools==4.1.0
 nose==1.3.7
-PyJWT==1.7.0
```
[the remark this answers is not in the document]
colincadams: Nevermind it's just lowercase now :)

rasmi:
Agreed, I did feel some redundancy in re-creating request and task objects. I think at the very least a queues.py module with `create_task` and `purge_queue` methods would be neater.

`create_task` would accept a request_body dict, a scraper_work_url, and a queue name, which would abstract away enough functionality to make it similar to the old `taskqueues.add`:
```
queues.create_task(
    url=self.scraper_work_url,
    queue_name=self.get_region().queue,
    params={
        'region': self.get_region().region_code,
        'task': task_name,
        'params': task_params
})
```

And then we can mock out `queues.create_task` and `queues.purge_queue` rather than the Cloud Tasks API itself, and the Cloud Tasks client will be created in `queues.py` rather than in the scraper. I can also name it `pushqueues.py` rather than `queues.py` to disambiguate from the docket if you think that's helpful. Is that the sort of decoupling you had in mind?

rasmi:
@colincadams -- integrated your recommendations, thanks!

recidiviz/ingest/queues_test.py
```diff
@@ -0,0 +1,71 @@
...
+import unittest
+from mock import patch
+
+from google.cloud import tasks_v2beta3
+
+from recidiviz.ingest import queues
+from recidiviz.utils import metadata
+
+
+class QueuesTest(unittest.TestCase):
```
jessex: This should be packaged under `recidiviz/tests/ingest` instead of `recidiviz/ingest`, yeah?

recidiviz/ingest/queues_test.py
```diff
@@ -0,0 +1,71 @@
...
+    """Cloud Tasks queue helper function tests."""
+
+
+    @patch('google.cloud.tasks_v2beta3.CloudTasksClient.create_task')
+    def testCreateTask(self, mock_create_task):
```
colincadams: nit: I believe all of our other test methods are formatted lowercase with underscores: `test_create_task`.

rasmi: I love a good nit, thanks!

recidiviz/ingest/queues_test.py
rasmi: Good catch, thanks!

rasmi:
I've integrated your changes, thank you both for your comments and guidance throughout this PR!

PR #325, opened 2018-12-15
Small archonix fixes

arian487:


recidiviz/ingest/base_scraper.py
```diff
@@ -118,7 +118,7 @@ def _generic_scrape(self, params):
         # Let the child transform the post_data if it wants before sending the
         # requests.  This hook is in here in case the child did something like
         # compress the post_data before it put it on the queue.
-        post_data = self.transform_post_data(post_data)
+        self.transform_post_data(post_data)
```
jessex: We're in a weird spot. In `base_scraper`, this returns the input straight-away, a no-op. But in `archonix_scraper`, the overridden implementation modifies the given argument. The parent and all children should have the same return type, or no return type at all. But they shouldn't differ.

arian487: Good point.  I've changed the base_scraper to a pass as I think the behaviour should be a transform of the given object.

recidiviz/ingest/base_scraper.py
```diff
@@ -245,7 +245,7 @@ def transform_post_data(self, data):
         Args:
             data: dict of parameters to send as data to the post request.
         """
-        return data
+        pass
```
jessex: I know @zacharylawrence explicitly changed this from a `pass` to a `return data`, so I want to let him comment before approving.

arian487: Yea I'll ask him on Monday but I don't think its too pivotal either way right now since archonix is the only one that needs to use transform_data

zacharylawrence: First, I agree this doesn't seem pivotal and I'm fine moving forward with any approach.

My preference for returning the transformed data is that it lets us avoid using arguments as the outputs of functions (ie. I'm the most used to reasoning about functions that takes inputs in and return an output).

The book [Clean Code (pg 45)](https://www.investigatii.md/uploads/resurse/Clean_Code.pdf) expands on this idea further and says:

> Arguments are most naturally interpreted as inputs to a function.

> In general output arguments should be avoided. If your function must change the state of something, have it change the state of its owning object. 

These Stack Overflow answers also have recommendations on the topic:
- https://softwareengineering.stackexchange.com/questions/245767/is-modifying-an-incoming-parameter-an-antipattern
- https://softwareengineering.stackexchange.com/questions/208828/passing-an-object-into-a-method-which-changes-the-object-is-it-a-common-anti

However, I think it's also important to note that if we plan to have the `transform_post_data` method mutate the passed in `data` argument then we shouldn't also return it unless we make a defensive copies of the mutable `data` argument before passing it into the `transform_post_data` method.

If we decide that defensive copies aren't worth the extra work (unless we make the `data` argument immutable - seems unlikely!), I'm fine with just mutating the input parameter directly as long as we're clear in our pydocs that that's what we're doing. It's also worth noting that this seems like the recommended approach in python, as explained [here](https://stackoverflow.com/questions/26027694/correct-style-for-python-functions-that-mutate-the-argument):

> Generally, in python, we expect a function to either mutate the arguments, or return something. The reason for this is because if a function doesn't return anything, then it makes it abundantly clear that the function must have had some side-effect in order to justify it's existence (e.g. mutating the inputs).

arian487: Discussed offline, we'll leave it as is for now.

jessex: For the record, I also don't like side-effects and agree with @zacharylawrence's excellent thinking in this comment.

PR #326, opened 2018-12-15
Add the missing libs

arian487:
This PR adds all of the libs that were missing for appengine.


arian487:
Yea this change is only adding the missing libs.  No real code changes.  Thanks for the late night approval!

PR #327, opened 2018-12-16
Added us_fl_martin_county scraper

HalfdanJ:
Here is my first scraper. Its a very basic scraper since all current inmates are listed on one page.


recidiviz/ingest/us_fl_martin_county/us_fl_martin_county_scraper.py
```diff
@@ -0,0 +1,133 @@
...
+class UsFlMartinCountyScraper(BaseScraper):
+    """Scraper implementation for us_fl_martin_county."""
+
+    def __init__(self, mapping_filepath=None):
+        super(UsFlMartinCountyScraper, self).__init__('us_fl_martin_county')
+
+    def set_initial_vars(self, content, params):
+        pass
+
+    def populate_data(self, content, params, ingest_info):
```
arian487: This is probably where we need better documentation for our tools.  I believe this case is handled entirely by our data extractor.  Can you try using it?  It should change your populate data length to only a few lines.  In short the data extractor automagically extracts the data based on a yaml file you write so you don't need to figure out the html makeup of a site.

Look at us_pa_greene for an example on how to use it.

recidiviz/ingest/us_fl_martin_county/us_fl_martin_county_scraper.py
```diff
@@ -0,0 +1,133 @@
...
+                arrest = booking.create_arrest()
+
+                for i in range(len(table[0])):
+                    key = table[0][i].text_content()
+                    value = table[1][i].text_content().rstrip()
+                    if not isinstance(value, str):
+                        continue
+
+                    # Unsure what Cfn stands for
+                    # if table[0][i].text_content().startswith('Cfn'):
```
arian487: Lets make this person_id

recidiviz/ingest/us_fl_martin_county/us_fl_martin_county_scraper.py
```diff
@@ -0,0 +1,133 @@
...
+                        if key.startswith('Bond'):
+                            charge.create_bond(amount=value)
+
+                        if key.startswith('Charge'):
+                            charge.name = value
+
+        return ingest_info
+
+    def get_more_tasks(self, content, params):
+        task_type = params.get('task_type', self.get_initial_task_type())
```
arian487: This will only be called when the scrape type is get_more_tasks.  Since that is only over the first call in your case, I think its unncessary to wrap it around an if asking if we are in initial task, that'll let you remove the return [] at the end./

arian487:
The data extractor should work for both rows and column data.  What was the error I'll take a look.

HalfdanJ:
if I do this

```
    def populate_data(self, content, params, ingest_info):
        data_extractor = DataExtractor(self.mapping_filepath)
        ingest_info = data_extractor.extract_and_populate_data(content, ingest_info)
        print(ingest_info)
```

with a mapping file like this
```
key_mappings:
  "Last Name": person.surname
  "First Name": person.given_names
  "Race": person.race
```

I get this:

```
IngestInfo:
   person[0]: _Person:
      surname: Suffix
      given_names: Middle Name
      race: Sex
      booking[0]: _Booking:
   person[1]: _Person:
      surname: Suffix
      given_names: Middle Name
      race: Sex
```

So its looking in the next cell instead of next row. Its also not noticing all the people on the site.

arian487:
Yea definitely need better documentation :)

The extractor only works if you have a complete yaml file.  In short the reason it thinks suffix is a value, is because it doesn't exist in your keys.  Try mapping everything properly and it should work.  (Make sure to make Cfn to person.person_id)

HalfdanJ:
I pushed a change that uses dataextractor. It seems like there is a bug:

```
            get_recent_name = 'get_recent_' + class_to_set
>           object_to_set = getattr(parent, get_recent_name)()
E           AttributeError: 'NoneType' object has no attribute 'get_recent_arrest'
```
Second, how do I concatenate two fields to one key? (First name + middle name)

HalfdanJ:
Also, it adds all charges to the first person

arian487:
I know why the first error is happening I think.  I'll investigate why all charges are added to the first person.  Will try and get to that today.

arian487:
Alright, so I've sent out a small fix for the data_extractor here:  https://github.com/Recidiviz/recidiviz-data/pull/328

Second, the all charges being added to the first person seems to be a bug related to our inline cell conversion.  To make your test pass for now comment out line 98-99 in the data_extractor, I'll try and figure out the underlying issue and send out a fix on Monday.  Doing that, plus my PR I linked will result in the right structure.  Note your test had some issues being different from the report.html so I had to fix that up too, but that is unrelated to the extractor.  Here is the yaml I used:

key_mappings:
  Cfn: person.person_id
  Last Name: person.surname
  First Name: person.given_names
  Race: person.race
  Sex: person.gender
  Date of Birth: person.birthdate
  Arresting Agency: arrest.agency
  Arrest Date: arrest.date
  
keys_to_ignore:
  - Suffix
  - Arrest Time
  - Middle Name
  - Inmate Photo

multi_key_mapping:
  Charge Type: charge.charge_class
  Statute: charge.statute
  Charge: charge.name
  Bond: bond.amount

Thanks again!

arian487:
The scraper writers are not supposed to know or care about the converter.  We think it makes most sense for a scraper writer to write assertions on what they see on the website and not our internal conformed standards.  That requires you to go and look at our internal maps (trivial for race, but harder for other things).  So yes, in tests just verify the actual strings in the HTML, not the converted formats.

HalfdanJ:
Fixes work, and PR is updated. Should be ready to merge i believe.

recidiviz/tests/ingest/us_fl_martin_county/fixtures/report.html
```diff
@@ -0,0 +1,124 @@
...
+                <td class="Data">&nbsp;</td>
+                <td class="Data">ARTHUR</td>
+                <td class="Data">A</td>
+                <td class="Data">B</td>
+                <td class="Data">M</td>
+                <td class="Data">01/01/1994</td>
+                <td class="Data">AAA</td>
+                <td class="Data">01/01/2018</td>
+                <td class="Data">10:00:00</td>
+                <td class="Image"><img src="images/inmates/1111111.JPG" width="88" height="60" border="1"></td>
```
jessex: That's not a real path to an image, right?

recidiviz/ingest/us_fl_martin_county/us_fl_martin_county.yaml
```diff
@@ -0,0 +1,21 @@
+key_mappings:
```
jessex: Curious: were there any fields on the page that there was no mapping for? Are we leaving anything behind?

arian487:
Looks good, please fix the pylint issues though before merging.

HalfdanJ:
Pr should be ready

arian487:
Awesome work, I'll merge this in!

PR #335, opened 2018-12-17
Refactor docket to use pubsub

colincadams:
This also removes snapshot scrape logic from the docket. It uses ndb models and needs to be re-written any way, and it was simpler to just remove it for now.


recidiviz/ingest/docket.py
```diff
@@ -65,24 +64,69 @@
...
+
+def _docket_subscription_name(scrape_key):
+    return "v1.{}-{}".format(scrape_key.region_code, scrape_key.scrape_type)
+
+def _get_topic_and_subscription(scrape_key):
+    topic_path = publisher().topic_path(
+        metadata.project_id(), _docket_topic_name(scrape_key))
+    try:
+        publisher().get_topic(topic_path)
+    except exceptions.NotFound:
```
jessex: Is this the recommended flow for get-or-create of new topics? I don't love using exceptions as a means of non-exceptional control flow.

recidiviz/ingest/docket.py
```diff
@@ -173,7 +212,7 @@ def load_background_target_list(region_code, name_file, query_name, load,
             elif start_line <= row_num < max_row:
                 names.append(list(row))
             else:
-                next_line = row_num
+                # next_line = row_num
```
jessex: Do we need this?

recidiviz/ingest/docket.py
```diff
@@ -197,6 +236,7 @@ def load_background_target_list(region_code, name_file, query_name, load,
         add_to_query_docket(ScrapeKey(region_code, scrape_type), names)
 
     if not end_of_list:
+        pass
```
jessex: Why are we keeping the code inside of the same conditional immediately after a `pass`?

recidiviz/ingest/docket.py
```diff
@@ -615,53 +404,24 @@ def purge_query_docket(scrape_key):
...
-    q = taskqueue.Queue(DOCKET_QUEUE_NAME)
+    (_, subscription_path) = _get_topic_and_subscription(scrape_key)
 
-    empty = False
-    while not empty:
-        docket_items = q.lease_tasks_by_tag(lease_seconds,
-                                            max_tasks,
-                                            tag_name,
-                                            deadline_seconds)
+    # TODO(colinadams): Use subscriber().seek(subscription_path, time=timestamp)
```
jessex: Should we file an issue for this?

recidiviz/ingest/docket.py
```diff
@@ -615,53 +404,24 @@ def purge_query_docket(scrape_key):
...
 
-        if docket_items:
-            q.delete_tasks(docket_items)
-        else:
-            empty = True
 
-
-def delete_docket_item(item_name):
+def ack_docket_item(scrape_key, ack_id):
     """Delete a specific docket item from the docket pull queue.
```
jessex: We should update this doc string in at least a couple of places.

recidiviz/ingest/docket.py
```diff
@@ -615,53 +404,24 @@ def purge_query_docket(scrape_key):
...
 
     Args:
-        item_name: (string) The docket item's name, which we use to select and
-            delete it
+        ack_id: (string) Id used to ack the message
 
     Returns:
-        True if successful
-        False if not
+        N/A
```
jessex: I've noticed we're inconsistent about whether our docstrings include a returns clause or not for "void" functions. I just filed #341 for this, which calls for removing returns clauses entirely for these functions.

recidiviz/ingest/tracker.py
```diff
@@ -36,23 +36,24 @@ def iterate_docket_item(scrape_key, back_off=5):
...
+        scrape_key, return_immediately=return_immediately)
 
     if not docket_item:
         logging.info("No items in docket for %s. Ending scrape.", scrape_key)
         return None
 
-    item_content = json.loads(docket_item.payload)
-
-    item_added = sessions.add_docket_item_to_current_session(docket_item.name,
+    item_content = json.loads(docket_item.message.data.decode('utf-8'))
```
jessex: The docket item message could contain names from our own database in the future case of snapshotting, right? I'm wondering if we have any safeguards in place to ensure we don't write non-utf-8 characters into our database? Not really pertinent to this PR, but an idle question.

recidiviz/ingest/docket.py
colincadams: (Un?)surprisingly I couldn't find a recommended flow for this in the docs, although I would've expected it be on [this page](https://cloud.google.com/pubsub/docs/admin). Here are the methods for [publisher]() and [subscriber]() clients. From my read, the easiest way to avoid exceptions would be to use the `list_*` methods but in the case where we have thousands of topics and subscriptions that seems quite sub-optimal, so I opted for this get and handle exception flow.

It's definitely not optimal, I wish that the api would just return none or an optional that we could explicitly check, instead of raising an exception. Totally open to other ideas you have to handle this.

recidiviz/ingest/docket.py
colincadams: Sorry about this, this was leftover test code that I missed in my commit 😳

recidiviz/ingest/docket.py
colincadams: Same here 😳

recidiviz/ingest/docket.py
colincadams: Filed #342

recidiviz/ingest/docket.py
colincadams: Updated this docstring as well as some others

recidiviz/ingest/docket.py
colincadams: ack

recidiviz/ingest/tracker.py
colincadams: I filed #343, we should definitely look into this at some point.

recidiviz/ingest/docket.py
jessex: Yeah I don't see a better alternative in the API options, at least not in the clients themselves. Returning an optional or none would be delightful, and a simple `does_topic_exist` check would be nice, too. But I think you've got the best approach here given what's afforded us.

recidiviz/tests/ingest/docket_test.py
```diff
@@ -19,248 +19,89 @@
...
-                       method='PULL').add(docket.DOCKET_QUEUE_NAME)
+        docket.add_to_query_docket(scrape_key_add, get_payload())
 
-        docket.purge_query_docket(scrape_key)
-        assert not docket.get_new_docket_item(scrape_key, back_off=1)
+        docket.purge_query_docket(scrape_key_purge)
+        assert not docket.get_new_docket_item(scrape_key_purge,
+                                              return_immediately=True)
+
+        # Cleanup
```
jessex: Should this go in a teardown method instead?

recidiviz/tests/ingest/tracker_test.py
```diff
@@ -19,141 +19,164 @@
...
+        sessions.end_session(scrape_key)
+
+    def test_iterate_docket_item_no_open_session_to_update(self):
+        scrape_key = ScrapeKey("us_ny", constants.BACKGROUND_SCRAPE)
+
+        docket.add_to_query_docket(scrape_key, get_payload())
+
+        assert not tracker.iterate_docket_item(scrape_key)
+
+        docket.purge_query_docket(scrape_key)
```
jessex: Similar question: should these calls go in a teardown method?

jessex: Similar question: should put this in a teardown? Wouldn't hurt to apply it to all methods in this class equally.

colincadams: I moved this to teardown and also added removing all of the sessions

recidiviz/tests/ingest/docket_test.py
colincadams: Done

PR #337, opened 2018-12-17
A few small changes for prod to go e2e

arian487:
Just a few changes left for prod.


arian487:
Note the changes in converter are not permanent, this is just in there to unblock prod I have opened an issue to better handle this: https://github.com/Recidiviz/recidiviz-data/issues/338

recidiviz/persistence/converter.py
```diff
@@ -151,6 +151,9 @@ def _convert_booking(self, booking_id):
         if ingest_booking.HasField('custody_status'):
             booking.custody_status = self._convert(
                 CustodyStatus.from_str, ingest_booking.custody_status)
+        else:
```
zdg2102: Can you mark these fallbacks with a TODO so it's easy for someone else coming in after to see that they're only intended as temporary workarounds?

arian487: Done.

recidiviz/__init__.py
```diff
@@ -34,7 +34,7 @@
     db_name = secrets.get_secret('sqlalchemy_db_name')
 
     sqlalchemy_url = \
-        'postgresql://{db_user}:{db_password}@{db_host}/{db_name}'.format(
+        'mysql+pymysql://{db_user}:{db_password}@{db_host}/{db_name}'.format(
```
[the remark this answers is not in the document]
zdg2102: It's the format SQLAlchemy specifies for passing URLs to configure its internal engine: https://docs.sqlalchemy.org/en/latest/core/engines.html

arian487:
Yup, definitely high priority :)

PR #340, opened 2018-12-17
Add more libraries needed for mysql calls

arian487:
More libs needed for pymysql.  This is somewhat temporary as python3 migration will let us use postgres.


arian487:
I know I hate lib/ too...

PR #344, opened 2018-12-18
First take at entity matching

terinpw:
* Create db query to find people in a region with open bookings
* Create basic entity equality methods for all entity types in entity_matching.py
* Defined basic entity matching flow in entity_matching.match_entities 
* Added region to entities.Person (already in schema.Person)

Caveats:
- Assumes that the ingested entities passed into entity matching do not have inferred admission dates set. If the inferred dates are populated on the ingested entities some logic will need to change. 
- Only matches on People and Bookings for now (everything else in a follow up PR), but this means that old charges will be orphaned when a booking is re-scraped. 

Co-Authored-By: jamwalla


recidiviz/persistence/database/database.py
```diff
@@ -64,9 +78,20 @@ def read_open_bookings_scraped_before_time(session, region, time):
...
-        .filter(Person.region == region)\
-        .filter(Booking.release_date.is_(None))\
+    query = _query_people_and_open_bookings(session, region)\
         .filter(Booking.last_seen_time < time)
     return [booking for _, booking in query.all()]
+
+
+def _query_people_and_open_bookings(session, region):
+    """
+    Returns a tuple of (person, booking) for all people with open bookings
```
jessex: This might be _slightly_ ambiguous in that people could infer a return value other than a list, e.g. a generator of tuples or a dictionary with tuples as values. I might make that explicit with "Returns a list of tuples of ..."

recidiviz/persistence/entity_matching.py
```diff
@@ -24,19 +24,148 @@ class EntityMatchingError(Exception):
     pass
 
 
-def get_entity_match(session, person):
+# TODO: in Python 3, make these fields named and required
```
jessex: Let's add a reference to #176 on this TODO.

recidiviz/persistence/entities.py
```diff
@@ -34,6 +34,7 @@ class Person(object):
     birthdate_inferred_from_age = attr.ib(default=None)
     gender = attr.ib(default=None)
     race = attr.ib(default=None)
+    region = attr.ib(default=None)
```
jessex: I just noticed further up in this class that the top-level docstring still mentions overriding `__eq__` but that no longer happens here. Would you mind removing that reference?

recidiviz/persistence/entity_matching.py
```diff
@@ -24,19 +24,148 @@ class EntityMatchingError(Exception):
...
+def is_person_match(db_person=None, ingested_person=None):
     """
-    Queries the database to find all matching entities for the given Person.
+    Given a database person and an ingested person, determine if they should be
```
jessex: I think that since the number of ways we could attempt entity matching are variable, it would be a good idea to briefly document the current state of each entity matching algorithm in the method docstrings. What do you think?

recidiviz/persistence/entity_matching.py
```diff
@@ -24,19 +24,148 @@ class EntityMatchingError(Exception):
...
+    considered the same person.
+    Args:
+        db_person: (entities.Person)
+        ingested_person: (entities.Person)
+    """
```
jessex: I would add a `Returns:` clause in here that makes explicit we are returning a boolean. Same for all functions in here.

recidiviz/persistence/entity_matching.py
```diff
@@ -24,19 +24,148 @@ class EntityMatchingError(Exception):
...
+    # TODO: we assume that dates are inferred after entity matching. If they're
+    # inferred before we can simplify the following two cases.
+
+    # If the db booking's admission date was scraped (not inferred), the
+    # ingested booking must have the same admission date to be a match.
+    if not db_booking.admission_date_inferred:
+        return db_booking.admission_date == ingested_booking.admission_date
+
+    # If the db booking's admission date was inferred and then ingested
+    # booking's admission date was scraped, this is not a match.
```
jessex: I think I follow the logic here but I'm curious under what cases this would happen. A given source data system will typically always have inferred dates or always have scraped dates, right? Is this because booking matching can be cross-jurisdiction?

recidiviz/persistence/entity_matching.py
```diff
@@ -24,19 +24,148 @@ class EntityMatchingError(Exception):
...
+
+    return db_person.given_names == ingested_person.given_names and \
+        db_person.surname == ingested_person.surname and \
+        db_person.birthdate == ingested_person.birthdate
+
+
+def is_booking_match(db_booking=None, ingested_booking=None):
+    """
+    Given a database booking and an ingested booking, determine if they should
+    be considered the same booking. Does not compare on person_id.
```
jessex: Why don't we compare on person id? I'm worried that two obviously people admitted on the same date in a system without external booking ids and without inferred admission dates will have their bookings merged.

recidiviz/persistence/entity_matching.py
```diff
@@ -24,19 +24,148 @@ class EntityMatchingError(Exception):
...
+        return False
+
+    # return True if they are both active
+    return not db_booking.release_date and not ingested_booking.release_date
+
+
+def is_charge_match(db_charge=None, ingested_charge=None):
+    """
+    Given a database charge and an ingested charge, determine if they should be
+    considered the same charge. Does not compare on booking_id.
```
jessex: I have a similar question here as for the "person_id" statement above. Is there a safeguard in here I'm not aware of to ensure we don't merge entities that are clearly in different object graphs, or am I maybe misunderstanding how these methods are used?

recidiviz/persistence/entity_matching.py
```diff
@@ -24,19 +24,148 @@ class EntityMatchingError(Exception):
     pass
 
 
-def get_entity_match(session, person):
+# TODO: in Python 3, make these fields named and required
+def is_person_match(db_person=None, ingested_person=None):
```
jessex: Do these need to be optional args? We assert on the presence of the args as the first line in each function, so it seems like a signature of `is_person_match(db_person, ingested_person)` would be safer.

recidiviz/persistence/entity_matching.py
```diff
@@ -24,19 +24,148 @@ class EntityMatchingError(Exception):
...
+    Args:
+        session: (Session)
+        region: (str)
+        ingested_people: List[entities.Person]
+    """
+    db_people = database.read_people_with_open_bookings(session, region)
+
+    for ingested_person in ingested_people:
+        for db_person in db_people:
+            if is_person_match(ingested_person=ingested_person,
```
jessex: If you do follow my suggestion of switching from optional args to normal args, remember to switch the order here.

recidiviz/persistence/entity_matching.py
```diff
@@ -24,19 +24,148 @@ class EntityMatchingError(Exception):
...
+                               db_person=db_person):
+                if ingested_person.person_id:
+                    raise EntityMatchingError('matched ingested person to more'
```
jessex: I'm curious what we would do with this error. At first thought, it seems unrecoverable without manual updates in the database?

recidiviz/persistence/entity_matching.py
```diff
@@ -24,19 +24,148 @@ class EntityMatchingError(Exception):
...
+        for db_booking in db_person.bookings:
+            if is_booking_match(db_booking=db_booking,
+                                ingested_booking=ingested_booking):
+                if ingested_booking.booking_id:
+                    raise EntityMatchingError('matched ingested booking to'
+                                              'more than one database entity')
+                ingested_booking.booking_id = db_booking.booking_id
 
-    return existing_people[0] if existing_people else None
+                # TODO: logic will need to change if inferring happens before
```
jessex: I see a few TODOs related to this concept. Do we need to file an issue for this or are you all planning to figure this out before merging?

recidiviz/tests/persistence/entity_matching_test.py
```diff
@@ -0,0 +1,166 @@
...
+
+    def test_booking_match_open_bookings(self):
+        db_booking = Booking(admission_date=_DATE,
+                             admission_date_inferred=True)
+        ingested_booking = Booking()
+        self.assertTrue(is_booking_match(db_booking, ingested_booking))
+        ingested_booking.release_date = _DATE
+        self.assertFalse(is_booking_match(db_booking, ingested_booking))
+
+    def test_charge_match(self):
```
jessex: I think it's worthwhile adding a TODO here mentioning that once we've figured out which fields should not change on these entities, the test cases need to be updated/extended.

recidiviz/persistence/database/database.py
```diff
@@ -51,6 +51,20 @@ def read_bookings(session):
     return session.query(Booking).all()
 
 
+def read_people_with_open_bookings(session, region):
```
zdg2102: I know this is in line with the other functions in this file, but this naming convention is strange to me. "Fetch" or "get" seem more appropriate than "read".

recidiviz/persistence/database/database.py
zdg2102: Can we rephrase this as "for all open bookings"? Since really it's the booking that acts as the filter here, and the person is just an associated object.

recidiviz/persistence/entity_matching.py
zdg2102: That seems like it would be easy to get out of sync. I'm generally a big fan of extensive documentation, but this seems to me like one of those cases where in the end you'll have to read the code itself to determine exactly how matching is determined.

zdg2102: "Database person" and "ingested person" isn't technically a constraint, is it? I'd prefer keeping this naming more generic, something like:

def are_matching_people(person_a, person_b):
    """Returns a bool indicating whether two person entities represent the same person."""

Same for the other functions throughout.

recidiviz/persistence/entity_matching.py
```diff
@@ -24,19 +24,148 @@ class EntityMatchingError(Exception):
...
-    Queries the database to find all matching entities for the given Person.
+    Given a database person and an ingested person, determine if they should be
+    considered the same person.
+    Args:
+        db_person: (entities.Person)
+        ingested_person: (entities.Person)
+    """
+    assert db_person and ingested_person
+
+    if db_person.region != ingested_person.region:
```
zdg2102: Will we ever be doing comparisons between two people who aren't in the same region?

recidiviz/persistence/entity_matching.py
zdg2102: I thought the assumption was that comparisons will be done by walking down the tree? I.e. that we will never even end up comparing bookings if they don't belong to the same person. Although if that's the case, we should definitely update the docstring to make that clear.

recidiviz/persistence/entity_matching.py
zdg2102: Also if we switch the arguments to not distinguish ingested vs. DB, the order won't matter.

recidiviz/tests/persistence/entity_matching_test.py
```diff
@@ -0,0 +1,166 @@
...
+from datetime import datetime
+from unittest import TestCase
+import attr
+
+from recidiviz import Session
+from recidiviz.common.constants.booking import CustodyStatus
+from recidiviz.persistence.database import schema, database_utils
+from recidiviz.persistence.entities import (Bond, Booking, Charge, Person,
+                                            Sentence)
+from recidiviz.persistence.entity_matching import (is_bond_match,
```
zdg2102: Nit: can we import this with an "as", so the functions are prefixed instead of being free-floating?

recidiviz/tests/persistence/entity_matching_test.py
```diff
@@ -0,0 +1,166 @@
...
+_DATE = datetime(2018, 12, 13)
+_DATE_ANOTHER = datetime(2017, 12, 13)
+
+
+class TestEntityMatching(TestCase):
+    """Tests for entity matching logic"""
+
+    def setup_method(self, _test_method):
+        fakes.use_in_memory_sqlite_database()
+
```
zdg2102: Nit: double empty lines

recidiviz/persistence/entity_matching.py
```diff
@@ -24,19 +24,148 @@ class EntityMatchingError(Exception):
...
+
+def is_charge_match(db_charge=None, ingested_charge=None):
+    """
+    Given a database charge and an ingested charge, determine if they should be
+    considered the same charge. Does not compare on booking_id.
+    Args:
+        db_charge: (entities.Charge)
+        ingested_charge: (entities.Charge)
+    """
+    assert db_charge and ingested_charge
```
arian487: Same assertion comment.

recidiviz/persistence/entity_matching.py
```diff
@@ -24,19 +24,148 @@ class EntityMatchingError(Exception):
...
+def is_bond_match(db_bond=None, ingested_bond=None):
+    """
+    Given a database bond and an ingested bond, determine if they should be
+    considered the same bond. Does not compare on charge_id.
+    Args:
+        db_bond: (entities.Bond)
+        ingested_bond: (entities.Bond)
+    """
+    assert db_bond and ingested_bond
+    # TODO: decide which fields cannot change
```
arian487: Can you open an issue on deciding which fields can and cannot change?

recidiviz/persistence/persistence.py
```diff
@@ -106,23 +106,17 @@ def write(ingest_info, region, last_seen_time):
...
-        except Exception:
-            session.rollback()
-            raise
-        finally:
-            session.close()
+    session = Session()
+    try:
+        entity_matching.match_entities(session, region, people)
+        for person in people:
+            session.merge(database_utils.convert_person(person))
```
arian487: Just to be clear, if there is no match, then this merge call won't fail correct.  In that case the actual id will be none and the database will need to know to put a new row.

recidiviz/tests/persistence/entity_matching_test.py
```diff
@@ -0,0 +1,166 @@
...
+    def test_person_mismatch_region(self):
+        db_person = Person(external_id=_EXTERNAL_ID, region=_REGION)
+        ingested_person = Person(external_id=_EXTERNAL_ID, region=_REGION)
+        self.assertTrue(is_person_match(db_person, ingested_person))
+        ingested_person.region = _REGION_ANOTHER
+        self.assertFalse(is_person_match(db_person, ingested_person))
+
+    def test_person_match_birthdate(self):
+        db_person = Person(birthdate=_DATE)
```
arian487: Wait...why does this pass?  The only field in the object is birthdate...

recidiviz/persistence/entity_matching.py
terinpw: Chatted with @jamwalla, and we tend to agree that it should be in the docstring, or at least in a readme in persistence/. I think since the entity matching isn't finalized (this is a first take and we're going to send out a doc for everyone this week) we're going to add a TODO to add this later. Does that sound good?

recidiviz/persistence/entity_matching.py
terinpw: I don't think we're building for cross-jurisdiction matching. Both this if statement and the one above it are only really useful if a website doesn't always have admission_date filled in. 

I think you're right though that this case can go away, because if admission date wasn't filled in on a previous day, and then is today, then we should still match the bookings if they're both open.

recidiviz/persistence/entity_matching.py
terinpw: Added issue #348

recidiviz/persistence/database/database.py
terinpw: Do you feel ok if we do this in a follow up PR? I think we should update all of the database methods to do so.

recidiviz/persistence/entity_matching.py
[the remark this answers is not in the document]
terinpw: Chatted offline, but db_person is an entity type, not a schema type. Per @jessex 's comment, we're making all of the arguments required.

recidiviz/persistence/entity_matching.py
zdg2102: sgtm

recidiviz/persistence/database/database.py
zdg2102: Yep, works for me

recidiviz/persistence/entity_matching.py
terinpw: +1 to @zdg2102 's comment. Updated docstring as well to make this clearer.

recidiviz/persistence/entity_matching.py
terinpw: Same as comment above, let us know if you still have concerns.

recidiviz/persistence/entity_matching.py
terinpw: Chatted w @jamwalla, if a person or booking gets matched multiple times this does seem like a catastrophic error that someone should have to manually look at. 

However, for charges/bonds/sentences it's totally valid to have the same data for multiple entities, so we need a better way of handling this (opened #349 to track)

recidiviz/persistence/entity_matching.py
terinpw: Chatted offline, used entity_a/entity_b naming for all methods but matching on booking (where it actually does matter which one is from the db and which is ingested).

recidiviz/persistence/entity_matching.py
terinpw: Nope, removed

recidiviz/tests/persistence/entity_matching_test.py
terinpw: I think we want single lines for methods within a class

recidiviz/persistence/entity_matching.py
```diff
@@ -24,19 +24,148 @@ class EntityMatchingError(Exception):
...
+        return False
+
+    if db_person.external_id or ingested_person.external_id:
+        return db_person.external_id == ingested_person.external_id
+
+    # TODO: consider not matching if birthdate_inferred_from_age
+    # TODO: if matching on inferred birthdate, inferred birthdates will change
+    # as people age in custody.
+
+    return db_person.given_names == ingested_person.given_names and \
```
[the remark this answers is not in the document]
terinpw: Updated

recidiviz/persistence/entity_matching.py
terinpw: done in #350

recidiviz/persistence/persistence.py
terinpw: Yup, that's the functionality.

recidiviz/persistence/entity_matching.py
jessex: That sounds good. I think Zach's right that it could easily get out of sync and may not be suitable for a method-level docstring, but I think it would be good to include in public docs _somewhere_. Entity matching is a core feature of the system that people do and will ask about.

recidiviz/persistence/entity_matching.py
jessex: Got it, that sounds right. As part of general monitoring/operational work to come, we should make sure we have a runbook that includes catastrophic errors and how to address them if they arise.

recidiviz/persistence/entities.py
```diff
@@ -88,7 +88,6 @@ class Charge(object):
     external_id = attr.ib(default=None)
     offense_date = attr.ib(default=None)
     statute = attr.ib(default=None)
-    offense_code = attr.ib(default=None)
```
jessex: Why'd we remove these two fields?

recidiviz/persistence/entity_matching.py
terinpw: Created issue #354 to track.

recidiviz/persistence/entity_matching.py
terinpw: #355 to track.

recidiviz/persistence/entities.py
terinpw: They were removed in PR #315 from the schema so needed to be removed from the entities. Needed in this PR because the fields being present were breaking tests.

recidiviz/persistence/entity_matching.py
```diff
@@ -24,19 +26,199 @@ class EntityMatchingError(Exception):
...
+        person_a: (entities.Person)
+        person_b: (entities.Person)
+    Returns: (bool)
+    """
+    if person_a.external_id or person_b.external_id:
+        return person_a.external_id == person_b.external_id
+
+    # TODO(351): consider not matching if birthdate_inferred_from_age
+
+    if not all([person_a.given_names, person_b.given_names,
```
zacharylawrence: Should these be explicit `None` checks? For example, an empty string will match to `False`.

recidiviz/persistence/entity_matching.py
```diff
@@ -24,19 +26,199 @@ class EntityMatchingError(Exception):
...
+    Given a database booking and an ingested booking, determine if they should
+    be considered the same booking. Should only be used to compare bookings for
+    the same person.
+    Args:
+        db_booking: (entities.Booking)
+        ingested_booking: (entities.Booking)
+    Returns: (bool)
+    """
+
+    assert db_booking and ingested_booking
```
zacharylawrence: Why check this assertion here but not in other methods?

@zdg2102 Did we decide to avoid using `assert` in prod code? Instead favoring throwing our own exceptions?

recidiviz/persistence/entity_matching.py
```diff
@@ -24,19 +26,199 @@ class EntityMatchingError(Exception):
...
+    # they're inferred before we can simplify the following two cases.
 
-    Returns:
-        The entity matched Person or None if no matching can be done
+    # If the db booking's admission date was scraped (not inferred), the
+    # ingested booking must have the same admission date to be a match.
+    if not db_booking.admission_date_inferred:
+        return db_booking.admission_date == ingested_booking.admission_date
+
+    # return True if they are both active
```
zacharylawrence: Consider extracting a method `_is_active()` instead of this comment?
ie: `return _is_active(db_booking) and _is_active(ingested_booking)`

recidiviz/persistence/entity_matching.py
```diff
@@ -24,19 +26,199 @@ class EntityMatchingError(Exception):
...
+        region: (str)
+        ingested_people: List[entities.Person]
+    """
+    db_people = database.read_people_with_open_bookings(
+        session, region)
+
+    for db_person in db_people:
+        for ingested_person in ingested_people:
+            if is_person_match(ingested_person, db_person):
+                if ingested_person.person_id:
```
zacharylawrence: This is a bit tricky since it relies on setting the `person_id` below. Do you think it improves readability to create a list of all people that match, then check that that list has a length of 1?

recidiviz/persistence/entity_matching.py
```diff
@@ -24,19 +26,199 @@ class EntityMatchingError(Exception):
...
+           _sanitize_sentence(sentence_b)
+
+
+def _sanitize_sentence(sentence):
+    sanitized = copy.deepcopy(sentence)
+    sanitized.sentence_id = None
+    return sanitized
+
+
+def match_entities(session, region, ingested_people):
```
zacharylawrence: Comparing all these objects feels like a separate concept from the above definitions of "matching". Does it make sense to break these methods out into a new file/new suite of tests?

recidiviz/tests/persistence/database/database_test.py
```diff
@@ -81,3 +83,30 @@ def test_readOpenBookingsBeforeDate(self):
...
+                                         bookings=[closed_booking])
+
+        session = Session()
+        session.add(person)
+        session.add(person_no_open_bookings)
+        session.commit()
+
+        people = database.read_people_with_open_bookings(session, _REGION)
+
+        assert people == [database_utils.convert_person(person)]
```
zacharylawrence: I think we're trying to move away from `assert` to using `self.assertEqual()` with the `unittest` library: https://docs.python.org/2/library/unittest.html

recidiviz/tests/persistence/entity_matching_test.py
```diff
@@ -0,0 +1,212 @@
...
+    def test_person_match_name_and_birthdate(self):
+        db_person = entities.Person(given_names=_GIVEN_NAMES,
+                                    surname=_SURNAME,
+                                    birthdate=_DATE)
+        ingested_person = entities.Person(given_names=_GIVEN_NAMES,
+                                          surname=_SURNAME,
+                                          birthdate=_DATE)
+        self.assertTrue(entity_matching.is_person_match(
+            db_person, ingested_person))
+        ingested_person.birthdate = _DATE_ANOTHER
```
zacharylawrence: Nit: Feel free to ignore but this is actually testing 2 cases and should probably be two separate tests.

recidiviz/persistence/entity_matching.py
jamwalla: no, I think we shouldn't match either way - maybe the converter shouldn't let empty strings get through to here, though. What do you think?

recidiviz/persistence/entity_matching.py
jamwalla: sorry, removed! We eventually want `is_booking_match` to have named and required parameters since order matters a lot.

recidiviz/persistence/entity_matching.py
jamwalla: Added `_get_only_match` to get one match and error if there are more. We also need the check on `ingested_person.person_id` to make sure the match we find wasn't already matched to an earlier `db_person`.

recidiviz/tests/persistence/entity_matching_test.py
jamwalla: zach vs zach offline chat

PR #347, opened 2018-12-18
Add README to database package

zdg2102:


recidiviz/persistence/database/README.md
```diff
@@ -0,0 +1,73 @@
+# Database
+
+## Migrations
+
+### General Warnings
+
+- Do not make any changes to the schema file without also updating the database according to the workflow below. The schema and the database must not be out of sync.
+
+- A migration should consist of a single change to the schema. Do not group changes.
```
jessex: How strict is this rule? Is it literally only a single update statement per migration? Or is it a more conceptual guideline similar to how we think about commits or pull requests?

recidiviz/persistence/database/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+- Updates to the database are made before the new prod release containing the updated schema file. Keep this in mind when determining dependencies (e.g. if you are dropping a column, all references to it in the application code must be removed from the current release **before** running the migration).
```
jessex: Sorry if I'm being dense but I didn't totally follow. "Updates to the database are made before the new prod release" but also updates "in the application code must be [made in] the current release **before** running the migration." Is it that when we make a schema update, it should be in a commit/release with the accompanying code changes?

jessex: Oh after reading the workflow below I understand. Question: what's the state of our system in between the migration and the new release with the updated schema.py? Depends on what the migration is, and whether it's backwards compatible?

zdg2102: So after thinking through this a bit more, I realized the order will depend on if the change is an add or a drop. I updated this bullet and the workflow steps to reflect this.

Essentially the idea is the database and release will always be compatible, and it's the responsibility of the person making the change to ensure that's the case. This may involve making some changes in stages over multiple releases.

recidiviz/persistence/database/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+### Database Constraints
+
+- For historical tables, the primary key exists only due to requirements of SQLAlchemy, and should never be used or referenced by another table.
+
+- Foreign key constraints cannot reference historical tables because their keys are not unique.
+
+- If adding a column of type `String`, the length must be specified.
```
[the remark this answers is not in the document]
zdg2102: Sorry, github isn't making it really clear which line you're referring to. Which point should be clarified?

recidiviz/persistence/database/README.md
zdg2102: Yeah, I meant it as the latter. Not one statement, but one conceptual change. So adding an enum column and the enum that goes with it could be one change, but not adding a column on one table and dropping a column on another.

jessex: Let's just clarify that a slight little bit, again briefly.

recidiviz/persistence/database/README.md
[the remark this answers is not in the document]
jessex: "If adding a column of type String, the length must be specified."

I think optional string length in SQL is something most engineers are accustomed to, so it's good to briefly say why the length must be specified.

recidiviz/persistence/database/README.md
zdg2102: Done

recidiviz/persistence/database/README.md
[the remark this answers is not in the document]
zdg2102: Done

recidiviz/persistence/database/README.md
```diff
@@ -0,0 +1,95 @@
+# Database
+
+## Migrations
+
+### General Warnings
+
+- Do not make any changes to the schema file without also updating the database according to the workflow below. The schema and the database must not be out of sync.
```
zacharylawrence: Like we were talking about yesterday, do you know if it's possible to block prod pushes on this conditional?

recidiviz/persistence/database/README.md
```diff
@@ -0,0 +1,95 @@
...
+
+1. Get dev database schema in sync with prod database schema.
+
+2. Generate the migration against the prod database.
+
+3. Run the migration against the dev database. Make sure it looks as expected.
+
+4. Check in the schema file changes and migration.
+
+5.
```
zacharylawrence: Thoughts on our discussion yesterday about this? Namely the idea of turning down our entire server (letting all state rest in cloud tasks & pub/sub), pushing the database change, then turning the server back up. This would allow us to avoid worrying about the order things happen in.

Dangers of planned downtime might include:
1. Disrupting direct ingest (ie. people relying on our API always being up)
2. Interrupting scrape runs mid session (is our server actually stateless?)
3. Others??

recidiviz/persistence/database/README.md
zdg2102: Because of the reversed order for adds vs. drops (migrate then release vs. release then migrate), we could only enforce that if we follow the pattern in your comment below, of stopping the whole system and making all the changes at once.

recidiviz/persistence/database/README.md
zdg2102: Yeah, I still think that's an interesting idea worth considering. I don't think it matters for this PR, since that would be such a different workflow that I'd essentially just delete most of this doc and rewrite it. I opened #393 for comparing the alternative workflows.

PR #352, opened 2018-12-18
Fix python3 build

colincadams:


recidiviz/ingest/sessions.py
```diff
@@ -313,8 +311,13 @@ def get_open_sessions(region_code,
...
 
-    if open_only:
-        session_query.add_filter('end', '=', None)
+    if not (include_open and include_closed):
+        if include_open:
+            session_query.add_filter('end', '=', None)
+        elif include_closed:
+            session_query.add_filter('end', '>', None)
+        else:
+            return []
```
jessex: I think it's worth documenting in the docstring that if you provide false for both of the `include`s that you'll receive an empty list. That might seem obvious if you've thought about sessions for a minute but it might not be as obvious as we think.

jessex:
I'm curious, were these issues related to Python 3 in some way, or just required by previous changes on the way to Python 3?

recidiviz/ingest/sessions.py
colincadams: Done.

colincadams:
This PR does three things that arguably should've been split up:

1. Change `infer_release_test` to use the new auth logic.
1. Fix how `infer_release` uses `sessions` to work with the syntax of the non `ndb` `sessions`.
1. Change `get_open_sessions` to `get_sessions` and add support for getting just closed sessions.

The first two were required to fix the build, but while I was in there I felt compelled to do the third as well. Once I made the change @terinpw pointed out that the query doesn't support inequality filters on `end` so it couldn't be as clean as I had thought, which makes that change less compelling. Happy to split it out to a separate PR since it isn't necessary and bloats the changes that actually fix the build.

recidiviz/ingest/sessions.py
```diff
@@ -323,7 +326,14 @@ def get_open_sessions(region_code,
     if most_recent_only:
         limit = 1
 
-    return _sessions_from_entities(session_query.fetch(limit=limit))
+    results = session_query.fetch(limit=limit)
+
+    # Datastore doesn't allow an inequality filter on `end` because we are
+    # sorting on `start`, so we have to do the filter after we get the results.
+    if not include_open:
+        results = (result for result in results if result.get('end'))
```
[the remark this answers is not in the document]
colincadams: Alright now this will work, but open to objections of structuring it this way.

Issue #151, opened 2018-10-20
Update license text


jessex:
"A platform for tracking granular recidivism metrics in real time" is now too specific. It should read "A data platform for criminal justice measurement and improvement."

ohinds:
I think this text was updated?

jessex:
Not updated yet!

PR #353, opened 2018-12-18
Move SQLAlchemy engine setup from __init__.py to server.py

zdg2102:
Also update URL with unix_socket param required by Cloud SQL


recidiviz/server.py
```diff
@@ -18,16 +18,45 @@
...
     app.register_blueprint(test_populator, url_prefix='/test_populator')
+
+# Create URL and connect to database
+db_user = secrets.get_secret('sqlalchemy_db_user')
+db_password = secrets.get_secret('sqlalchemy_db_password')
+db_host = secrets.get_secret('sqlalchemy_db_host')
+db_name = secrets.get_secret('sqlalchemy_db_name')
+db_connection_name = secrets.get_secret('sqlalchemy_db_connection_name')
+
+sqlalchemy_url = '{url_prefix}://{db_user}:{db_password}@{db_host}/'\
```
arian487: This only applies for prod correct?  We shouldn't have the unix socket if running this locally.

zdg2102: @colincadams I was under the impression that `server.py` only runs in prod, not on local, is that correct?

arian487: If thats the case, then we are good, I'll just approve this :)

colincadams: @zdg2102 this runs both in prod and locally, the entrypoint is set in [`base.yaml`](https://github.com/Recidiviz/recidiviz-data/blob/master/base.yaml#L14) which is used by both. We could create a separate `local_server.py` and give that to `local_app.yaml` but I would prefer to just have one `server.py` and wrap it in a conditional.

colincadams:
Looks like some people like [`flask-sqlalchemy`](http://flask-sqlalchemy.pocoo.org) and it can handle some of this initialization for us. It isn't totally clear to me what advantages it has so might not be worth adding the dependency, but worth looking into more.

recidiviz/server.py
zdg2102: Fixed to now only run after the GAE check

recidiviz/__init__.py
```diff
@@ -1,54 +1,18 @@
...
 """Top-level recidiviz package."""
 import datetime
 from typing import Optional
 
 import cattr
-import sqlalchemy
 from sqlalchemy.engine import Engine
 from sqlalchemy.orm import sessionmaker
 
 from recidiviz.ingest.models.ingest_info import IngestInfo
```
colincadams: we should remove all of these imports in a separate PR

recidiviz/__init__.py
```diff
@@ -1,54 +1,18 @@
-# Recidiviz - a platform for tracking granular recidivism metrics in real time
```
jessex: Just curious why we removed this license text?

colincadams: oops missed this in my review. @zdg2102 can you add it back?

zdg2102: I thought we were supposed to be getting rid of it, for #151?

jessex:
No, I'm going to update just the small description at the top, we shouldn't be removing it entirely. Would you mind adding it back when you have a moment?

recidiviz/__init__.py
jessex: No, I'm going to update just the small description at the top, we shouldn't be removing it entirely. Would you mind adding it back when you have a moment?

zdg2102: Yep, #1391

PR #369, opened 2018-12-21
Remove deferred

colincadams:


colincadams:
Unfortunately at this point there isn't a good way to run everything end-to-end locally. Once we've finished all of the incompatible changes to python3, then we should be able to write a script that runs it locally. That will of course be based on the assumptions that we have for how cloud tasks works so it won't totally cover us, but it should be able to test this piece pretty fully.

For right now I added some further unit tests and tried to think hard about how this will change overall flow control, but there will probably be some bugs that we can't find until later 😬

PR #370, opened 2018-12-21
Change persistence actions to use flask blueprint

colincadams:
I mounted the `actions` blueprint at the prefix `/persistence`, so the full url would be `/persistence/v1/records`. Would you rather it be something else?


jessex:
"Persistence" feels a bit implementation-specific to me, honestly. It happens to be the implementation that we _persist_ ingested records to some permanent datastore, but that could change, and in any event isn't the primary interaction that is happening. The primary interaction is that we are _ingesting_ records.

I would mount this at `/ingest`. What do you all think?

colincadams:
That was my initial thought as well so I think that's a good way to go! It makes me wonder if we should move the `actions` file to the ingest directory and have it call into persistence, instead of having it live directly in persistence. Something we can think about as we expand direct ingest.

Also not sure if we will expose this url handler on our appengine app publicly directly, or put a separately managed api proxy in front which could have a separate endpoint.

Issue #366, opened 2018-12-20
DataExtractor not able to handle new scraper


HalfdanJ:
@arian487 

Data extractor is failing in PR #329. 

```
    for hier_class in hierarchy:
            # If we are a multi key class instead of getting the most recent
            # object as a parent, we use the index of the value we found
            if hier_class in self.multi_key_classes:
>               parent = getattr(parent, hier_class)[val_index]
E               IndexError: list index out of range

recidiviz/ingest/extractor/data_extractor.py:237: IndexError
```
And if i comment the part of the key yaml file out (bonds), then it produces too many charges.

PR #329, opened 2018-12-16
WIP us_fl_nassau_county scraper (and NewWorld vendor)

HalfdanJ:
Scraper for us_fl_nassau_county using NewWorld vendor that is included here. get_tasks should work, but populate data doesnt. The data extractor bugs out when trying to add a bond

```
    for hier_class in hierarchy:
            # If we are a multi key class instead of getting the most recent
            # object as a parent, we use the index of the value we found
            if hier_class in self.multi_key_classes:
>               parent = getattr(parent, hier_class)[val_index]
E               IndexError: list index out of range

recidiviz/ingest/extractor/data_extractor.py:237: IndexError
```

And without the bond, charges are not parsed correctly:
```
IngestInfo:
   person[0]: _Person:
      gender: Male
      race: White
      place_of_residence: FERNANDINA BEACH, FLORIDA 32034
      booking[0]: _Booking:
         release_date: 3/1/2017 12:13 PM
         admission_date: 1/1/2017 11:41 AM
         booking_id: 2017-00000000
         charge[0]: _Charge:
            name: CRIMINAL REGISTRATION (NOT AN ARREST)
            sentence: _Sentence:
            offense_date: 3/15/2017 11:44 AM
         charge[1]: _Charge:
            name: FELONY BATTERY
            sentence: _Sentence:
            offense_date: 5/19/2016 2:38 PM
            charge_class: Felony
         charge[2]: _Charge:
            name: AGG ASSAULT
            sentence: _Sentence:
            offense_date: 12/9/2015 4:50 PM
            charge_class: Misdemeanor
         charge[3]: _Charge:
            name: VIOLATE DOM VIOL COND
         charge[4]: _Charge:
            offense_date: 5/19/2016 2:39 PM
         charge[5]: _Charge:
            charge_class: Felony
         charge[6]: _Charge:
         charge[7]: _Charge:
            name: FEL BATT DOM W STRANGULATION
         charge[8]: _Charge:
            name: AGG ASSAULT W INT COMT FEL
         charge[9]: _Charge:
            offense_date: 11/15/2015 8:34 PM
         charge[10]: _Charge:
            offense_date: 11/15/2015 8:34 PM
         charge[11]: _Charge:
            charge_class: Felony
         charge[12]: _Charge:
            charge_class: Felony
         charge[13]: _Charge:
         charge[14]: _Charge:
         arrest: _Arrest:
      booking[1]: _Booking:
         release_date: 6/9/2016 5:30 AM
[...]

PR #378, opened 2018-12-21
DataExtractor proper cell ordering

jamwalla:
Fixes an issue where all `th` cells appeared before all `td` cells. #329 #366


jessex:
Is there a unit test case we can write for this?

arian487:
This lets us add back the functionality for unstructured data correct?

jamwalla:
@jessex- added a test, good call! @arian487, no, unfortunately that's a separate issue :/

PR #379, opened 2018-12-21
Extend Attr Entities to allow the Builder pattern to be used when creating Entity People

zacharylawrence:
Address the issue described in #363. Before calling that bug fixed, this change should also be propagated to the rest of the Entity objects.


recidiviz/common/buildable_attr.py
```diff
@@ -0,0 +1,114 @@
...
+            required_fields = set(attr.fields_dict(self.cls).keys())
+
+            fields_provided = set(self.fields.keys())
+            fields_with_defaults = {field for field, attribute in
+                                    attr.fields_dict(self.cls).iteritems() if
+                                    attribute.default is not attr.NOTHING}
+            fields_with_value = fields_provided | fields_with_defaults
+
+            if not required_fields == fields_with_value:
+                raise self.cls.BuilderException(
```
jamwalla: what do you think about raising this in `__setattr__` if the field isn't a real field? It seems more correct to raise on the line that the bad attr is set, but this exception is maybe explicit enough that it's ok.

recidiviz/persistence/converter/converter.py
```diff
@@ -51,12 +51,13 @@ def convert(self):
 
     def _convert_person(self, ingest_person):
         """Converts an ingest_info proto Person to a persistence entity."""
-        new_person = person.convert(ingest_person)
+        person_builder = person.convert(ingest_person)
```
terinpw: Let's have this method take in a builder and mutate it instead of returning a new builder.

terinpw: Also chatted with @zacharylawrence offline, but we should be more explicit in our naming/organization than person.convert

a) we should rename converter to be more descriptive of what it's actually converting, and 
b) maybe have the organizational structure of:
```
converter/
    converter.py
    converter_utils/
        common.py
        person.py
        bookings.py
        ...
```

That way when you call the method we explicitly preface the conversion with converter_utils.person.method_name

terinpw: Captured in Issue #385, not necessary for this PR.

recidiviz/common/buildable_attr.py
zacharylawrence: As we discussed offline, neither of us have strong feelings on this and will leave as is since it allows us to collect all the errors and print them out in one Exception.

@terinpw may have more thoughts on this?

recidiviz/persistence/converter/converter.py
zacharylawrence: Great point, done!

zacharylawrence: SGTM - will look into this in a follow up PR

recidiviz/common/buildable_attr.py
terinpw: Don't have a strong opinion here. I think the descriptive error message here has enough info to be really useful in debugging. 

We could add an additional check in` __setattr__` but I don't think it's necessary now.

zacharylawrence: SGTM - will leave this as is for now.

PR #380, opened 2018-12-21
DataExtractor: create parent multi_key classes if none exist

### LINKED ISSUE #283: data extractor should not create completely empty objects
In applying the data extractor to UsNy, I ran into an issue where the NYS DOCCS site always has 4 table rows to hold charges, even if there are less than 4 charges. So the extractor happily creates empty charges for each of these blank rows. The extractor should check if no fields of an entity have been filled in, and refuse to return an entity in that case.

### LINKED ISSUE #366: DataExtractor not able to handle new scraper

jamwalla:
The bug is caused by trying to attach bonds to charges before the charges are scraped.

close #283 
close #366


jessex:
To test this, I'd suggest writing a unit test case with data from the source site mentioned in the original issue. What do you think?

arian487:
At first glance I think it definitely fixes it, but definitely need to test it out.

jamwalla:
We can fall back on the create/get_recent logic for single-key classes when the parent class hasn't been created. This also fixes #283.

jamwalla:
Thanks Arian and Josh!

close #283 
close #366

PR #381, opened 2018-12-21
Delete us vt scraper

colincadams:


jessex:
How does this jive with #368?

colincadams:
The thought is that since there will be little to no shared code between the two versions, it makes sense to just delete it for now to unblock `python3` related work and then rewrite it later. @ohinds reached out and said that he thought that was the best way to proceed.

PR #388, opened 2018-12-22
Added manual index for sessions to fix inferred_release flow

terinpw:


terinpw:
Also added issue #389 to avoid manually adding indexes.

PR #390, opened 2018-12-22
us_fl_hendry_county scraper

HalfdanJ:
The html of this site challenging the data extractor, had to solve it with some complicated regex. Open to suggestions how to solve it more elegantly.


HalfdanJ:
Give the HTML a look. Its not separated by elements, it's all inline, so doubt it can be solved neatly

HalfdanJ:
Perhaps some of them can be solved, like this one. 

<p><strong>Charge:</strong> 893.13.6a / DRUGS-POSSESS</p>

queue.yaml
```diff
@@ -117,3 +117,14 @@ queue:
     min_backoff_seconds: 5
     max_backoff_seconds: 300
     task_age_limit: 3d
+
+# us-fl-hendry-county-scraper - Hendry County County, Florida
+- name: us-fl-hendry-county-scraper
+  mode: push
+  rate: 5/m
```
ohinds: Should we leave a comment in here that we should empirically determine a good rate here?

queue.yaml
```diff
@@ -117,3 +117,14 @@ queue:
     min_backoff_seconds: 5
     max_backoff_seconds: 300
     task_age_limit: 3d
+
+# us-fl-hendry-county-scraper - Hendry County County, Florida
+- name: us-fl-hendry-county-scraper
```
ohinds: Dupilcate of lower comment. Just flagging here in the first place the region name occurs. 

We've been leaving out the type of geographic entity from these names to keep them shorter. So I think it's ok to drop 'county' from the region name in this PR.

recidiviz/ingest/us_fl_hendry_county/us_fl_hendry_county_scraper.py
```diff
@@ -0,0 +1,149 @@
...
+
+    def populate_data(self, content, params, ingest_info):
+        data_extractor = DataExtractor(self.mapping_filepath)
+        ingest_info = data_extractor.extract_and_populate_data(content,
+                                                               ingest_info)
+
+        person = ingest_info.person[0]
+
+        person.full_name = \
+            content.cssselect('#WADAPageTitle > h3')[0].text_content()
```
ohinds: The full name can currently be grabbed with the extractor with a yaml css_key_mappings section like:
`css_key_mappings:
  '#WADAPageTitle > h3': person.full_name
`

recidiviz/ingest/us_fl_hendry_county/us_fl_hendry_county_scraper.py
```diff
@@ -0,0 +1,149 @@
...
+
+        person.get_recent_booking().custody_status = \
+            content.cssselect('td[align="right"] h2')[0].text_content()
+
+        tables = content.cssselect('table')
+        for table in tables:
```
ohinds: @arian487 you are more on top of how one might handle these table rows for bookings in the data extractor. Can you file issues on why these aren't handled, or otherwise delegate this task?

recidiviz/ingest/us_fl_hendry_county/us_fl_hendry_county_scraper.py
```diff
@@ -0,0 +1,149 @@
...
+        return {
+            'endpoint': self.get_region().base_url +
+                        "/inmate_search/INMATE_Results.php",
+
+            'task_type': constants.GET_MORE_TASKS
+        }
+
+    def _get_next_page(self, content, endpoint):
+        next_button = content.cssselect('[title="Next"]')
+        if len(next_button) is not 0:
```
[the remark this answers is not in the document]
HalfdanJ: It warns 
```
Do not use `len(SEQUENCE)` to determine if a sequence is empty pylint(C1801)
```

queue.yaml
jessex: I commented the same on another PR: I think we should file an issue for rate limiting / throttling in general, which mentions needing to update `queue.yaml`, or its Cloud Tasks replacement if necessary.

recidiviz/tests/ingest/us_fl_hendry/fixtures/search_results.html
```diff
@@ -0,0 +1,294 @@
...
+              <strong>Birth Date:</strong>
+              01-11-1997<br />
+                              
+              <strong>Custody Status:</strong> 
+              	<font color=red><strong>IN CUSTODY</strong></font>
+ 
+                <br />
+                <strong>Book Date:</strong>
+                03-18-2018                <br />
+                <strong>Booking #:</strong>HCSO18JBN000696<br />
```
jessex: These HCS* codes aren't PII or sensitive somehow, are they?

HalfdanJ: Good catch

recidiviz/ingest/us_fl_hendry_county/us_fl_hendry_county_scraper.py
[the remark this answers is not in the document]
ohinds: Oh right, sorry. Just `if next_button`, I think.

jamwalla:
I gave this another look over, I think the data extractor supports this page as well as we want it to now. If you like, we could use the extractor to get booking data as follows:

```yaml
key_mappings:
  - Booking Date: booking.admission_date
  - Release Date: booking.release_date
...
css_key_mappings:
  - 'a[href^="javascript:ChargeList"]': booking.booking_id
```
(actually, I need to land #535 to get the last part to work)

I don't think the data extractor can handle the charges. They're formatted as tables, but the layout is nonsensical. Here's a row from the charge table:
```html
<td><p><b>Charge:</b>CHARGE STATUTE</p><p>CHARGE NAME</p></td>
<td><p>Counts:</p><p>Bond Amount</p></td>
<td><p>1</p><p>$0</p>
```

One thing that will help us is that the charge tables are contained in `div`s with ids, so `#charge1` gets us the first charge table. If we use the DataExtractor to get a list of bookings, we can iterate over the bookings in order and add the corresponding charge info.

Hope this helps!

recidiviz/ingest/us_fl_hendry_county/us_fl_hendry_county_scraper.py
jamwalla: Mostly these data extractor issues are fixed; I left a comment arguing we shouldn't try to support the charge table. Marking this as resolved, but feel free to disagree!

HalfdanJ:
@jamwalla that worked. I've cleaned up the pr a bit, but I'd like some advice on how to do the manual html parsing correctly. This way seems a bit fragile.

HalfdanJ:
Race missing: `recidiviz.common.constants.mappable_enum.EnumParsingError: Could not parse INDIAN when building <enum 'Race'>`

arian487:
I'd add INDIAN to the global map as it seems generic enough.  In the recidiviz/common/constants/person.py look for _RACE_MAP and add 'INDIAN' mapping to Race.AMERICAN_INDIAN_ALASKAN_NATIVE.  We've added documentation for this in the create a scraper doc recently.

arian487:
Thanks for the fixes!  Approved.  I assume we are ready to merge?

HalfdanJ:
From my side yes.

arian487:
Wonderful, I'll merge this in too.

Issue #408, opened 2018-12-26
There should be a simple scraper simulator for easy development


ohinds:
Think a CLI tool that simulates basic queuing and requesting functionality, and does validation checks on the generated `ingest_info`.

ohinds:
There should be a system for catching and logging all errors.

arian487:
This is complete/

PR #420, opened 2018-12-27
Cli script

arian487:
Adds a simple CLI script which runs a scraper in memory rather than running it using appengine.  This makes the running much faster, and gives the writer the satisfaction of seeing successful runs.  Because it interacts with the persistence layer, it runs all of the conversions and validations.  This handles #408


arian487:
Additionally, this remove the need for any environment setup at all from the users end which is a huge win.

recidiviz/ingest/run_scraper.py
```diff
@@ -0,0 +1,119 @@
...
+        'scraper_start_time': datetime.datetime.now()})
+
+
+def _create_parser():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--region', required=True, help='The region to test')
+    parser.add_argument(
+        '--num_tasks', required=False,
+        default=num_tasks_left,
+        help='The number of tasks to complete, default is 5')
```
ohinds: You could .format() in the num_tasks_left var to protect against the default global being changed here.

recidiviz/ingest/run_scraper.py
```diff
@@ -0,0 +1,119 @@
...
+    parser.add_argument(
+        '--sleep_between_requests', required=False,
+        default=sleep_between_requests,
+        help='The number of seconds to sleep in between requests, default is 1')
```
ohinds: Same here on formatting in the sleep_between_requests

recidiviz/ingest/scraper_utils.py
```diff
@@ -172,8 +172,7 @@ def get_proxies(use_test=False):
     in_prod = environment.in_prod()
 
     if not in_prod or use_test:
-        user_var = "test_proxy_user"
-        pass_var = "test_proxy_password"
+        return None
```
ohinds: I think this is why the tests are failing?

recidiviz/ingest/run_scraper.py
arian487: Done.

recidiviz/ingest/scraper_utils.py
arian487: Tests should be fixed in the most recent commits

recidiviz/ingest/run_scraper.py
```diff
@@ -0,0 +1,119 @@
...
+import datetime
+import logging
+import time
+import types
+from google.appengine.api import apiproxy_stub_map
+from google.appengine.api import urlfetch_stub
+from recidiviz.ingest import constants
+from recidiviz.utils import regions
+
+# Sleep 1 seconds per task
```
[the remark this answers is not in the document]
arian487: Done

recidiviz/ingest/run_scraper.py
```diff
@@ -0,0 +1,119 @@
...
+sleep_between_requests = 1
+
+# Number of people to scrape by default
+num_tasks_left = 5
+
+
+def add_task(self, task_name, params):
```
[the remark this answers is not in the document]
arian487: Done.

recidiviz/ingest/run_scraper.py
```diff
@@ -0,0 +1,119 @@
...
+    """Overwritten version of add task which simply runs the task
+    and ignores the usage of the task_queue."""
+    logging.info('***')
+    global num_tasks_left
+    global sleep_between_requests
+    num_tasks_left -= 1
+    # If we are done, we can exit.
+    if num_tasks_left == 0:
+        logging.info('Successfully completed the test run!')
```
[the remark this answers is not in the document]
arian487: True, Done.

recidiviz/ingest/run_scraper.py
arian487: Done

arian487:
Thanks oliver, I'll wait for Josh's comments before mergine.

recidiviz/ingest/scraper_utils.py
```diff
@@ -216,7 +215,12 @@ def get_headers():
         proceed without this
 
     """
-    user_agent_string = secrets.get_secret("user_agent")
+    in_prod = environment.in_prod()
+    if not in_prod:
+        user_agent_string = ('For any issues, concerns, or rate constraints,'
```
jessex: I don't love having any user agent string that identifies us being listed in an open source codebase. In fact, Andrew had the good idea at one point to include some sort of hashed token in the user agent somewhere that would uniquely identify us, that we could use to prove or disprove that a given request came from us.

PR #421, opened 2018-12-28
Switch to python 3 interpreter

colincadams:
This makes a bunch of syntax changes as well


colincadams:
Note I haven't tried to run this on appengine yet, I'm saving that for tomorrow 😄

recidiviz/tests/utils/auth_test.py
```diff
@@ -90,9 +90,9 @@ def test_authenticate_request_from_iap_invalid(self, mock_jwt):
         response = self.client.get(
             '/', headers={'x-goog-iap-jwt-assertion': '0'})
         assert response.status_code == 401
-        assert response.get_data() == 'Error: INVALID TOKEN'
+        assert response.get_data() == b'Error: INVALID TOKEN'
```
jessex: I see a variety of strings that we don't explicitly mark as bytes literals. Why these? Because the Flask `response.get_data()` method returns a bytes literal?

recidiviz/calculator/recidivism/calculator.py
```diff
@@ -304,25 +304,25 @@ def stay_length_bucket(stay_length):
     """
     if stay_length is None:
         return None
-    elif stay_length < 12:
+    if stay_length < 12:
```
rasmi: Out of curiosity, is there any reason to remove the `elif`s? I think this could also be rewritten with some kind of `list(range(12, 120 + 1, 12))` and `'{}{}'.format(lower, upper)` logic but I think this way is clear too.

recidiviz/common/buildable_attr.py
```diff
@@ -18,23 +18,23 @@
 import attr
 
 
-class BuildableAttr(object):
+class BuildableAttr():
     """Abstract base class for Attr objects that can be built with a Builder."""
 
     # Consider BuildableAttr abstract and only allow instantiating subclasses
-    def __new__(cls, *args, **kwargs):
+    def __new__(cls, *_args, **_kwargs):
```
rasmi: Out of curiosity, why the `_`?

recidiviz/ingest/base_scraper.py
```diff
@@ -227,7 +227,6 @@ def get_more_tasks(self, content, params):
             A list of dicts each containing endpoint and task_type at minimum
             which tell the generic scraper how to behave.
         """
-        pass
```
rasmi: I'm surprised these don't raise syntax errors! I didn't know about this: https://stackoverflow.com/a/45826761

recidiviz/ingest/extractor/data_extractor.py
```diff
@@ -138,7 +139,7 @@ def extract_and_populate_data(self, content, ingest_info=None):
         # We use this set to keep track of keys we have seen, by the end of this
         # function it should be the empty set.  If not we throw an error to let
         # the user know we have a problem.
-        needed_keys = set(self.keys.keys() + self.multi_keys.keys())
+        needed_keys = set(self.keys.keys()) | set(self.multi_keys.keys())
```
rasmi: Isn't the result equivalent?

recidiviz/ingest/queues.py
```diff
@@ -21,10 +21,19 @@
 
 from google.cloud import tasks_v2beta3
 
-from recidiviz.utils import metadata
+from recidiviz.utils import environment, metadata
 
-client = tasks_v2beta3.CloudTasksClient()
+_client = None
```
rasmi: This is better, thanks :)

recidiviz/tests/calculator/recidivism/calculator_test.py
```diff
@@ -334,12 +334,12 @@ def test_for_characteristics():
     combinations = calculator.for_characteristics(characteristics)
 
     assert combinations == [{},
-                            {'age': '<25'},
                             {'race': 'black'},
                             {'sex': 'female'},
+                            {'age': '<25'},
```
rasmi: ?

recidiviz/tests/ingest/queues_test.py
```diff
@@ -21,22 +21,28 @@
...
-from google.cloud import tasks_v2beta3
-
 from recidiviz.ingest import queues
 from recidiviz.utils import metadata
 
 
 class QueuesTest(unittest.TestCase):
     """Cloud Tasks queue helper function tests."""
 
+    def setup_method(self, _test_method):
```
rasmi: This is better, thanks!

recidiviz/tests/utils/auth_test.py
```diff
@@ -28,17 +28,17 @@
...
+BEST_ALBUM = b'Music Has The Right To Children'
 APP_ID = "recidiviz-auth-test"
 
 dummy_app = Flask(__name__)
 
 @dummy_app.route('/')
 @auth.authenticate_request
 def boards_of_canada_holder():
-    return (BEST_ALBUM, 200)
+    return (BEST_ALBUM.decode(), 200)
```
rasmi: See main comment on `.decode()`

recidiviz/calculator/recidivism/calculator.py
colincadams: The updated pylint version gave a "no-else-return R1705" warning so I removed it so Travis would pass. I definitely think this could be rewritten but would opt to do it in a different PR.

recidiviz/ingest/queues.py
colincadams: np, I still don't love it but creating the client in python3 was throwing some errors about unset environment variables other than `APPLICATION_ID`, so I thought it was best to lazily create the client to avoid this. In the future I'm hoping to come up with a better strategy for these global singletons.

recidiviz/tests/calculator/recidivism/calculator_test.py
colincadams: The changes in calculator caused it to create the combinations in a different order, so I updated the test to reflect that. It should probably just create sets of both and compare those however.

colincadams:
* Happy to remove the empty params on class definitions if that is preferred.
* I think I'd opt for decode? I'll update to use this consistently.
* I don't have a strong preference here. Unfortunately afaik the [logger still uses c-style format strings](https://cloud.google.com/logging/docs/setup/python#using_the_python_root_logger) so we can't be completely consistent by moving to the `'{}'.format()` style, but there are definitely some places where we could update it: `grep -Er ' % ' recidiviz | wc -l` gives `26`.

I'll make the other `2to3` changes, thanks for the review!

recidiviz/calculator/recidivism/calculator.py
rasmi: Makes sense!

recidiviz/common/buildable_attr.py
rasmi: Yeah I think `#pylint: disable=unused-argument` is the way to go, I think it's best to stick to established style!

recidiviz/ingest/base_scraper.py
rasmi: Amazing

recidiviz/ingest/extractor/data_extractor.py
rasmi: Makes sense!

recidiviz/tests/calculator/recidivism/calculator_test.py
rasmi: Got it, thanks!

rasmi:
The following command removes parens from class names on OS X:
```
grep -E -l -r 'class [a-zA-Z]+\(\)' recidiviz | xargs sed -E -i '' 's/class ([a-zA-Z]+)\(\)/class \1/g'
```

recidiviz/common/buildable_attr.py
colincadams: Done

recidiviz/tests/utils/auth_test.py
colincadams: fixed re @rasmi's comments

recidiviz/tests/utils/auth_test.py
colincadams: Done

rasmi:
I realize that `response.get_data().decode() == 'My string.'` may fail less gracefully than `response.get_data() == b'My string.'` in the case where `response.get_data()` returns `None` (is that ever the case?). Both fail, but the error message for the latter is more helpful. Not sure if it's worth changing back. Up to you!

colincadams:
It is not currently ever the case although I think it is theoretically possible (maybe if the response is `204 NO CONTENT`? I'd say that it's fine for now, and if debugging tests becomes an issue in the future we can swap it out.

PR #424, opened 2018-12-28
ensure scraper_start_time is passed as datetime, not date

terinpw:
Update scraper_start_time to be a datetime not date
Update persistence.infer_release_on_open_bookings to store release_date as date instead of datetime.


recidiviz/common/common_utils.py
```diff
@@ -55,18 +55,34 @@ def parse_date_string(date_string):
             # Without a default day of `01` to fall back on, it falls back to
             # the local machine day, which will fail because February does not
             # have that many days.
-            result = parser.parse(date_string, default=datetime(2018, 1, 1))
-            result = result.date()
+            result = parser.parse(datetime_string, default=datetime(2018, 1, 1))
+            # result = result.date()
```
jessex: Should we be deleting this instead of keeping a commented line in the codebase? "Use it or lose it"

PR #426, opened 2018-12-28
Documentation for the data extractor.

arian487:
First pass at documentation for the data extractor.


recidiviz/ingest/extractor/README.md
```diff
@@ -0,0 +1,147 @@
...
+```
+
+This lets you avoid needing to extract this out yourself.  
+
+#### multi_key_mapping
+
+The final thing to fill out in the yaml is the multi key mapping.  Some fields are a one to many mapping.  For example, a person has multiple charges and the list of charges is listed in a table.  Therefore, there isn't just one value, but many values that need to be scraped for a single key.  Here is an example from the same webpage that lists a persons charges:
+
+<div align="center"><img src="https://i.imgur.com/y3L4MEX.png" border=0></div>
+
```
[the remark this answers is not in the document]
arian487: What do you mean exactly by 'on its own'?  Its already in its own section.

recidiviz/ingest/extractor/README.md
```diff
@@ -0,0 +1,147 @@
...
+Note that we added Offense DateTime and Arrest DateTime to the ignored keys because it is also always empty.  This simple yaml file allows us to easily look at a webpage and tell the data extractor how to extract the fields.  The user doesn't need to worry about any HTML parsing!
+
+### The Code
+
+Now that the yaml is written, the last step is to tell the scraper to use the extractor.  This process is extremely simple, the first step is to import the data extractor:
+```python
+from recidiviz.ingest.extractor.data_extractor import DataExtractor
+```
+
+Next, get the directory of the yaml, which should be saved in your regions directory:
```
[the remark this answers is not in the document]
arian487: Done.

recidiviz/ingest/extractor/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+```python
+yaml_file = os.path.join(os.path.dirname(__file__), 'my_yaml.yaml')
```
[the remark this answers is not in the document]
arian487: Well it doesn't really need to be in the init.  It can be defined on populate_data if they want.  I suppose if I were doing a PR I'd tell them to define it in the __init__.  I'll clear that up.

recidiviz/ingest/extractor/README.md
```diff
@@ -0,0 +1,147 @@
...
+def populate_data(self, content, params, ingest_info):
+        """
+        Populates the ingest info object from the content and params given
+
+        Args:
+            content: An lxml html tree.
+            params: dict of parameters passed from the last scrape session.
+            ingest_info: The IngestInfo object to populate
+        """
+        data_extractor = DataExtractor(yaml_file)
```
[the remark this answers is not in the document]
arian487: Done.

recidiviz/ingest/extractor/README.md
```diff
@@ -0,0 +1,148 @@
+
+# Data Extractor
+
+The Data Extractor is a tool designed to make the extraction of data from a website much simpler.  At a high level the data extractor takes a webpage, and attempts to extract values from provided keys so the user doesn't need to worry about walking HTML trees or writing css selectors to get the data they need.  It will smartly find all of the keys the user lists and maps them to the correct fields in the IngestInfo object.  The result is a populated and properly construct object containing the information that exists on a page.
```
jessex: Grammar nit: "It will smartly find all of the keys the users lists and _map_ them to the correct fields in the IngestInfo object."

Also: "The result is a populated and properly _constructed_ object containing the information that exists on a page." But I might reword that slightly to, "The result is a consistently structured, populated object containing..."

recidiviz/ingest/extractor/README.md
```diff
@@ -0,0 +1,148 @@
...
+
+## How To Use
+
+### The Yaml File
+
+The first step is to create a yaml file which contains all of the keys that you care about, and how they map to the underlying ingestion object fields.  To see a listing of all of the possible fields and their descriptions, see [IngestInfo](https://github.com/Recidiviz/recidiviz-data/tree/master/recidiviz/ingest/models/README.md).  
+
+#### key_mappings
+
+The first piece to fill in is the key_mappings.  Specifically this is data on a page that is listed as single key value pairs on a web page.  That is one value per key.  An example is the following:
```
jessex: Backticks around anything that's a reference, i.e. `key_mappings`.

jessex: Nit: I tend to like to use key-value pairs, in the hyphenated form.

recidiviz/ingest/extractor/README.md
```diff
@@ -0,0 +1,148 @@
...
+
+<div align="center"><img src="https://i.imgur.com/oHVTVcU.png" border=0></div>
```
jessex: Bookkeeping question: is that Imgur link stable? Will it ever expire? Would it make more sense to host doc images out of a specified directory?

recidiviz/ingest/extractor/README.md
```diff
@@ -0,0 +1,148 @@
...
+  Race: person.race
+  Committed By: booking.hold
+  Booking Date-Time: booking.admission_date
+  "Booking #": booking.booking_id
+```
+
+Note that in this case, this particular site has fields for custody status and release date-time, but in fact they are never populated.  We want to let the data extractor know to ignore these fields.
+
+#### keys_to_ignore
+The next thing that needs to be filled out is the keys_to_ignore.  This tells the data extractor to ignore these fields and not try to extract them.  The important thing to consider here is that the specific key should be ignored for all extractions.  This can be used for fields such as height, weight, hair colour, etc... because they do not map to anything in our internal database and we don't care about them.  The keys_to_ignore for the yaml file would be:
```
jessex: Backticks around `keys_to_ignore` in both places here. I promise not to make anymore backtick comments :)

recidiviz/ingest/extractor/README.md
```diff
@@ -0,0 +1,147 @@
...
+```
+
+Note that in this case, this particular site has fields for custody status and release date-time, but in fact they are never populated.  We went to let the data extractor know to ignore these fields.
+
+#### keys_to_ignore
+The next thing that needs to be filled out is the keys_to_ignore.  This tells the data extractor to ignore these fields and not try to extract them.  The key thing to consider here is that the specific key should be ignored for all extractions.  This can be used for fields such as height, weight, hair colour, etc... because they do not map to anything in our internal database and we don't care about them.  The keys_to_ignore for the yaml file would be:
+
+```yaml
+keys_to_ignore:
+  - Custody Status
```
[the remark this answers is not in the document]
jessex: It might be worth clarifying _why_ this is an interesting example in the docs, then. To me, this comment makes the line, "We could have mapped it if we wanted" vague enough that I'm not sure what we're trying to say.

recidiviz/ingest/extractor/README.md
```diff
@@ -0,0 +1,148 @@
...
+  - Release Date-Time
+```
+
+In this case, these two fields do in fact map to our database, but this particular website never fills it in so we just tell the extractor to ignore it.  We could have mapped it if we wanted, but it would just always be returned as empty.
+
+#### css_key_mappings
+
+In the example above, the persons name does not have a key that we can tell the extractor to use.  The user could use the extractor to fill in all of the fields, and then in code also extract the name using html parsing but we also support css key mappings for this reason.  Similar to key_mappings, the user can put any css keys to also be extracted.  In the example above the css for the name looks like:
```
jessex: We've got a run-on sentence here. Break it up like so (with a few other suggestions _italicized_): "In the example above, the _person's_ name does not have a key that we can tell the extractor to use. The user could use the extractor to fill in all of the fields and then manually extract the name using html parsing. However, we also support css key mappings for this reason. Similar to `key_mappings`, the user can _list_ any css keys to also be extracted. In the example above, the css for the name looks like:"

recidiviz/ingest/extractor/README.md
```diff
@@ -0,0 +1,148 @@
...
+<span id="ctl00_ContentPlaceHolder1_spnInmateName" class="name">
+	Name, Test
+</span>
+```
+
+We can update our yaml with:
+
+```yaml
+css_key_mappings:
+  "#ctl00_ContentPlaceHolder1_spnInmateName": person.surname
```
jessex: I may be misinterpreting the example above, but the text value of the element is "Name, Test" which to me reads as `person.full_name` instead of `person.surname`. But I think given the fact that we are trying to concretize it all with examples, we should avoid vague example values in favor of obvious but fake ones. So if that's supposed to be a fake name, I'd go with something like "Knowles, Beyoncé" that everyone will get.

recidiviz/ingest/extractor/README.md
```diff
@@ -0,0 +1,148 @@
...
+```
+
+This lets you avoid needing to extract this out yourself.
```
jessex: "This lets you avoid extracting the value manually."

recidiviz/ingest/extractor/README.md
```diff
@@ -0,0 +1,148 @@
...
+```yaml
+css_key_mappings:
+  "#ctl00_ContentPlaceHolder1_spnInmateName": person.surname
+```
+
+This lets you avoid needing to extract this out yourself.  
+
+#### multi_key_mapping
+
+The final thing to fill out in the yaml is the multi key mapping.  Some fields are a one to many mapping.  For example, a person has multiple charges and the list of charges is listed in a table.  Therefore, there isn't just one value, but many values that need to be scraped for a single key.  Here is an example from the same webpage that lists a persons charges:
```
jessex: _person's_

recidiviz/ingest/extractor/README.md
jessex: Also, is the data in that image faked? If so, remember to update the faked name if you go with my Beyoncé suggestion below.

recidiviz/ingest/extractor/README.md
```diff
@@ -0,0 +1,148 @@
...
+
+<div align="center"><img src="https://i.imgur.com/y3L4MEX.png" border=0></div>
```
jessex: Are the case numbers in that image already faked? Should they be?

recidiviz/ingest/extractor/README.md
```diff
@@ -0,0 +1,148 @@
...
+  - Arrest DateTime
+
+multi_key_mapping:
+  Statute Code: charge.statute
+  Description: charge.name
+  CaseNumber: charge.case_number
+  Bond Amount: bond.amount
+```
+
+Note that we added Offense DateTime and Arrest DateTime to the ignored keys because it is also always empty.  This simple yaml file allows us to easily look at a webpage and tell the data extractor how to extract the fields.  The user doesn't need to worry about any HTML parsing!
```
jessex: "... because _they are_ also always empty."

recidiviz/ingest/extractor/README.md
```diff
@@ -0,0 +1,148 @@
...
+
+### The Code
+
+Now that the yaml is written, the last step is to tell the scraper to use the extractor.  This process is extremely simple, the first step is to import the data extractor:
```
jessex: "This process is extremely simple. The first step is to import the data extractor:"

recidiviz/ingest/extractor/README.md
```diff
@@ -0,0 +1,148 @@
...
+        data_extractor.extract_and_populate_data(content, ingest_info)
+        return ingest_info
+```
+
+The data extractor will have a properly formatted object that you shouldn't need to touch. Note:  In some cases you may need to do some additional work on the ingest_info object (if a field couldn't be extracted properly), but in most cases we prefer to expand the functionality of the data extractor to handle the case.
+
+## When Does It Work?
+
+The data extractor works in a wide range of cases.  It works in the following situations:
+* A page that has data displayed in a table, whether well structured or poorly structured
```
jessex: _well-structured_

recidiviz/ingest/extractor/README.md
```diff
@@ -0,0 +1,148 @@
...
+	* The 'value' can appear to the right cell, or the cell below the key
+	* The values don't have to be consistently displayed, for example one value can be in the lower cell, and another can be the right cell on the same page
```
jessex: "can be _in_ the right cell..."

recidiviz/ingest/extractor/README.md
```diff
@@ -0,0 +1,148 @@
...
+* Unstructured key value pairs that are separated by a ':' and not in a table
+* It can handle multiple people per page
+
+The general rule of thumb is to attempt to the use the data extractor and ask questions later.  In most cases it'll just work!
+
+## When Doesn't it work
```
jessex: _It_

recidiviz/ingest/extractor/README.md
```diff
@@ -0,0 +1,148 @@
...
+
+The data extractor sometimes fails with very oddly structured data.  It is hard to enumerate these examples as we have not encountered them all.  There are cases when the data is displayed in a table that has many nested tables (it still catches most of these cases) resulting in data extraction errors.  
+
+Another current failure is when data for a single person is on many pages.  For example if their booking information is on one page, but to get at their charge information you need to click through to another page.  This behaviour is untested but a user can attempt to create two yaml files, one per page and use those populate the IngestInfo object.  The data extractor should properly update the object that is passed in but use at your own risk!  We plan to test and expand this functionality in the future.
```
jessex: "... use those _to_ populate the IngestInfo object."

jessex: Also, I noticed that in most places we camelCase `IngestInfo`, but in some we snake_case it as `ingest_info`. We should pick a convention and stick with it.

recidiviz/ingest/extractor/README.md
arian487: Done.

recidiviz/ingest/extractor/README.md
arian487: Fixed the nit, and added back ticks everywhere.

recidiviz/ingest/extractor/README.md
arian487: The imgur link is very stable and will never expire as long as imgur stays up :).  Personally I think its a fine place to host images for this purpose but we can certainly host it ourselves if needed.

recidiviz/ingest/extractor/README.md
arian487: Done.

recidiviz/ingest/extractor/README.md
arian487: I was the run-on sentence king back when I had english classes, as you are now discovering with this review :P

recidiviz/ingest/extractor/README.md
arian487: Fixed name, and made it fullname.

recidiviz/ingest/extractor/README.md
arian487: Done.

recidiviz/ingest/extractor/README.md
arian487: Done.

recidiviz/ingest/extractor/README.md
arian487: You're right, I scrubbed that out and redid the image :)

recidiviz/ingest/extractor/README.md
arian487: Done.

recidiviz/ingest/extractor/README.md
arian487: Done.

recidiviz/ingest/extractor/README.md
arian487: Done.

recidiviz/ingest/extractor/README.md
arian487: Done.

recidiviz/ingest/extractor/README.md
arian487: Done.

recidiviz/ingest/extractor/README.md
arian487: Done.

recidiviz/ingest/extractor/README.md
jessex: As long as the images have on sensitive information then it's not a bad option. But whenever we update the images, we want to make sure we delete the previous ones.

arian487:
Addressed comments, thanks for the review

recidiviz/ingest/extractor/README.md
jessex: Oh, it's looking great!

arian487:
I replied in the other PR, but it seems imgur doesn't recognize my device anymore (I didn't upload them on an account).  In order to remove the images I'd have to contact support.  Sorry about that my mistake.  Shall I go ahead and do that?

Also regarding hosting, imgur photos don't expire so IMO its a fine platform to host on, but we can definitely host them ourselves if you feel that is better?

jessex:
Yeah I thought about it some more this evening and I'm not super comfortable hosting images of sources we scrape, even scrubbed ones, on another domain that we don't have any ounce of control over. I'd rather we host them in a suitable directory here, like `/docs/img` or just `/img`. I think it just makes more sense for a project that deals with sensitive information and concepts.

arian487:
All fair points, I've committed the images to github instead for both documentation PRs and I'll also contact imgur to take the hosted images down.

PR #436, opened 2019-01-02
parse bond type in amount field and set defaults

### LINKED ISSUE #403: Bond conversion needs to better handle string amounts
If the 'amount' is 'no bond' or 'bond denied' it should additionally update the actual status of the bond to reflect that instead of just setting the amount to 0.

### LINKED ISSUE #404: NO BOND can also appear in total_bond_amount
Currently total_bond_amount is not set up to handle 'no bond' as a value, similarly to bond.amount.  This should be updated to handle it.

jamwalla:
close #403 + close #404


jamwalla:
@arian487, how do you feel about moving the checks for 'NO BOND' etc. into all dollar parsing? We can also make a `parse_bond_amount` function if you think fees etc. will be significantly different.

arian487:
Hm, I think we'll need a parse bond function of some kind.  Putting it in parse dollars will just return 0 but I think the behaviour for NO BOND and BOND DENIED need to be different.  If the site puts BOND DENIED in the amount, it probably makes more sense for us to set the bond_type and not set bond amount to 0.

jamwalla:
Added a converter_utils function `parse_bond_amount_and_infer_type`. The behavior is as follows:
- If the parsed amount is a nonzero dollar amount, set BondType.CASH and BondStatus.ACTIVE.
- If the parsed amount is zero or a string like `NO BOND`, set BondType.NO_BOND. (Leave amount and status empty.)
- If the parsed amount is a string like `DENIED`, set BondType.BOND_DENIED. (Also leave amount and status empty.)

recidiviz/persistence/converter/converter_utils.py
```diff
@@ -150,6 +151,19 @@ def split_full_name(full_name):
     raise ValueError('cannot parse full name: %s' % full_name)
 
 
+def parse_bond_amount_and_infer_type(amount):
+    if 'NO' in amount.upper():
```
jessex: Have we considered any values for which this would be a false negative? I don't have any myself but want to make sure.

jamwalla: I talked about it with Arian and couldn't think of any case where the data is scraped correctly and we hit a false negative, and we were inclined to not try to defend against weirdly scraped values. But if you think we should, we could try erroring when both `NO` and `$` are in the string - what do you think?

recidiviz/persistence/converter/bond.py
```diff
@@ -26,14 +26,14 @@ def convert(proto):
...
 
     new.external_id = fn(normalize, 'bond_id', proto)
-    new.amount_dollars = fn(_parse_bond_amount, 'amount', proto)
-    new.bond_type = fn(BondType.from_str, 'bond_type', proto)
-    new.status = fn(BondStatus.from_str, 'status', proto, BondStatus.POSTED)
+    new.amount_dollars, inferred_bond_type = fn(
+        converter_utils.parse_bond_amount_and_infer_type, 'amount', proto,
+        (None, None)) # default to a pair of Nones instead of None
+    # |inferred_bond_type| defaults to CASH unless the bond amount contains
+    # information about the type (i.e. NO_BOND or BOND_DENIED)
```
ohinds: Does this happen even when the bond amount is $0 or empty? I think in many cases that will correspond to 'NO BOND' rather than 'CASH'. Let's discuss.

recidiviz/persistence/converter/bond.py
```diff
@@ -26,14 +26,14 @@ def convert(proto):
...
+    new.bond_type = fn(BondType.from_str, 'bond_type', proto,
+                       inferred_bond_type)
+    new.status = fn(BondStatus.from_str, 'status', proto,
```
ohinds: I was thinking that 'BOND DENIED' could also be 'ACTIVE'. I could be convinced otherwise.

recidiviz/persistence/converter/converter_utils.py
ohinds: It might be a little safer to enforce startswith 'NO' here?

recidiviz/persistence/converter/bond.py
ohinds: nm, I see this later on. I like it!

jamwalla:
Added TODO(#491), but for now we'll leave bonds with BondType.BOND_DENIED without a BondStatus.

arian487:
Are there any blockers left on pushing this PR

PR #443, opened 2019-01-02
Pass metadata into data_converter

zacharylawrence:


recidiviz/persistence/entities.py
```diff
@@ -132,3 +132,10 @@ class Sentence(object):
     # To avoid recursive references, store only 1 level of related_sentences
     # (ie. don't store related_sentences of these related_sentences).
     related_sentences = attr.ib(default=None)
+
+
+@attr.s
+class Metadata(BuildableAttr):
```
terinpw: Let's call this something more specific than Metadata? Maybe ProtoToEntityMetadata?

recidiviz/persistence/converter/converter.py
```diff
@@ -53,7 +54,9 @@ def _convert_person(self, ingest_person):
         """Converts an ingest_info proto Person to a persistence entity."""
         person_builder = entities.Person.builder()
 
-        person.copy_fields_to_builder(ingest_person, person_builder)
+        person.copy_fields_to_builder(
+            person_builder, ingest_person, self.metadata)
```
terinpw: nit - feel free to ignore if you don't like this, but I think it'd be nice to explicitly send in any args here (i think it's just region). That way the entity specific conversions only have access to the fields they need.

same for booking.convert below.

zacharylawrence: As discussed offline, we'll leave this as is since the `metadata` isn't tied to specific parts of the ingest_info.

PR #444, opened 2019-01-02
Start using Builder for Booking

zacharylawrence:


recidiviz/persistence/converter/converter.py
```diff
@@ -66,22 +66,27 @@ def _convert_person(self, ingest_person):
 
     def _convert_booking(self, ingest_booking):
         """Converts an ingest_info proto Booking to a persistence entity."""
-        new_booking = booking.convert(ingest_booking, self.metadata)
+        booking_builder = entities.Booking.builder()
 
-        new_booking.arrest = \
+        booking.copy_fields_to_builder(booking_builder, ingest_booking,
```
jessex: Did you consider making a static `copy_from` (or just `from`) method on `Booking.Builder`? I think a more slick experience for developers would be something like `booking_builder = entities.Booking.Builder.from(ingest_booking).with(self.metadata)`.

recidiviz/persistence/converter/booking.py
```diff
@@ -17,14 +17,17 @@
...
 
 
-def convert(proto, metadata):
-    """Converts an ingest_info proto Booking to a persistence entity."""
-    new = entities.Booking()
+def copy_fields_to_builder(booking_builder, proto, metadata):
+    """Mutates the provided |booking_builder| by converting an ingest_info proto
+     Booking.
+
+     Note: This will not copy children into the Builder!
```
jessex: Why don't we want to copy children in the builder construction itself?

recidiviz/persistence/database/database_utils.py
```diff
@@ -77,7 +74,10 @@ def _convert_booking(booking_src):
                                    booking_src.charges]
         else:
             setattr(booking_dst, k, getattr(booking_src, k))
-    return booking_dst
+
+    if entity_to_db:
```
jessex: This feels like a prime candidate for drift leading to a bug, though test coverage should catch it. Can we be more explicit and just check the type of `booking_dst` here instead?

recidiviz/persistence/converter/converter.py
zacharylawrence: Normally I'd agree and like the idea of moving logic into the `Entity`/`Builder`, but I'm concerned in this case because `Builder.from()` would need to either not recursively set the children entities or would need to take the full `ingest_info` object and recursively build all children.

The first case feels dangerous since `Builder.from()` would either return an incomplete `Builder` or a built `Entity` that is missing it's children.

The second feels like the Builder might be assuming too much responsibility since it's also building all children entities. For example, `entities.Person.Builder.from(ingest_person).with(ingest_info, metadata)` would return the entirely converted `entities.Person` (although this might not be bad?).

I don't feel strongly about this and if you think of a clean way to move this conversion into the entities objects then I'm all for it!

recidiviz/persistence/converter/booking.py
zacharylawrence: Converting all the children here would require passing the whole `ingest_info` as a parameter. By breaking out field conversions from nested object conversions, we can create a method that just converts between `ingest_info.Booking` -> `entities.Booking`. I think the benefits of this isolation may be less tangible now that the data converter is broken up with seperate files per Entity Type.

I'm inclined to leave this change as is for now since it's how the the `Person` converter currently works and explore this proposal in another PR (which has benefits - like not passing in a builder as a parameter and mutating it).

recidiviz/persistence/converter/converter.py
jessex: >For example, entities.Person.Builder.from(ingest_person).with(ingest_info, metadata) would return the entirely converted entities.Person

I view that as a feature, personally. All classes contain other classes as fields, whether they are user-defined or from the standard library. `Setting` a stdlib field does not feel any different from `Setting` a `Built` user-defined field. I would expect that if I call `Person.build` that all of the fields on it are finalized, and if that means that some of those fields themselves need to have some method like `.build` called on them then so be it. Hand-wavingly, I'd structure this by building all children. Invoking the `.build` on a child class, if it exists, doesn't feel like too much responsibility to me as long as the logic of `.build` is within the child class itself.

Of course, my comment isn't providing anything particularly novel, it's about design preference.

PR #445, opened 2019-01-02
DataExtractor: don't convert keys that are already cells

### LINKED ISSUE #330: fix data extractor bug where key conversion changes the order of cells
calling `_convert_key_to_cells` on each key in `self.keys` can cause charges to be read as all belonging to the first person on a page (#327).

jamwalla:
Fixes an issue found in #327 where table headers containing key values were getting converted by `convert_key_to_cell` to `td` instead of `th`.

close #330.


jamwalla:
part of #366 as well

Issue #348, opened 2018-12-18
Determine where in ingestion pipeline to infer values based on scraped values


terinpw:
Examples:
if admission_date is None, we want to infer an admission date.

arian487:
Done in converter.

PR #446, opened 2019-01-02
Add logic for inferring admission_date

zacharylawrence:


recidiviz/persistence/converter/booking.py
```diff
@@ -29,26 +31,37 @@ def copy_fields_to_builder(booking_builder, proto, metadata):
...
     release_date_inferred = None if release_date is None else False
 
     return release_date, release_date_inferred
+
+
+def _parse_admission(proto):
+    admission_date = fn(parse_date, 'admission_date', proto)
+
+    if admission_date is None:
+        admission_date = datetime.datetime.now()
```
terinpw: I think we want this to be the last_seen_time.

terinpw:
Also now that we're inferring before entity matching, we can probably close #348. 

-- There's a small TODO in entity_matching.py related to it (need to update `if db_booking.admission_date_inferred:` to become `if db_booking.admission_date_inferred and ingested_booking.admission_date_inferred`.

recidiviz/persistence/converter/booking.py
zacharylawrence: Good point, updated!

zacharylawrence:
> Also now that we're inferring before entity matching, we can probably close #348.
> 
> -- There's a small TODO in entity_matching.py related to it (need to update `if db_booking.admission_date_inferred:` to become `if db_booking.admission_date_inferred and ingested_booking.admission_date_inferred`.

Good catch, done!

zacharylawrence:
> You should be able to mark 348 closed?

Looks like it's already closed

Issue #157, opened 2018-10-26
Decompose scraper_utils


jessex:
As discussed here (https://github.com/Recidiviz/recidiviz-data/pull/147#discussion_r226817845), `scraper_utils` has a variety of functions which do not necessarily belong in the same module, and it is at risk of becoming a difficult-to-read monolith as we add in more logic (as we will do in #156 and other future work). We should decompose it into more targeted modules with collections of related functions. A strawperson proposal for decomposing the module in its current state is included in the linked thread.

arian487:
Most of the logic in scraper_utils is obsolete, some of the stuff makes sense.  This should be a cleanup process instead.  Marking this as medium.

PR #447, opened 2019-01-02
return to dateparser

jamwalla:
We can use dateparser again! (#249) Also deletes some scraper_utils (#157) functions that are handled elsewhere.


recidiviz/persistence/converter/converter_utils.py
```diff
@@ -123,10 +123,15 @@ def parse_days(time_string):
     if time_string == '' or time_string.isspace():
         return 0
     try:
-        # TODO: use dateparser in Python3 (#176)
         return int(time_string)
-    except Exception:
-        raise ValueError('cannot parse time duration: %s' % time_string)
+    except ValueError:
+        # dateparser.parse interprets the string '1 YEAR 2 DAYS' to mean the
```
jessex: Clever.

recidiviz/persistence/converter/converter_utils.py
```diff
@@ -123,10 +123,15 @@ def parse_days(time_string):
...
+        # datetime 1 year and 2 days ago.
+        date_ago = dateparser.parse(time_string)
+        if date_ago:
+            return (datetime.datetime.now() - date_ago).days
```
jessex: I wonder if we'd be better off performing this date difference with `timedelta` instead of the builtin subtraction operation. timedelta just in general is built to consider edge cases around differences in time, so it feels like a wise move to use it for as many of these use cases as we have. What do you think?

jamwalla: The subtraction operation returns a timedelta, so I'm not sure I understand. What are you thinking we should use here?

jessex: Ah my apologies. I thought the `-` operator was performing raw arithmetic on the `date` components, I didn't know it was using the same logic as timedelta under the hood. Never mind!

PR #448, opened 2019-01-02
us_ny add sentence to all charges

jamwalla:
use a second DataExtractor to get sentence information, then add the extracted sentence to each charge. The sentence length string format can be parsed by dateparser (#447).


recidiviz/ingest/us_ny/us_ny_scraper.py
```diff
@@ -63,8 +63,10 @@ class UsNyScraper(BaseScraper):
     TOKEN_NAME = 'DFH_STATE_TOKEN'
 
     def __init__(self):
-        self.mapping_filepath = os.path.join(os.path.dirname(__file__),
-                                             'us_ny.yaml')
+        self.booking_mapping_filepath = os.path.join(os.path.dirname(__file__),
```
jessex: Interesting use of multiple extractors! Is this the recommended pattern for any website that has multiple distinct tables we want to extract on a given page?

recidiviz/ingest/us_ny/us_ny_scraper.py
```diff
@@ -303,6 +305,16 @@ def populate_data(self, content, params, ingest_info):
                 '//*[@headers="t1l"]')[0].text.split()[0]
             ingest_info.person[0].booking[0].release_date = release_date_string
 
+        sentence_extractor = DataExtractor(self.sentence_mapping_filepath)
+        sentence_info = sentence_extractor.extract_and_populate_data(content)
+        if len(sentence_info.person) != 1 or \
+                len(sentence_info.person[0].booking) != 1 or \
+                len(sentence_info.person[0].booking[0].charge) != 1 or \
+                sentence_info.person[0].booking[0].charge[0].sentence is None:
+            logging.error("Data extraction did not produce a single "
```
jessex: This assumes a non-empty sentence table, right? Have we encountered any empty ones in the wild?

jamwalla: Haven't yet, so let's leave this as is and break if we're wrong.

recidiviz/ingest/us_ny/us_ny_scraper.py
jamwalla: Yes, I think so! I was considering a smaller api, like `extract_one_sentence` or something, but I'm inclined to not unless we need more things like this.

PR #449, opened 2019-01-02
Add simple monitoring to persistence layer

colincadams:


recidiviz/utils/monitoring.py
```diff
@@ -0,0 +1,43 @@
...
+    if not _stats:
+        new_stats = stats_module.Stats()
+        exporter = stackdriver.new_stats_exporter(stackdriver.Options(
+            project_id=metadata.project_id(),
+        ))
+        new_stats.view_manager.register_exporter(exporter)
+        _stats = new_stats
+    return _stats
+
+@contextmanager
```
jessex: Let the Python 3 roll in!

recidiviz/utils/monitoring.py
```diff
@@ -0,0 +1,43 @@
...
+def measurements(tags=None):
+    mmap = stats().stats_recorder.new_measurement_map()
```
jessex: Do you have some API/tech docs you can point me to to understand what measurement maps are and how they work? I had trouble finding docs on them from a quick search.

recidiviz/utils/monitoring.py
```diff
@@ -0,0 +1,43 @@
...
+    yield mmap
+    mmap.record(tag_map_tags=tags)
```
jessex: What does this call do such that we'd like to yield the map up before invoking it?

jessex: Oh I understand now. Yielding the map before recording the tags is done so that the code making use of the measurement map can change the tags before the managed context ends, so the tags set on the measured stats can be dynamically adjusted based on use case. Would still love to see more about how these maps work.

recidiviz/persistence/persistence.py
```diff
@@ -99,25 +115,30 @@ def write(ingest_info, region, last_seen_time):
...
+            entity_matching.match_entities(session, region, people)
+            database.write_people(session, people)
+            session.commit()
+            persisted = True
+        except Exception:
+            session.rollback()
+            raise
+        finally:
+            session.close()
+            mtags[PERSISTED_KEY] = persisted
```
jessex: Just to be sure, this key won't show up on metrics unless `should_persist=true` is also a tag, right?

recidiviz/persistence/persistence.py
```diff
@@ -99,25 +115,30 @@ def write(ingest_info, region, last_seen_time):
...
-        entity_matching.match_entities(session, region, people)
-        database.write_people(session, people)
-        session.commit()
-    except Exception:
-        session.rollback()
-        raise
-    finally:
-        session.close()
+    mtags = {REGION_KEY: region, SHOULD_PERSIST_KEY: _should_persist()}
+    with monitoring.measurements(mtags) as measurements:
```
jessex: I see why we're using the measurement map `with` context here and that it required wrapping all of the logic of this method with an outer layer. But for more simple measurements where we just want to count something every time we reach a point in the code, will we need to also use a context-managed measurement map? I'm wondering about how lean the instrumentation code can be.

recidiviz/persistence/persistence.py
```diff
@@ -19,12 +19,28 @@
...
 from recidiviz.persistence.converter import converter
 from recidiviz.persistence.database import database
-from recidiviz.utils import environment
+from recidiviz.utils import environment, monitoring
+
+m_people = measure.MeasureInt("persistence/num_people",
+                              "The number of people persisted", "1")
+REGION_KEY = 'region'
+SHOULD_PERSIST_KEY = 'should_persist'
+PERSISTED_KEY = 'persisted'
```
jessex: I'm guessing we plan to extract these constants out to a shared location as our monitoring footprint grows, yeah? How much do we anticipate extracting out? Can these views and their registration live in simple methods somewhere? Should we have a `recidiviz.monitoring` package containing constants, views, and the logic currently in `recidiviz/utils/monitoring.py`?

jessex:
What is the story for testing monitoring/instrumentation code? We shouldn't see coverage drop just because we add monitoring calls around the codebase, but do we intend to mock those calls out or can we just let OpenCensus do its thing in memory? I'd also like to add test coverage for any custom monitoring logic itself.

recidiviz/persistence/persistence.py
arian487: I think the region should definitely not be up to the module owner to define.  Ideally we want some type of internal dashboard page on a per scraper level, and anyone that calls the monitoring library weill automatically have the same region key assigned to its metric.

Similarly, I'd like for opencensus logic to be fully abstracted away into our own library, in case we move away from opencensus.  The monitoring should expose some library for counting various metrics with an allowance for custom labels passed in.

I think for now this is fine as a proof of concept so we see some dashboards though.

jessex: Agreed in full on this comment.

colincadams:
I agree we should have some tests for the `monitoring.py` class itself.  I think as far as the rest of the code goes I would just let `opencensus` do its thing without actually exporting the views. For very complex metrics we could add some assertions that the correct increments were made or values were passed, but I don't expect this to be necessary too often.

recidiviz/persistence/persistence.py
colincadams: I moved the tag keys out to monitoring, and they could end up in their own file.

Personally, as these views and metrics are specific to what is happening in this file, I prefer having them with the code they are monitoring. I'd be okay with moving the `view` definition out of the file, but I don't see a clear advantage to doing that. I think then we will end up with a giant file of view definitions which feels clunkier than having them inline.

 I'd push back on fully abstracting away opencensus. To me the point of using opencensus is that we can switch between different monitoring solutions using the same library. I think it would be difficult to abstract it away without either (1) replicating the api in or own code or (2) losing some of the power of the api (i.e. have to pass in region label in every call).

I want to restrict the bloat that monitoring adds to any file so that it doesn't distract from the business logic, and to make it as easy as possible to add metrics to new code. I don't mind if they have to import `opencensus` specific code.

recidiviz/persistence/persistence.py
colincadams: Nope, I think if the control flow is simple (i.e. just one exit point), we could just call record manually at the end.

recidiviz/persistence/persistence.py
colincadams: Correct! The actual columns of the metric as defined right now are `region` and `persisted`, so `persisted` will be `None` in those cases. We could also include `should_persist` in the output metric and just have `persisted` be `False` whenever `should_persist` is.

recidiviz/utils/monitoring.py
colincadams: The real reason I use `@contextmanager` and yield the map, is to ensure that `mmap.record` is always called. Due to the return and exception handling there are multiple exit points and I thought it would be error prone to manually ensure record was called in each place.

recidiviz/utils/monitoring.py
colincadams: The current docs are very sparse and incomplete 😢 The main docs are on [this site](https://opencensus.io/) including a [python quickstart](https://opencensus.io/quickstart/python/metrics/) and [a more detailed walkthrough](https://opencensus.io/stats/), but there is plenty that is missing or out of date.

I've mostly been splunking around in [the github repo](https://github.com/census-instrumentation/opencensus-python), including looking at the readme, [some examples](https://github.com/census-instrumentation/opencensus-python/tree/master/examples/stats) and just the implementation itself.

I know this story isn't great right now, but it should be improving.

recidiviz/persistence/persistence.py
arian487: Your comment makes sense to me, and it would be helpful to understand the amount of work involved in say adding a new metric to each county that is represented as a graph on the dashboard end.  Ultimately to me that is the goal.  The 'users' are the people adding monitoring points and metrics which is just the internal components.  

I'm game if the number of lines and the ease of use on that end are good.

arian487:
I replied to the one comment I made, I'd like to get this in soon so we can get some dashboards up and running.  My goal is to have a couple of us (not you) add metrics to the various components of our pipelines to see how easy that would be and go from there.

colincadams:
Sounds good! Agreed we should deploy it and see what happens

PR #455, opened 2019-01-03
Log all the things

arian487:
First set of logs, still need to add logs for converter and data extractor, but this is a good start and will help in the immediate term.


recidiviz/ingest/base_scraper.py
```diff
@@ -171,6 +177,9 @@ def _generic_scrape(self, params):
                     'IngestInfo must be populated if there are no more tasks')
             scraper_start_time = common_utils.parse_datetime_string(
                 params['scraper_start_time'])
+            logging.info('Writing the ingest_info to the database for %s:' % self.region.region_code)
+            logging.info(ingest_info)
```
colincadams: This could generate some huge log messages, some of which could be over the limit and get dropped: https://cloud.google.com/logging/quotas

I know it is probably useful to log the whole thing in some cases, but do you think there is a summary that would give us most of what we need? I.e. person's name, number of bookings/charges/etc.?

recidiviz/ingest/base_scraper.py
```diff
@@ -177,9 +177,11 @@ def _generic_scrape(self, params):
                     'IngestInfo must be populated if there are no more tasks')
             scraper_start_time = common_utils.parse_datetime_string(
                 params['scraper_start_time'])
-            logging.info('Writing the ingest_info to the database for %s:' % self.region.region_code)
+            logging.info('Writing the ingest_info to the database for %s:' %
```
colincadams: nit: in general I think the style is to allow logging to do the formatting for you, instead of doing it yourself, i.e.:
`logging.info('some data: %s', data)`
instead of
`logging.info('some data: %s', % data)`

https://docs.python.org/3/howto/logging.html#logging-variable-data

recidiviz/ingest/base_scraper.py
colincadams: sounds good!

recidiviz/ingest/base_scraper.py
jessex: I echo this one. If this is still the case anywhere in the codebase, we should clean it up.

jessex:
A high-level comment after the merge: all of these new log statements are INFO-level but they don't all feel INFO-level to me. Was this a conscious choice or would you agree that some of these should be DEBUG-level?

arian487:
It was a conscious choice for now because python3 logging on GCP is really weird.  Colin and I haven't figured out how to group the logs cleanly together, or even have everyhting show up.  Sending them as info logs lets us easily find them because they show up.  Once we figure out why logging is so weird in GCP, we can better formalize the logs.

PR #460, opened 2019-01-03
Set queue token bucket limits to max of 500.

### LINKED ISSUE #459: Tasks are not running concurrently
Only one task is running at a time due to the `bucket_size` [parameter](https://cloud.google.com/appengine/docs/standard/python/config/queueref#syntax) being set to 1 for all queues in `queue.yaml`.

rasmi:
Fixes #459.


jessex:
We should also update the scraper generator to set bucket_size to 500 by default.

PR #462, opened 2019-01-03
Changed max_concurrent_requests to max of 5000.

### LINKED ISSUE #459: Tasks are not running concurrently

rasmi:
Fixes #459.


rasmi:
After this change, `rate` is the limiting factor on concurrent requests.

jessex:
Same comment here as on the other PR: we should update the scraper generator to make this the default value for new counties.

Issue #371, opened 2018-12-21
Convert ingest info _Charge.number_of_counts into multiple entity charges


terinpw:
currently ingest_info._Charge has field number_of_counts, and the entity/schema versions of charge do not (the assumption was that we would create separate charges for each count).

ohinds:
Update this google doc that references this issue when it's closed:

https://docs.google.com/document/d/16j-Ng8c4MGyKArDrTsyLjQ-MQfYbxTsoWWmZBBLG0iA/edit#

PR #467, opened 2019-01-03
Ingest info fields documentation

ohinds:
Add the complete field documentation for the ingest info object to the README in the ingest/models directory.


recidiviz/ingest/models/README.md
```diff
@@ -0,0 +1,417 @@
...
+</tr>
+<tr class="even">
+<td><strong>full_name</strong></td>
+<td><p>A person’s name.</p>
+<p>Use this for sites that provide names in a single field. Use surname and given_names for sites that provide them separately.</p></td>
+</tr>
+<tr class="odd">
+<td><strong>surname</strong></td>
+<td><p>A person’s surname.</p>
+<p>Use this for sites that provide surname and given names in separate fields. Use full_name for sites that provide names in a single field.</p></td>
```
arian487: Maybe be a bit more explicit about not setting surname/given_names AND fullname?

recidiviz/ingest/models/README.md
```diff
@@ -0,0 +1,417 @@
...
+</tr>
+<tr class="even">
+<td><strong>given_names</strong></td>
+<td><p>A person’s given names (including any middle names), separated by whitespace. If first and middle names are provided in separate fields, concatenate all given name fields into one string and put that string in this field.</p>
+<p>Use this for sites that provide surname and given names in separate fields. Use full_name for sites that provide names in a single field</p></td>
+</tr>
+<tr class="odd">
+<td><strong>birthdate</strong></td>
+<td><p>Date the person was born. The string must be parsable as a date by dateutil.parser.</p>
```
arian487: I'm not sure an implementation detail like what parser we are using belongs in here.  IMO that sort of thing should come from the errors that are raised by the converter.

recidiviz/ingest/models/README.md
```diff
@@ -0,0 +1,559 @@
+# ingest\_info Field Documentation
+
+Author: Oliver Hinds \<ohinds@vera.org\>
```
jessex: I don't think we want or need to list authors or dates (directly below) in versioned files on Github. That information is derivable.

recidiviz/ingest/models/README.md
```diff
@@ -0,0 +1,559 @@
...
+
+Version: 2019/01/03
+
+The ingest\_info object is how a scraper passes information scraped about an individual person into the system for storage. This is the documentation of the available data fields, which defines the data that can be input into the system.
```
jessex: I would remove direct references to scraping from this document. I think we want it to be general-purpose, i.e. this is the model we use for data transfer of all kinds.

recidiviz/ingest/models/README.md
```diff
@@ -0,0 +1,559 @@
...
+
+All data fields in the ingest\_info object are stored as strings. There is logic in the data conversion system that confirms whether each string can be converted into a valid value to be stored in the database. Certain fields are stored as enumerated types: the values of these fields are taken from a pre-specified set of values.
+
+## Person
+
+The ingest\_info object is just a container that holds a list of Person objects. Each Person holds details about the individual scraped from a website, as well as a list of Booking objects created from scraped data.
+
+<table>
```
jessex: Can we use Markdown instead of raw HTML? That will make the raw Markdown easier to read and easier to update. [See here](https://help.github.com/articles/organizing-information-with-tables/).

recidiviz/ingest/models/README.md
```diff
@@ -0,0 +1,559 @@
...
+</tr>
+<tr class="odd">
+<td><strong>surname</strong></td>
+<td><p>A person’s surname.</p>
+<p>Only use this for sites that provide surname and given names in separate fields. Use full_name for sites that provide names in a single field.</p></td>
+</tr>
+<tr class="even">
+<td><strong>given_names</strong></td>
+<td><p>A person’s given names (including any middle names), separated by whitespace. If first and middle names are provided in separate fields, concatenate all given name fields into one string and put that string in this field.</p>
+<p>Only use this for sites that provide surname and given names in separate fields. Use full_name for sites that provide names in a single field</p></td>
```
jessex: So as I mentioned above, I would like to remove direct references to scraping. But I would also rephrase sentences like this one, e.g. "Only use this for sites..." Again, I think we should make this into a data transfer spec as opposed to "here's how to structure the data you, the scraper writer, are accumulating." I think there's value in the latter, as well, but I think it should perhaps be a separate document.

recidiviz/ingest/models/README.md
```diff
@@ -0,0 +1,559 @@
...
+<td><p>Date the person was born.</p>
+<p>Use this for sites that provide it. When a person’s age but not birthdate is reported, use age instead.</p></td>
+</tr>
+<tr class="even">
+<td><strong>gender</strong></td>
+<td><p>A person’s gender.</p>
+<p>Valid values:</p>
+<ul>
+<li><blockquote>
+<p>Female</p>
```
jessex: I see that in these tables we use Titular Capitalization but link to Python code that is all-caps. We should note that are values are case-insensitive in general. Maybe a general _Notes_ section documenting noteworthy things that apply to all relevant fields?

recidiviz/ingest/models/README.md
```diff
@@ -0,0 +1,559 @@
...
+<td><strong>admission_date</strong></td>
+<td>The date this person was booked into this facility.</td>
+</tr>
+<tr class="odd">
+<td><strong>projected_release_date</strong></td>
+<td>The date this person is scheduled to be released on this booking.</td>
+</tr>
+<tr class="even">
+<td><strong>release_date</strong></td>
+<td>The date this person was released on this booking.</td>
```
jessex: It might be worth adding a bit of explanation about what it means if projected_release_date is after or before release_date. That might seem obvious but I suspect it's something future readers would appreciate being spoon-fed.

recidiviz/ingest/models/README.md
jessex: Another thing that could be included in a _Notes_ section: are all fields generally optional?

recidiviz/ingest/models/README.md
```diff
@@ -0,0 +1,559 @@
...
+
+The Arrest object represents information about the arrest that led to the associated booking.
+
+| **Field Name**    | **Description**                                                                         |
+| ----------------- | --------------------------------------------------------------------------------------- |
+| **arrest\_id**    | Unique identifier for an arrest. If not specified, one will be generated automatically. |
+| **date**          | The date this person was arrested on this booking.                                      |
+| **location**      | The location of arrest.                                                                 |
+| **officer\_name** | The name of the arresting officer.                                                      |
+| **officer\_id**   | The ID of the arresting officer (e.g., badge number)                                    |
```
jessex: Missing a period at the end of this one and also _agency_'s description.

recidiviz/ingest/models/README.md
```diff
@@ -0,0 +1,559 @@
...
+<p>Minimum</p>
+</blockquote></li>
+<li><blockquote>
+<p>Work Release</p>
+</blockquote></li>
+</ul>
+<p><a href="https://github.com/Recidiviz/recidiviz-data/blob/master/recidiviz/common/constants/booking.py#L63"><span class="underline">Complete list of valid values</span></a></p></td>
+</tr>
+<tr class="even">
+<td><strong>total_bond_amount</strong></td>
```
jessex: Adding this comment here for no real reason: is there a specific formula we're following for determining the order of fields in their respective tables?

recidiviz/ingest/models/README.md
```diff
@@ -0,0 +1,559 @@
...
+| **arrest\_id**    | Unique identifier for an arrest. If not specified, one will be generated automatically. |
+| **date**          | The date this person was arrested on this booking.                                      |
+| **location**      | The location of arrest.                                                                 |
+| **officer\_name** | The name of the arresting officer.                                                      |
+| **officer\_id**   | The ID of the arresting officer (e.g., badge number)                                    |
+| **agency**        | The arresting agency (e.g., police department, Sheriff’s office)                        |
+
+## Charge
+
+The Charge object holds information on a single charge.
```
jessex: It may be worth briefly describing the relationship between a booking and a set of charges, and any special cases. I'm wondering about things like how we model a booking with 4 counts of the same charge versus 1 count of 4 separate charges.

recidiviz/ingest/models/README.md
```diff
@@ -0,0 +1,559 @@
...
+</tr>
+</thead>
+<tbody>
+<tr class="odd">
+<td><strong>charge_id</strong></td>
+<td>Unique identifier for a charge. If not specified, one will be generated automatically.</td>
+</tr>
+<tr class="even">
+<td><strong>offense_date</strong></td>
+<td>The date of the (alleged) offense that led to this charge.</td>
```
jessex: Maybe a nit but I don't think we need to put "alleged" in parentheses. We should just say "alleged offense" to be more matter of fact about the fact that the person is not necessarily guilty.

recidiviz/ingest/models/README.md
```diff
@@ -0,0 +1,559 @@
...
+</blockquote></li>
+<li><blockquote>
+<p>Probation violation</p>
+</blockquote></li>
+</ul>
+<p><a href="https://github.com/Recidiviz/recidiviz-data/blob/master/recidiviz/common/constants/charge.py#L79"><span class="underline">Complete list of valid values</span></a></p></td>
+</tr>
+<tr class="even">
+<td><strong>level</strong></td>
+<td>Charge level (e.g. 1, 2, A, B, etc.)</td>
```
jessex: Missing a period at the end of this one.

recidiviz/ingest/models/README.md
```diff
@@ -0,0 +1,559 @@
...
+</tr>
+<tr class="odd">
+<td><strong>fee_dollars</strong></td>
+<td><p>Fee associated with this charge (e.g., booking fee, court fee).</p>
+<p>Note, this is different than a fine, which is imposed as part of a sentence.</p></td>
+</tr>
+<tr class="even">
+<td><strong>charging_entity</strong></td>
+<td>The entity that brought this charge (e.g., Boston Police Department, Southern District of New York)</td>
```
jessex: Missing a period at the end of this one.

recidiviz/ingest/models/README.md
```diff
@@ -0,0 +1,559 @@
...
+<tr class="even">
+<td><strong>sentence</strong></td>
+<td>A link to the Sentence object associated with this charge.</td>
+</tr>
+</tbody>
+</table>
+
+## Bond
+
+A Bond object holds information on a bond. A bond can be per charge, apply to multiple charges, or be a total bond across the whole booking.
```
jessex: We should briefly explain how we model each of these situations, i.e. what is the relationship between booking, charge, and bond in these different cases.

recidiviz/ingest/models/README.md
```diff
@@ -0,0 +1,559 @@
...
+<thead>
+<tr class="header">
+<th><strong>Field Name</strong></th>
+<th><strong>Description</strong></th>
+</tr>
+</thead>
+<tbody>
+<tr class="odd">
+<td><strong>bond_id</strong></td>
+<td>Unique identifier for a charge. If not specified, one will be generated automatically.</td>
```
jessex: What does this mean exactly? What does the generated id correspond to? Is it random?

recidiviz/ingest/models/README.md
```diff
@@ -0,0 +1,559 @@
...
+<td>Flag indicating whether parole is a possibility</td>
+</tr>
+<tr class="odd">
+<td><strong>post_release_supervision_length</strong></td>
+<td>Duration of community supervision to be served after release from incarceration.</td>
+</tr>
+</tbody>
+</table>
+
+## Generate a README.md from this document
```
jessex: What's the idea for this one? I personally don't think we should have a separate Word doc somewhere that is the source of truth that we periodically convert into Markdown and update here. We should just have this README that is the versioned source of truth.

recidiviz/ingest/models/README.md
ohinds: Unfortunately not. The GSM table format doesn't support multiline cells without writing the cell contents in html anyhow. And the process of exporting the google doc and converting is in any case much easier than maintaining that unwieldy combination of markdown and html, I think.

recidiviz/ingest/models/README.md
ohinds: Yup, that's leftover from when this was just a gdoc. Removed

recidiviz/ingest/models/README.md
```diff
@@ -0,0 +1,559 @@
...
+<td>A list of Charge objects associated with this booking.</td>
+</tr>
+</tbody>
+</table>
+
+## Arrest
+
+The Arrest object represents information about the arrest that led to the associated booking.
+
+| **Field Name**    | **Description**                                                                         |
```
zdg2102: Why is this table a different format than all the other tables?

recidiviz/ingest/models/README.md
ohinds: True. Done.

recidiviz/ingest/models/README.md
ohinds: Sure, done

recidiviz/ingest/models/README.md
ohinds: I added a Value format notes section at the top.

recidiviz/ingest/models/README.md
ohinds: Done.

recidiviz/ingest/models/README.md
ohinds: Done

recidiviz/ingest/models/README.md
ohinds: I just took them from the source files. I'm not sure? @arian487?

recidiviz/ingest/models/README.md
ohinds: Yes, but as above, I think the markdown file would be very difficult to maintain. I'm open to other ideas though, like a wiki...

recidiviz/ingest/models/README.md
ohinds: Because there are no multi-line cells. Then markdown can represent it, so the converter just uses the markdown table format

recidiviz/ingest/models/README.md
ohinds: The doc has a comment in it that when https://github.com/Recidiviz/recidiviz-data/issues/371 is resolved, we should update this. I will drop a note in the issue, as well.

recidiviz/ingest/models/README.md
ohinds: Done

recidiviz/ingest/models/README.md
ohinds: Done

recidiviz/ingest/models/README.md
ohinds: Done

recidiviz/ingest/models/README.md
```diff
@@ -0,0 +1,559 @@
+# ingest\_info Field Documentation
```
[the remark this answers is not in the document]
ohinds: :shrug:? The converter did it?

recidiviz/ingest/models/README.md
```diff
@@ -0,0 +1,559 @@
...
+
+Author: Oliver Hinds \<ohinds@vera.org\>
+
+Version: 2019/01/03
+
+The ingest\_info object is how a scraper passes information scraped about an individual person into the system for storage. This is the documentation of the available data fields, which defines the data that can be input into the system.
+
+All data fields in the ingest\_info object are stored as strings. There is logic in the data conversion system that confirms whether each string can be converted into a valid value to be stored in the database. Certain fields are stored as enumerated types: the values of these fields are taken from a pre-specified set of values.
```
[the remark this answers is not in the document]
ohinds: Done

recidiviz/ingest/models/README.md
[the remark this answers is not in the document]
ohinds: That's why I had "concepts" instead of "values" in the first place. It needs to be made clear somewhere that these are what we can represent, not what the string values need to be.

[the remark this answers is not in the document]
ohinds: I added a sentence to the 'Certain fields are stored' bullet under value format notes that tries to explain this. Also, it's unclear that people using this data transfer spec will have the ability to expand the string maps.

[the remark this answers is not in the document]
zdg2102: Got it, makes sense to me

recidiviz/ingest/models/README.md
zdg2102: I would be in favor of removing the "If not specified" sentence. If I understand correctly, `bond_id` here corresponds to `Bond.external_id` in our schema, which is never generated, while `Bond.bond_id` is a primary key and is always generated regardless of if an external ID is provided or not. So what type of ID we use internally seems irrelevant to someone passing data to ingest info.

recidiviz/ingest/models/README.md
jessex: I think a Markdown file is easier to maintain, in almost any case, than a Word doc that is just _somewhere_ and that needs to be run through a process native to someone's machine, with all of the installation and environmental and versioning issues that can bring. I strongly insist on not going a generated doc route like this. It has never gone well for me in the past.

A wiki that is kept close to the code, i.e. the Github wiki feature, seems reasonable for longer form documentation and I would be up for maintaining a wiki here on the repository for all forms of technical documentation in the long-term. The wiki here will use Markdown, though.

recidiviz/ingest/models/README.md
jessex: Looks great! 🎉

recidiviz/ingest/models/README.md
jessex: That's such a shame about the lack of multiline support! But can we go with Markdown-heavy and HTML-only-where-necessary, e.g. by structuring the table and its cells with Markdown but using something like `<br>` to make line breaks in a given cell? That would be much easier to maintain, I think.

jessex:
This looks great, @ohinds! The only outstanding remarks from me are about getting away from generating this from a Word/Google doc in favor of some doc here in the repository (either a README or a Wiki page) being the single source of truth, and a potential improvement for how we do tables.

recidiviz/ingest/models/README.md
ohinds: Done.

recidiviz/ingest/models/README.md
ohinds: Ah, cool. @zdg2102, are the other places this phrase 'if not specified' appears correct?

recidiviz/ingest/models/README.md
ohinds: rst to the rescue on here. Substantially better than either the doc or markdown solutions.

PR #468, opened 2019-01-03
add middle names

jamwalla:
#206
Adds a `middle_names` field to IngestInfo (probably we'll get initials, which could help match)


recidiviz/persistence/converter/person.py
```diff
@@ -30,7 +30,8 @@ def copy_fields_to_builder(person_builder, proto, metadata):
     new = person_builder
 
     new.external_id = fn(normalize, 'person_id', proto)
-    new.surname, new.given_names = _parse_name(proto)
+    # TODO(466): Replace with new.full_name
+    new.surname, new.given_names = _parse_name(proto), None
```
arian487: I don't like this stop gap change.  It breaks entity matching until zach gets the DB change in, and the relevant entity matching change is made.  I would rather you code this as if full name exists on the DB, and then submit this whenever the relevant change is made.

arian487:
I definitely agree with this approach assuming we are happy matching on a person who only has one of their names set.  I have the one comment though as submitting this now will break entity matching.

PR #471, opened 2019-01-03
Add types to entities class

zacharylawrence:


recidiviz/persistence/entities.py
```diff
@@ -20,115 +20,122 @@
...
-    birthdate_inferred_from_age = attr.ib()
-    gender = attr.ib()
-    race = attr.ib()
-    region = attr.ib()
-    ethnicity = attr.ib()
-    place_of_residence = attr.ib()
+    external_id: str = attr.ib()
+    surname: str = attr.ib()
+    given_names: str = attr.ib()
+    birthdate: datetime = attr.ib()
```
terinpw: date not datetime same with some below

zacharylawrence: Done.

PR #472, opened 2019-01-03
enable logging format pylint check

colincadams:
Wanted to open this more as a discussion, totally fine if either of you disagree and would prefer to be able to do it either way.

I like the idea of enforcing consistency across the code base, and thought this might be a good place to start.


arian487:
Thanks for doing this, I think I agree with your version.  I don't have strong opinions either way but the way you're doing it in this PR is the way Google does so I'm inclined to do that.  I'll approve but happy to defer to Josh if he feels strongly.

Issue #383, opened 2018-12-21
Add a retry limit to the tasks


arian487:
Currently if a task fails, it will retry indefinitely.  There should probably be a max retry limit enforced.  Its possible to have a specific type of request failing and it would fill up the queue and prevent any other tasks from running.  Alternatively we could enforce that a failed task go to the end of the queue and not the front.

PR #477, opened 2019-01-04
Retry limit set to 5

arian487:
This PR also updates the other fields that were changed in a previous PR so the bucket size is increased.


arian487:
This PR addresses https://github.com/Recidiviz/recidiviz-data/issues/383

PR #487, opened 2019-01-04
Fix vuln

arian487:
Fix the vulnerability


jessex:
Thanks a lot, @arian487.

Issue #465, opened 2019-01-03
us_mo_stone charges are not split


rasmi:
It seems that multiple charges in `us_mo_stone` are being coupled into a single Charge object, for example:

```
RECEIVING STOLEN PROPERTYLEAVE SCENE MV ACCIDENTDRIVERS LIC SUSPENDEDFAIL MOVE VEH OBSTR TRAFFIC
```

EDIT: this may be an issue with `BrooksJeffreyScraper`.

PR #494, opened 2019-01-04
Fix brooks

arian487:
Fixes https://github.com/Recidiviz/recidiviz-data/issues/465


arian487:
Note that there might be a separate conversation here on util methods that we expose in the data extractor, but for now just expose the one that I need for this.

jamwalla:
closed in favor of #503.

Issue #480, opened 2019-01-04
DataExtractor: Empty field creates invalid object


HalfdanJ:
Coming up in #329 

I have following (simplified) html

```
<table>
    <thead>
        <tr>
            <th>Charge Description</th>
            <th>Arresting Agency</th>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>FELONY BATTERY</td>
            <td></td>
        </tr>
    </tbody>
</table>
```

And mapping:

```
multi_key_mapping:
  Charge Description: charge.name
  Arresting Agency: arrest.agency
```

Arrest is not mentioned anywhere else in the mapping. The data extractor creates an arrest on the booking in this instance, but doesn't add any data to it (since arresting agency is empty), making it an invalid object according to the validator. 

I assume the correct behavior is the data extractor shouldn't create  the arrest if there is no data to add to it.

PR #495, opened 2019-01-04
prune IngestInfo objects

jamwalla:
#480 

- Adds `.prune()` method to IngestInfo objects with children
- DataExtractor prunes IngestInfo before returning
- Adds truthiness to all IngestInfo objects


recidiviz/ingest/extractor/data_extractor.py
```diff
@@ -167,7 +167,7 @@ def extract_and_populate_data(self, content, ingest_info=None):
             # TODO 183: actually have real warning codes
             warnings.warn("The following keys could not be found: %s" %
                           needed_keys)
-        return ingest_info
+        return ingest_info.prune()
```
arian487: I think this should be private?

jamwalla: I liked it, but everything's in a different class (DataExtractor -> IngestInfo -> Person -> etc.) so I get lots of `protected-access` errors from pylint. I think it's ok to expose these.

recidiviz/ingest/models/ingest_info.py
```diff
@@ -46,6 +49,10 @@ def get_recent_person(self):
             return self.person[-1]
         return None
 
+    def prune(self):
+        self.person = [p.prune() for p in self.person if p]
```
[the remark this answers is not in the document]
jamwalla: Totally agree - I added you to #512, thanks for the suggestion.

Issue #346, opened 2018-12-18
Standardize testing style


rasmi:
Our tests in the code have various styles and use different libraries. I think it would be nice to standardize them around [unittest](https://docs.python.org/3/library/unittest.html#basic-example).

Ideally:
* All test files `import unittest`.
* All tests classes inherit from `unittest.TestCase`.
* We use [`unittest`'s assertion methods](https://docs.python.org/3/library/unittest.html#assert-methods) rather than Python's `assert`.
* We use [`from unittest.mock import patch`](https://docs.python.org/3/library/unittest.mock.html) (python3 only) rather than `from mock import patch` .
* Test function names follow `lower_with_under` naming.
* [Setup and tear-down method names](https://docs.python.org/3/library/unittest.html#unittest.TestCase.setUp) follow: `setUp`, `tearDown`, `setUpClass`, `tearDownClass`.

Some approximate counts of the above behaviors:
```
grep -r 'import pytest' recidiviz/ | wc -l
# 14
grep -r 'from mock import patch' recidiviz/ | wc -l
# 11
grep -r 'assert ' recidiviz/tests | wc -l
# 540
grep -r 'class Test[a-zA-Z]*(object)' recidiviz/tests | wc -l
# 46
grep -r 'def setup_method' recidiviz/tests | wc -l
# 32
```
A lot of these instances are in `us_ny` and `us_vt` tests, which are in the process of being rewritten/removed (to rely on `BaseScraperTest`), but there are other instances sprinkled throughout the codebase.

Such `unittest`-based tests would be [run by pytest](https://docs.pytest.org/en/latest/unittest.html) as-is. Thoughts?

jessex:
This sounds like the right approach. My only nit for this plan is that I believe @zacharylawrence has made an argument in the past for test function names that combine snake_casing with camelCasing based on the hierarchy of test cases being flexed in a class. It would be good to explicitly agree on that point as a team.

zacharylawrence:
Awesome, thanks for putting this together Rasmi!

@jessex My only point was that I like tests that clearly outline 3 points of information and putting those in the name often makes tests super readable:
1. What is the behavior being tested
2. What is the situation that you want that behavior to be tested in
3. What is the expected result following execution of that behavior

I prefer camelCasing test names (as I often see in internal style) because it lets us clearly separate these 3 facts. For example: `testConvertBooking_bookingWithTotalBondAndNoCharges_createsNewChargeForInferredBond()`

[...]

PR #496, opened 2019-01-04
Remove pytest dependencies.

rasmi:
As part of #346, this removes all uses of `pytest` that aren't `@pytest.mark.usefixtures("emulator")`.


colincadams:
I think it's worth looking into the advantages and disadvantages of `pytest` over `unittest` before committing to this fully. I think if we decide that we can get away without any `pytest` features and should fully migrate to `unittest` then I'm down to do it.

If we can't then we should decide if we want to fully use `pytest`, or live in a particular hybrid world that we can enforce, i.e. all test classes inherit from `unittest.TestCase` and we use `self.assert` instead of plain `assert`.

colincadams:
I agree that unittest being a part of the standard library is nice and I don't like the magic of pytest collection. However there are some niceties like fixtures that I'm not sure we want to give up.

Of course there are a lot of opinions online about both 😁

rasmi:
Closing this unmerged, but hopefully some ambitious future engineer can help standardize the testing style!

PR #497, opened 2019-01-04
make names file optional

### LINKED ISSUE #169: names_file should be optional in regions.yaml
Currently the scraper will fail if it is not provided

colincadams:
Fixes #169.

I don't think this is necessarily what @ohinds was after but it seemed like an easy first step without getting rid of all of the `target_list` logic. WDYT?

AFAICT `us_ny` doesn't depend on the aardvark thing any more, it looks like it just bakes `'a'` into `_first_search_page_base_data`, so I deleted it entirely.


recidiviz/ingest/docket.py
```diff
@@ -169,13 +169,17 @@ def load_target_list(scrape_key, given_names="", surname=""):
     logging.info("Getting target list for scraper: %s", scrape_key)
 
     if scrape_key.scrape_type == constants.BACKGROUND_SCRAPE:
-        name_list_file = regions.Region(scrape_key.region_code).names_file
+        region = regions.Region(scrape_key.region_code)
+        if region.names_file:
```
ohinds: Should this be if region.names_file is not None:?

PR #498, opened 2019-01-04
Add test of consistency between master and historical tables

zdg2102:


recidiviz/tests/persistence/database/schema_test.py
```diff
@@ -0,0 +1,123 @@
...
+            if not exclusion in table_b_column_names:
+                raise ValueError(
+                    '{exclusion} is not a column in {table_name}'.format(
+                        exclusion=exclusion, table_name=table_b.__name__))
+            table_b_column_names.remove(exclusion)
+
+        table_a_column_names.sort()
+        table_b_column_names.sort()
+
+        self.assertEqual(table_a_column_names, table_b_column_names)
```
zacharylawrence: These are really sets right? Then you could assert equality without sorting first?

zdg2102: Done

PR #503, opened 2019-01-04
fix BrooksJeffrey charges

### LINKED ISSUE #465: us_mo_stone charges are not split

jamwalla:
fixes #465 as an alternative to #494. I would prefer to handle the underlying issue (line breaks disappear) and then work with the DataExtractor (use `keys_to_ignore` and hack around the output) to partially exposing the implementation. @arian487 - what do you think?

- Adds a static `DataExtractor._process_html` method that does preprocessing on html content. (We found in another scraper that was never merged ([fl_glades](https://github.com/Recidiviz/recidiviz-data/tree/fl_glades)) that we should remove javascript from the page, this seems like a good place to put it.)


jessex:
Agreed. Looks like a wise fix.

PR #508, opened 2019-01-05
Check for None in __eq__

### LINKED ISSUE #481: IngestInfo deepdiff doesnt work on None objects
The new DeepDiff 🎉#458 doesnt work if the actual value is None: 

```

    def __eq__(self, other):
>       return self.__dict__ == other.__dict__
E       AttributeError: 'NoneType' object has no attribute '__dict__'
```

since `other` is None. I tried adding `if not other: return False`, in all the `__eq__`, and that gives a nice output, but not sure thats the right place to solve it.

HalfdanJ:
This checks that `other` isn't None when comparing. Fixes #481


arian487:
Thanks!

jamwalla:
went ahead and made the change - thanks, @HalfdanJ !

PR #515, opened 2019-01-07
Fix pytest warning

zacharylawrence:


recidiviz/tests/common/buildable_attr_test.py
```diff
@@ -24,7 +24,7 @@
 
 
 @attr.s
-class TestBuildableAttr(BuildableAttr):
+class FakeBuildableAttrClass(BuildableAttr):
```
colincadams: any reason for `FakeBuildableAttrClass` instead of `FakeBuildableAttr`?

zacharylawrence: Yeah, I agree. Updated

PR #516, opened 2019-01-07
prevent proto->entity converter from maintaining generated ids

### LINKED ISSUE #509: Entity matching broken when id is generated
When we are using a generated id, that id should not be stored to the database but it is and matching fials on that.  Matching only works on ids that are not generated.

terinpw:
closes #509


recidiviz/persistence/converter/converter.py
```diff
@@ -49,7 +51,34 @@ def __init__(self, ingest_info, metadata):
         self.sentences = {s.sentence_id: s for s in ingest_info.sentences}
 
     def convert(self):
-        return [self._convert_person(p) for p in self.ingest_info.people]
+        converted_people = [self._convert_person(p) for p in
+                            self.ingest_info.people]
+        self._clear_generated_ids(converted_people)
+        return converted_people
+
+    def _clear_generated_ids(self, people: List[entities.Person]):
```
arian487: Correct me if I'm wrong but can't we just do a check in the entity converter code before assigning the external_id?  That shouldn't prevent us from processing the proto since the generated ids would still exist on the proto which is what helps us create the structure.  

Its not a huge deal but sometimes this list can be very long, admittedly its a difference of ms but I'd rather not introduce another conversion step if it can be avoided.

zacharylawrence: Yeah, I agree with @arian487 too. For example, something like:
```
new.external_id = fn(parse_external_id, 'person_id', proto)
```

terinpw: Great point, done

Issue #475, opened 2019-01-04
Improve validation errors


HalfdanJ:
When a validation error happens in the `base_scraper_test` (through `validate_and_return_populate_data`), it currently doesnt give a lot of hints at what is wrong. All I see is

```
        if errors:
>           raise ValidationError(errors)
E           recidiviz.persistence.validator.ValidationError

recidiviz/persistence/validator.py:48: ValidationError
```

It seems in source that the error does contain more information, might just be a matter of adding some error logging in `base_scraper_test`?

jessex:
Better logging would help, but I believe that some of the Stackdriver tools automatically extract error/exception messages and surface them in various ways, so we should make sure anything we raise has an appropriate message included.

jamwalla:
I believe the errors raised in the converter are clearer now, ie only `EnumParsingError` and `ValueError`. I'm going to close this but reopen if I'm mistaken!
See #1094

PR #525, opened 2019-01-07
Allow for region-specific enum mappings

### LINKED ISSUE #397: Investigate region-specific value --> enum maps
It would be nice to be able to specify a mapping between strings and enum values that only apply to a specific region without having to implement special case logic in `populate_data`.

### LINKED ISSUE #410: Increase the validations on BaseScraperTest
Right now it doesn't validate everything it can, and also doesn't provide nice util functions for a test writer.  It should at least run the converter on the ingest info object.

### LINKED ISSUE #414: Decide on a process for when the scraper writer needs to update the global map
Sometimes the map needs to be updated, or it needs to be handled manually in the scraper.  We should throw more helpful errors on when this happens and notify the user.

terinpw:
#475

Fixes #414
Fixes #410
Fixes #397


recidiviz/ingest/scraper_template/README.md
```diff
@@ -132,6 +132,23 @@ Optional fields are:
 
 Please be mindful to sleep a reasonable amount of time between each request, we don't want to bring down or degrade any websites!  This can of course run through the entire roster if you set the number of tasks to be high enough, but doing 5-10 is usually reasonable enough.
 
+Resolving Enum Parsing Errors
+==================
+Although we are populating all fields in `IngestInfo` with scraped strings, later several of those strings are converted into Enums. When running your scraper (either Unit Tests or End to End Tests), you may have encountered an `EnumParsingError: Could not parse X when building <enum Y>` during this process. This indicates that the scraped string could not be parsed into an Enum, in which case you have two options:
+
+### 1. Adding to the default map
+If you suspect the new string->Enum mapping should be shared across all scrapers, you should add it to the enum's default map. Enums with their default maps are in the `recidiviz/common/constants/` directory.
```
arian487: I would actually link to the directory here.

terinpw: Great point! done.

jessex:
Already merged but chiming in to say that this is a great feature. Thanks, @terinpw!

PR #540, opened 2019-01-08
support one-to-many mappings in DataExtractor

jamwalla:
for #329


recidiviz/ingest/extractor/data_extractor.py
```diff
@@ -156,13 +156,14 @@ def extract_and_populate_data(self, content, ingest_info=None):
...
-                    ingest_info, class_to_set, ingest_key, values)
-                if cell_val in needed_keys:
-                    needed_keys.remove(cell_val)
+                value_classes = key_mapping_val if isinstance(
+                    key_mapping_val, list) else [key_mapping_val]
+                for value_class in value_classes:
+                    class_to_set, ingest_key = value_class.split('.')
+                    self._set_or_create_object(
+                        ingest_info, class_to_set, ingest_key, values)
+                    if cell_val in needed_keys:
```
jessex: This doesn't remove the cell too early, i.e. before we can finish iteration over the multi-field value?

jamwalla: yikes, good catch!

PR #545, opened 2019-01-08
add type hints to IngestInfo

jamwalla:


zacharylawrence:
@jamwalla As discussed offline, we should consider making all of these types `Optional[]`

PR #546, opened 2019-01-08
Add ingest_info stub for typing

zacharylawrence:


jessex:
Now that we have two separate proto files, it seems more important to have some documentation explaining how to work with the proto files, e.g. how to generate them and when we should.

PR #556, opened 2019-01-09
Add raw enum text fields to entity objects

zdg2102:


zacharylawrence:
Sorry this PR slipped by me! I think you'll also need to add logic to the ingest_info.proto -> entities converter?

zacharylawrence:
Duplicate of #613

PR #557, opened 2019-01-09
Make entities.Charge use Builder

zacharylawrence:
#363


recidiviz/persistence/converter/converter.py
```diff
@@ -93,17 +93,20 @@ def _convert_booking(self, ingest_booking):
...
+        charge.copy_fields_to_builder(charge_builder, ingest_charge,
+                                      self.metadata)
+
+        charge_builder.bond = \
             fn(lambda i: bond.convert(self.bonds[i], self.metadata),
                'bond_id',
                ingest_charge)
-        new.sentence = \
+        charge_builder.sentence = \
             fn(lambda i: sentence.convert(self.sentences[i]),
```
jamwalla: should this use `self.metadata` or are we not there yet?

zacharylawrence: Don't think this one needs `self.metadata` since none of the fields in sentence need the data inside the `metadata`.

PR #569, opened 2019-01-09
Add override for CourtType

zacharylawrence:


jessex:
I am curious, does this eventually end with nearly every field being overridable? Do we have proper docs/guidance on when something should be overridable?

zacharylawrence:
Yeah I was thinking about this more last night and think we can just make
the overrides field required on the from_str() method to avoid running into
this again.

zacharylawrence:
@jessex I'm not sure what you mean by "every field being overridable"? The `enum_overrides` is a dict of str->Enum mappings that take precedence over the default mapping. The documentation for this lives [here](https://github.com/Recidiviz/recidiviz-data/blob/9231c5624120ddc48c06f41c538ffdbc5e978a90/recidiviz/ingest/scraper_template/README.md#resolving-enum-parsing-errors).

jessex:
Ah never mind. I was being obtuse. Ignore that question.

Issue #514, opened 2019-01-07
Bad max_length in sentence for us_ny


colincadams:
looks specific to `us_ny`

values:
- `LIFE Years, 99 Months, 99 Days`
- `Years, Months, Days`

sample sentence:
```
sentences {
  sentence_id: "23107127724688_generate"
  min_length: "0029 Years, 08 Months,\r\n 00 Days"
  max_length: "LIFE Years, 99 Months,\r\n 99 Days"
}
```

sample trace:
```
app/log_exception : Exception on /scraper/work [POST]
Traceback (most recent call last):
  File "/env/lib/python3.7/site-packages/flask/app.py", line 2292, in wsgi_app
    response = self.full_dispatch_request()
  File "/env/lib/python3.7/site-packages/flask/app.py", line 1815, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/env/lib/python3.7/site-packages/flask/app.py", line 1718, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/env/lib/python3.7/site-packages/flask/_compat.py", line 35, in reraise
    raise value
  File "/env/lib/python3.7/site-packages/flask/app.py", line 1813, in full_dispatch_request
    rv = self.dispatch_request()
  File "/env/lib/python3.7/site-packages/flask/app.py", line 1799, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/srv/recidiviz/utils/auth.py", line 98, in auth_and_call
    return func(*args, **kwargs)
  File "/srv/recidiviz/ingest/worker.py", line 89, in work
    result = scraper_task(params)
  File "/srv/recidiviz/ingest/base_scraper.py", line 189, in _generic_scrape
    ingest_info), metadata)
  File "/srv/recidiviz/persistence/persistence.py", line 102, in write
    people = converter.convert(ingest_info, metadata)
  File "/srv/recidiviz/persistence/converter/converter.py", line 34, in convert
    return _Converter(ingest_info, metadata).convert()
  File "/srv/recidiviz/persistence/converter/converter.py", line 51, in convert
    return [self._convert_person(p) for p in self.ingest_info.people]
  File "/srv/recidiviz/persistence/converter/converter.py", line 51, in <listcomp>
    return [self._convert_person(p) for p in self.ingest_info.people]
  File "/srv/recidiviz/persistence/converter/converter.py", line 62, in _convert_person
[...]

PR #581, opened 2019-01-10
UsNy: handle empty sentence lengths

jamwalla:
#514


recidiviz/ingest/us_ny/us_ny_scraper.py
```diff
@@ -329,8 +329,14 @@ def populate_data(self, content, params, ingest_info):
                           "sentence, as it should")
         sentence = sentence_info.person[0].booking[0].charge[0].sentence
 
-        # Handle special case of life sentences.
-        if sentence.max_length.upper().startswith('LIFE'):
+        # Handle empty sentence lengths.
+        empty_length = 'Years,  Months,\r\n          Days'
```
jessex: I wonder if it would be more resilient to make the check _if the string only features these three words after stripped of whitespace_ or maybe more simply _if the string is 'Years,Months,Days' once stripped of whitespace_?

PR #591, opened 2019-01-10
Use opencensus to trace all calls

### LINKED ISSUE #531: Trace calls to cloudsql
Currently requests to cloud sql aren't traced by default.
https://github.com/googleapis/google-cloud-go/issues/1085

Tracing them will help us debug any performance issues.

colincadams:
Appengine doesn't seem to be tracing our calls to client libraries (not expected) or cloudsql (this one is expected) by default. I've filed a bug with them for the client libraries ones. In the meantime we should just trace everything with opencensus.

Fixes #531


colincadams:
I ran this in staging to ensure it works.

arian487:
This change looks good, just wanted to make sure that its normal for pypl to be removed everywhere?

colincadams:
I don't know what that means in the lock file, but I've them be added and deleted in various PR's

recidiviz/server.py
```diff
@@ -55,3 +61,17 @@
...
+
+# Setup tracing of requests not traced by default
+if environment.in_prod():
+    exporter = stackdriver_exporter.StackdriverExporter(
+        project_id=metadata.project_id(), transport=BackgroundThreadTransport)
+else:
+    exporter = file_exporter.FileExporter(file_name='traces')
+# Note: This will work with the next release of `opencensus`.
+app.config['OPENCENSUS_TRACE_PARAMS'] = {
+    'BLACKLIST_HOSTNAMES': ['metadata']  # Don't trace metadata reqeusts
```
jessex: reqeusts -> requests

recidiviz/server.py
```diff
@@ -55,3 +61,17 @@
...
 if not environment.in_prod():
     app.register_blueprint(test_populator, url_prefix='/test_populator')
+
+# Setup tracing of requests not traced by default
+if environment.in_prod():
+    exporter = stackdriver_exporter.StackdriverExporter(
+        project_id=metadata.project_id(), transport=BackgroundThreadTransport)
+else:
+    exporter = file_exporter.FileExporter(file_name='traces')
+# Note: This will work with the next release of `opencensus`.
```
jessex: Maybe add a TODO on here with a reference to an issue to upgrade to the latest release? Otherwise, the next engineer to read it won't know if we're already on a working release or not.

recidiviz/server.py
colincadams: Fixed

recidiviz/server.py
colincadams: Done! The next release should be out v soon

PR #593, opened 2019-01-10
match on inferred birthdate

### LINKED ISSUE #351: Determine how to match people with incomplete birthdate information
Currently we only match a person if full birthdates + full names are provided. 

Do we want to match in any of the following cases?
- Inferred birthdates + full names
    - if we do this case -- what do we do when a person ages (their inferred birthdate changes).
- No birthdates + full names

terinpw:
During entity matching, if the entities.Person objects have inferred birthdates, consider the birthdates a match if they're within a year of each other.

close #351


terinpw:
closes #351

recidiviz/persistence/entity_matching_utils.py
```diff
@@ -35,21 +37,35 @@ def is_person_match(*, db_entity: entities.Person,
...
+        return _is_inferred_birthdate_match(a.birthdate, b.birthdate)
+
+    if not a.birthdate_inferred_from_age \
+            and not b.birthdate_inferred_from_age:
+        return a.birthdate == b.birthdate
+
+    return False
+
+
+def _is_inferred_birthdate_match(a: datetime.date, b: datetime.date) -> bool:
```
zacharylawrence: Nit: Could you add a quick docstring about why this is how we consider inferred_birthdates to match? Something along the lines of how we infer a date from Age?

recidiviz/persistence/entity_matching_utils.py
```diff
@@ -35,21 +37,35 @@ def is_person_match(*, db_entity: entities.Person,
...
+    return db_entity.given_names == ingested_entity.given_names \
+           and db_entity.surname == ingested_entity.surname \
+           and _is_birthdate_match(db_entity, ingested_entity)
+
+
+def _is_birthdate_match(a: entities.Person, b: entities.Person) -> bool:
+    if a.birthdate_inferred_from_age and b.birthdate_inferred_from_age:
+        return _is_inferred_birthdate_match(a.birthdate, b.birthdate)
+
+    if not a.birthdate_inferred_from_age \
```
arian487: Is this if statement necessary?  You're essentially asking if one of the entities was not inferred right?  That would be true from the previous if statement failing.

PR #602, opened 2019-01-11
add Entity type

jamwalla:


recidiviz/persistence/entities.py
```diff
@@ -34,8 +34,12 @@
 from recidiviz.common.constants.person import Race, Ethnicity, Gender
 
 
+class Entity(BuildableAttr):
```
zacharylawrence: Since you shouldn't be allowed to instantiate this object, it should probably be Abstract? Although I don't think the `abc` library will work here since the concrete classes extending `Entity` don't have any abstract methods that they need to override. Maybe something like [this](https://github.com/Recidiviz/recidiviz-data/blob/1b972a705122deee3a5cce78338922e87992d8e9/recidiviz/common/buildable_attr.py#L26) would work?

Also, does it make sense for each class below to extend both Entity and BuildableAttr or just have Entity extend BuildableAttr itself? I'm not sure what makes the most sense, but right now a builder error will raise a `Person.BuilderException` and I think it'd be nice to make sure we maintain that functionality (which I think maybe the way you've structured the code now already does?)

jamwalla: Cool, made `Entity` abstract using the approach from BuildableAttr.

This shouldn't change the type of the `BuilderException`. I think the inheritance chain is easier to read/deal with than using multiple inheritance, but do you think it makes more sense to decouple BuildableAttr and Entity here?

PR #603, opened 2019-01-11
Update teh cron

arian487:
Update the cron to include the new task.


arian487:
Thats a good question, one of hte things on my docket for next week is to figure out appengine pricing and how the instances work.  It may be beneficial for us to start them all at the same time to have less instance up time?  I'll look into.

PR #609, opened 2019-01-14
Fail and log on any non-200 status code.

ohinds:


recidiviz/ingest/scraper.py
```diff
@@ -277,9 +277,9 @@ def fetch_page(url, post_data=None, json_data=None):
             return -1
 
         # Proxy errors don't raise an exception. : (
-        if page.status_code == 502:
-            logging.warning("Problem retrieving page (proxy error), failing "
-                            "task to retry.")
+        if page.status_code != 200:
```
colincadams: looks like we could just call `page.raise_for_status()` and in our try except and it would raise an `HTTPError` for us, if that is preferred. This would give us the status code, reason and url, and handle decoding the reason for us.

recidiviz/ingest/scraper.py
```diff
@@ -248,7 +248,8 @@ def fetch_page(url, headers=None, post_data=None, json_data=None):
             else:
                 page = requests.post(url, proxies=proxies, headers=headers,
                                      data=post_data, json=json_data)
-        except requests.exceptions.RequestException as ce:
+            page.raise_for_status()
```
colincadams: Looking at the [requests docs](http://docs.python-requests.org/en/master/user/quickstart/#response-status-codes) this raises a `requests.exceptions.HTTPError`, which is a subclass of `requests.exceptions.RequestException`. Which means I think we can just keep the original `except` clause.

recidiviz/tests/ingest/scraper_test.py
```diff
@@ -463,22 +464,35 @@ def test_fetch_page_error(self, mock_region, mock_proxies, mock_headers,
...
         error_response.headers = {}
-        exception_response = requests.exceptions.RequestException(
-            request=original_request, response=error_response)
-        mock_requests.side_effect = exception_response
+        for code, name in errors.items():
+            err = HTTPError(url, code, name, {}, None)
 
-        scraper = FakeScraper(region, initial_task)
-        assert scraper.fetch_page(url) == -1
+            setattr(err, 'request', original_request)
```
colincadams: Is there a reason we need to use setattr instead of just a dot operator?

recidiviz/tests/ingest/scraper_test.py
```diff
@@ -463,22 +464,35 @@ def test_fetch_page_error(self, mock_region, mock_proxies, mock_headers,
...
 
-        mock_region.assert_called_with(region)
-        mock_proxies.assert_called_with()
-        mock_headers.assert_called_with()
-        mock_requests.assert_called_with(url, proxies=proxies, headers=headers)
+            error_response.status_code = code
+            error_response.reason = name
+            setattr(err, 'response', error_response)
+
+            mock_requests.side_effect = err
```
colincadams: Instead of giving it a side effect exception, if we just have it return a response with the error code then an exception will get created by response.raise_for_status right?

Issue #587, opened 2019-01-10
Write raw strings to the database


arian487:
Currently there are no raw strings being written since they are dropped by the converter.  Need to be writing all raw strings that get converted to ENUMs so we know what the original value was.

PR #613, opened 2019-01-14
Use add raw_text to db write

arian487:
Fixes https://github.com/Recidiviz/recidiviz-data/issues/587

Write the raw strings to database in addition to the enum.


recidiviz/persistence/converter/bond.py
```diff
@@ -35,10 +35,12 @@ def convert(proto, metadata: IngestMetadata) -> entities.Bond:
     # information about the type (i.e. NO_BOND or BOND_DENIED)
     new.bond_type = fn(BondType.from_str, 'bond_type', proto,
                        metadata.enum_overrides, default=inferred_bond_type)
+    new.bond_type_raw_text = fn(normalize, 'bond_type', proto)
```
zacharylawrence: I think this is fine, but I just wanted to note that `normalize` will leave the UpperCase/LowerCase that we parse off the website. We could consider storing all `raw_text` fields as UPPER_CASE but I'm not sure if that's actually what we want.

PR #626, opened 2019-01-15
fixes for us_pa_dauphin

jamwalla:
updates UsPaDauphinScraper to use the new task params


jamwalla:
cc @colincadams, here's a change that I made to the line we were just talking about, but I think this issue will go away with a "scrape_data_or_more" behavior

Issue #71, opened 2018-05-02
Define and implement a deploy and release strategy


jessex:
Today, we do not create releases of the data platform. Further, we do deploys ad hoc from our local machines with the gcloud cli. These are both bad practice and likely to cause confusion and//or production issues if left unaddressed for long.

We want to implement continuous release, ideally tied to our continuous integration pipeline so that we are creating new releases on every merge into master.

We want to implement a sane deploy strategy. This may be continuous or otherwise as approved by the time, and it may be performed by hand or automatically as approved by the team.

jessex:
Here's an outline of the maximally continuous approach:
* Cut a new release with every new merge to master
* Deploy each such release to production immediately
* Discourage deploys from local machines by monitoring Google Cloud activity logs

We should closely examine the impact of app engine deploys while, say, ingest is in progress. I presume it would behave the same as what happens if our MapReduce pipeline from calculation is in progress, which is that any new instances stood up after the deploy is completed are using the new code, which means it is conceivable that simultaneous live workers for an ingest or calculation pipeline could be running different code. We need to decide if that's acceptable, or if we need a more controlled deploy cadence.

jessex:
For the record, Travis supports auto-deploy to Google App Engine: https://docs.travis-ci.com/user/deployment/google-app-engine/

PR #627, opened 2019-01-15
Deploy release

arian487:
Define a deploy and release strategy.  This fixes https://github.com/Recidiviz/recidiviz-data/issues/71


.travis.yml
```diff
@@ -1,27 +1,60 @@
...
-- export CLOUD_SDK_REPO="cloud-sdk-$(lsb_release -c -s)" && 
-    echo "deb http://packages.cloud.google.com/apt $CLOUD_SDK_REPO main" | sudo tee -a /etc/apt/sources.list.d/google-cloud-sdk.list && 
-    curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | sudo apt-key add - && 
-    sudo apt-get update -y && sudo apt-get install google-cloud-sdk -y &&
-    sudo apt-get install google-cloud-sdk-datastore-emulator -y &&
-    sudo apt-get install google-cloud-sdk-pubsub-emulator -y
+- openssl aes-256-cbc -K $encrypted_f7b0b16e1068_key -iv $encrypted_f7b0b16e1068_iv
+  -in secrets.tar.enc -out secrets.tar -d
+- tar xvf secrets.tar
+- export CLOUD_SDK_REPO="cloud-sdk-$(lsb_release -c -s)" && echo "deb http://packages.cloud.google.com/apt
```
colincadams: nit: I know the lines were long, but I found this section much more readable before with each command on a single line.

If we want to maintain a max line length, then I think we should still start each following command on it's own line. We could just make each one it's own yaml entry if that is preferred

arian487: Oh I didn't even do that, looks like the travis script automatically did that.

arian487: I've fixed it now.

README.md
```diff
@@ -116,19 +116,38 @@ described in [running tests](#running-tests)). In practice, this is not particul
...
+#### On Monday Morning
+
+1. Verify that the tests are all passing in [Travis](https://travis-ci.org/Recidiviz/recidiviz-data)
+2. Merge branch `master` into `staging`.  This will deploy to our staging environment and accordingly promote the scrapers to cron that are
+set to be deployed later in the week.
+3. Run `https://recidiviz-staging.appspot.com/scraper/start?region=us_fl_martin` [TODO #623](https://github.com/Recidiviz/recidiviz-data/issues/623)
+and verify that it is happy by looking at the monitoring page [TODO #59](https://github.com/Recidiviz/recidiviz-data/issues/59) and also checking
+the logs for errors.
+4. For the next two days periodically check to make sure the build is happy and monitoring is good for all regions
+including the test region.  Pay special attention to the regions that are going to be promoted to prod
```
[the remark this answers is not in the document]
arian487: We may have some regions that are fake/constant in staging for testing purposes and some regions that have recently had scrapers written for them that go through the short trial period before being moved from the staging cron to the main cron.

README.md
```diff
@@ -116,19 +116,38 @@ described in [running tests](#running-tests)). In practice, this is not particul
...
+that week will deploy it correctly to staging and production.
+
+
+The release engineer oncall should go through the following steps:
+
+#### On Monday Morning
+
+1. Verify that the tests are all passing in [Travis](https://travis-ci.org/Recidiviz/recidiviz-data)
+2. Merge branch `master` into `staging`.  This will deploy to our staging environment and accordingly promote the scrapers to cron that are
+set to be deployed later in the week.
```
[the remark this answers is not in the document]
arian487: Its just markdown, there is no line break in the actual doc.

README.md
```diff
@@ -116,19 +116,38 @@ described in [running tests](#running-tests)). In practice, this is not particul
 Tasks emulator at this time. The [appengine documentation]( https://cloud.google.com/appengine/docs/standard/python3/testing-and-deploying-your-app)
 has more information about running locally.
 
-#### Deployed
-To deploy to a live AppEngine project, navigate to the directory where you cloned `recidiviz-data` and run
-`gcloud app deploy`. This will upload the full project to the cloud and enable it as the current version in your environment.
+### Deployment
+
+### Depoying a scraper
```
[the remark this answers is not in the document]
arian487: Fixed.

.travis.yml
```diff
@@ -1,27 +1,60 @@
...
+  project: recidiviz-staging
+  on: staging
+  config: queue.yaml
+- provider: gae
+  keyfile: client-secret-staging.json
+  project: recidiviz-staging
+  on: staging
 notifications:
   slack:
-    secure: RRx5Klgwfr2ggUxugs2nBOt2atYoKDrv+RfVY2OeINsZqPeZoxpfKntpIctfdUfPIZofyNUfPFfIFjZdNWBwXHKTu1/3KkCslBDUoPW9En1np5RTrVUmZJJy7no+J19PuDUDQYcDRCcEw41PbjcE7NPfK12bTI7jCGKw3ZTX1tLWlXDz70BOIsJxW3gLrgopFo0mNGWKbIVBU4VeUxhOpxB3RzegtCNrTIcfiEpnuNcyNvzhTpzlQdinW4rXQjv5p7oxCEkMGTfywtVrUVLfrSMCCVde1rv2RUBJPmShmFbJjfWuozMMlVK82vXMjj+z3HrIYPAYMmbVM9zqn/SDjdwzfzbmldol8yKcW4DTmlH4phfw9qWGa/UdpcHPT5vUoNzg3QVvVTYwV8sG8O6XeO341BSelSdfmbFq6YJvgfgHgDXbjbP8ME2DqUP/0RE21FEXReaV8Ln3FvneMFK1MnKV3czKOOwh6m48M7zFcdQr1OFto2+aIdF8IS7nke3btzUW1HL7ppn+wlLdOGsYuTPsMUQ/ykVSMTvB9Iolj083mtGm/QOu316JKgZMbH82e9p6FbXil44i1UGc8H2j9RVri4U0m2sEps4MnHbuBRhejzh6fO4k7pbAKTn35sH7TEgfacCIzBOI8t3crPB1AaK8fmMGEiZo07WalWT/MS0=
```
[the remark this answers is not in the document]
arian487: Thats weird, the travis sceript took that out...I'll add it back in.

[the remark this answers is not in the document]
arian487: Fixed.

README.md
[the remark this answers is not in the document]
jessex: That's weird. The source for the file definitely looks like we accidentally hit enter between "are" and "set". It renders fine, but it doesn't need to be sent down to line 132.

README.md
```diff
@@ -116,19 +116,38 @@ described in [running tests](#running-tests)). In practice, this is not particul
 Tasks emulator at this time. The [appengine documentation]( https://cloud.google.com/appengine/docs/standard/python3/testing-and-deploying-your-app)
 has more information about running locally.
 
-#### Deployed
-To deploy to a live AppEngine project, navigate to the directory where you cloned `recidiviz-data` and run
-`gcloud app deploy`. This will upload the full project to the cloud and enable it as the current version in your environment.
+### Deployment
+
+### Deploying a scraper
+To deploy a scraper, simply edit the [staging cron](staging/cron.yaml).  The release engineer
```
jessex: It's not 100% explicit, but it seems implied that this process involves checking in a change to `staging/cron.yaml` to add a new scraper, and then checking in a change to remove that scraper from `staging/cron.yaml` and then add it to `cron.yaml`. Is that right? If so, it feels like a good amount of churn.

I wonder if there's a way to change the branch structure to avoid so much churn directly on master. We check the first change into master, merge it into staging, then check another change directly into staging, then merge that into production. This means that production and staging do not look exactly like master but behind by some number of commits. It's also not clear how the changes find their way from production back to master without yet another merge, which seems like a potential danger. They both include commits -- to move scrapers from one `cron.yaml` to another -- that were _never_ written to master. I think if we are going to move towards dedicated release branches, we should not have engineers manually pushing commits to those release branches that never go through master. I think that makes the release workflow a bit hairier than it should be, reduces the reliability of our CI pipeline on master by ensuring it is not _exactly_ what's in production (even if only a small amount), and increases the number of potential bugs by increasing inconsistency.

I think a better approach might be to consolidate down to a single cron.yaml file but find the ability to suppress all but a whitelisted set of entries in that cron on stage. If that hypothetical whitelist is maintained in code then we have to figure out how to not have exactly the same set of concerns, but maybe that whitelist could be managed through some direct API call to GCP performed on deploy through Travis? One way would be to pause/disable the _queues_ that are available through the Cloud Task Queue API and let the crons all run but immediately no-op. That would be a bit noisy in the logs, admittedly. That would be trading off operational complexity for a cleaner commit log and branch structure. There could be better approaches with a more obviously positive trade.

README.md
```diff
@@ -116,19 +116,38 @@ described in [running tests](#running-tests)). In practice, this is not particul
...
+and verify in the mornings that their last scrape session had no errors.  If there are errors try to fix them or
+contact the scraper writer to fix them.
+
+#### On Wednesday Morning
+
+1.  Verify that the scraper crons as well as the `infer_release` cron ran successfully
+over the last two days.  Check the monitoring page to see if anything is fishy.
+2.  If all looks good, be sure to move the successful scrapers from [staging cron](staging/cron.yaml) to
+the [prod cron](cron.yaml) file.  Don't move the test cron as we do not want that running prod.  This is an important step, we do not want to run the jobs in two places!
+3.  Push that change to the `staging` branch
```
jessex: Missing a period here.

README.md
```diff
@@ -116,19 +116,38 @@ described in [running tests](#running-tests)). In practice, this is not particul
...
+### Deploying a scraper
+To deploy a scraper, simply edit the [staging cron](staging/cron.yaml).  The release engineer
+that week will deploy it correctly to staging and production.
+
+
+The release engineer oncall should go through the following steps:
+
+#### On Monday Morning
+
+1. Verify that the tests are all passing in [Travis](https://travis-ci.org/Recidiviz/recidiviz-data)
```
jessex: Missing a period here.

README.md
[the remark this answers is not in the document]
arian487: I see it now, fixed.

README.md
arian487: Fixed.

README.md
arian487: Done

README.md
arian487: Hm I definitely see the concerns you're bringing up.  I'm not in favour personally of using the Queue method because I think it can make the logs a bit confusing.  Someone might be trying to debug a region only to see that it didn't run, and then be confused why it didn't run.  I like the idea of letting the crons handle it...let me think on this and get back to you tomorrow.

colincadams:
I haven't read through the whole conversation, but can we use tags instead of branches create versions for release. Then we don't have to deal with the merging back and forth?

arian487:
I don't think travis supports deploys from tags, so we have to use branches.

README.md
arian487: Ok I may have just mentally convinced myself not to flow the mid week release out from master.  That would actually be quite dangerous as it would introduce unverified code into prod.  I think we would have to put into staging directly if we wanted to respect the flow.  I think that wouldn't be so horrible even if we didn't merge back from master as long as ensure the only changes that happen directly on the release branches is the cron changes.

jessex: Almost every release strategy I have experience with in Git goes roughly like this: 1) cut a new tag/release when tested and ready, 2) deploy that release to staging and test and monitor, then 3) deploy that release to production. It's how Git is designed to work, if I can be so bold (obviously Git can work in a variety of ways, but this is the natural path). 

It seems like the two reasons that's not the current suggestion, as far as I can tell, are because we're not sure Travis can deploy from tags and we are trying to ensure that staging only runs a very limited but dynamic set of scrapers on schedule. 

The first one should be solveable with elbow grease. The second one should be mitigatable via tooling. For one, releases that don't involve new scrapers should have none of the complexity discussed here and should have a simpler strategy. For two, I still think there should be a way to get the testing we want without jumping through what seems to me like a counterintuitive release flow. Do we need to automatically schedule each new scraper, or are manual runs okay? If automatic is required, it's only for a couple of days, so do we need to use hard coded yaml for staging or is there a different mechanism? For example, could we put together an API call that is triggered in stage that specifically finds new scrapers since the last release and runs them? That specific idea may not be great, but I think the bigger point stands.

arian487: Ok I just mentally convinced myself that is a bad idea since it introduces unverified commits to staging and then prod.  One solution is to remove the cron deploy process from the general release process and have the release engineer manually deploy that.  It would be the same process of updating master and pushing those to the respective branches, except the engineer will have to run a deploy script in master on mondays and wednesdays which deploys the correct cron job.

That way any actual code is still running on the correct versions.  Thoughts?

arian487: Automatic runs are definitely not required, its just a bit more work on the part of the release engineer who has to know to manually run the new regions.  Its a bit more error prone and less guaranteed.  

As for the API call I'm not sure how it would work exactly?  What do we want the scrapers writers to do when wanting a region to be released?  If they just update the cron and nothing else I think it'll be pretty hard to programmatically compare that to the older version and see which regions were added?  I don't know off the top of my head if you can easily say 'compare the contents of this file from last weeks version'

arian487:
I've updated the process and documentation as per our discussion @jessex.  I threw together an extremely simply deploy script that the release engineer can run but I'm happy to take suggestions on how that should be structured.

README.md
```diff
@@ -116,19 +116,33 @@ described in [running tests](#running-tests)). In practice, this is not particul
...
+### Deploying a scraper
+To deploy a scraper, simply edit the [cron](./cron.yaml).  The release engineer
+that week should check for all of the newly added scrapers and manually test them.
+
+The release engineer oncall should go through the following steps:
+
+#### On Monday Morning
+
+1. Verify that the tests are all passing in [Travis](https://travis-ci.org/Recidiviz/recidiviz-data).
+2. The release engineer should tag a commit with "va.b.c".  For now just increment the numbers until a better method is in place [TODO #646](https://github.com/Recidiviz/recidiviz-data/issues/646) This will trigger a release to staging.
```
jessex: Need a period after the TODO.

README.md
```diff
@@ -116,19 +116,33 @@ described in [running tests](#running-tests)). In practice, this is not particul
...
+3. Once the release is complete, run `https://recidiviz-staging.appspot.com/scraper/start?region=us_fl_martin` [TODO #623](https://github.com/Recidiviz/recidiviz-data/issues/623)
+and verify that it is happy by looking at the monitoring page [TODO #59](https://github.com/Recidiviz/recidiviz-data/issues/59) and also checking
+the logs for errors.
```
jessex: I think you can push this back to the end of the previous line.

README.md
```diff
@@ -116,19 +116,33 @@ described in [running tests](#running-tests)). In practice, this is not particul
...
+4. Manually check to see which set of scrapers have been added to the cron.yaml since last release and manually run all of those scrapers as well.
+5. Every morning until Wednesday, rerun all of the to be deployed scrapers manually and make sure there aren't any failures.
+6. For the next two days periodically check to make sure the build is happy and monitoring is good for all of the regions.  If there are errors try to fix them or
+contact the scraper writer to fix them.
```
jessex: This can go back to the end of the previous line, too.

README.md
```diff
@@ -116,19 +116,33 @@ described in [running tests](#running-tests)). In practice, this is not particul
...
+
+#### On Wednesday Morning
+
+1.  Verify that the scraper crons as well as the `infer_release` cron ran successfully
```
jessex: "...ran successfully _in staging_ over the last two days..."

README.md
```diff
@@ -116,19 +116,33 @@ described in [running tests](#running-tests)). In practice, this is not particul
...
+over the last two days.  Check the monitoring page to see if anything is fishy.
+3.  Once you are ready to release to production, run `./deploy_production <release_tag>` to finally release the code to production.
```
jessex: Maybe append this number with, "This will checkout recent tags and deploy the most recent one automatically."

README.md
```diff
@@ -116,19 +116,33 @@ described in [running tests](#running-tests)). In practice, this is not particul
...
+
+### Deploying a scraper
+To deploy a scraper, simply edit the [cron](./cron.yaml).  The release engineer
+that week should check for all of the newly added scrapers and manually test them.
+
+The release engineer oncall should go through the following steps:
+
+#### On Monday Morning
+
+1. Verify that the tests are all passing in [Travis](https://travis-ci.org/Recidiviz/recidiviz-data).
```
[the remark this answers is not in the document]
jessex: That will help because the second list looks mis-numbered as is.

deploy_production.sh
```diff
@@ -0,0 +1,16 @@
...
+echo "Fetching all tags"
+git fetch --all --tags --prune
+
+echo "Checking out tag $1"
+git checkout tags/$1 -b $1
+
+echo "Starting deploy of cron.yaml"
+gcloud app deploy cron.yaml --project=recidiviz-123
+
+echo "Starting deploy of queues.yaml"
```
jessex: queue.yaml, singular

deploy_production.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
```
jessex: I know that we're going with auto-deploys to stage and manual deploys to prod for now, but I could see it being useful to toggle this script to manually deploy to stage, as well. If not necessary yet, we could file that away as a future enhancement.

recidiviz/staging/cron.yaml
```diff
@@ -0,0 +1,33 @@
...
+  url: /infer_release/release
+  schedule: every day 09:30
+  timezone: America/New_York
+  retry_parameters:
+    min_backoff_seconds: 2.5
+    max_doublings: 5
+    job_age_limit: 9h
+
+
+# Scrapers to be promoted to prod
```
jessex: Can remove this line safely now.

README.md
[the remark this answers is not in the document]
arian487: Oh, thats an excellent protip, thanks!  Done.

README.md
arian487: Done.

README.md
arian487: Done.

README.md
arian487: Done.

README.md
arian487: Done.

README.md
arian487: I updated the comment, but the script as is does not deploy the most recent one, I figured the user should have to explicitly provide a release tag.

deploy_production.sh
arian487: Done.

arian487: Filed an issue: https://github.com/Recidiviz/recidiviz-data/issues/649

I'll say that I think its risky to have the same script handle both staging and production deploy.  If we guard it with a flag for example, the user might have intended to do staging but accidentally do production.  Personally I think its safer to have a strong separation by just having them be different scripts so the user has to very deliberately run a deploy production script vs a deploy staging script.

deploy_production.sh
arian487: Filed an issue: #649

I'll say that I think its risky to have the same script handle both staging and production deploy. If we guard it with a flag for example, the user might have intended to do staging but accidentally do production. Personally I think its safer to have a strong separation by just having them be different scripts so the user has to very deliberately run a deploy production script vs a deploy staging script.

recidiviz/staging/cron.yaml
arian487: Done.

arian487:
I've addressed the comments, thanks for taking a look.  Once its in, I'll run through the release process myself (I'll do it over two hours instead of two days), to make sure it makes sense and works.  We'll have it in place for next weeks release.

recidiviz/staging/cron.yaml
```diff
@@ -0,0 +1,30 @@
+# LEAVE THESE HERE, THESE ARE TEST SCRAPERS AND JOBS FOR STAGING.  We are
```
colincadams: This file is no longer necessary correct?

.travis.yml
```diff
@@ -1,26 +1,44 @@
...
+  test-project > ds_emulator.out 2> ds_emulator.err &
+- gcloud beta emulators pubsub start --project test-project > ps_emulator.out 2> ps_emulator.err
+  &
 script:
 - pylint *.py recidiviz
 - pytest recidiviz/tests/ --with-emulator
 - mypy recidiviz
 after_success:
 - coveralls
+deploy:
```
colincadams: @zdg2102 Is there a way to make this run alembic migrations as well?

deploy_production.sh
```diff
@@ -0,0 +1,16 @@
...
+
+echo "Fetching all tags"
+git fetch --all --tags --prune
+
+echo "Checking out tag $1"
+git checkout tags/$1 -b $1
+
```
colincadams: @zdg2102 @arian487 and run alembic here as well?

recidiviz/staging/cron.yaml
arian487: I think we should keep it since the cron for staging and prod is different, and we should have a file for it.

deploy_production.sh
arian487: We can think about how alembic fits into this a bit later I hadn't thought about that.

arian487:
My thoughts on more frequent releases is that I think the end state will definitely not have more releases, and in this phase of very active development, I'm happy running releases piecemeal like we have been now (following this process still).  I don't want to add too much process while we have a lot of work to do.

arian487:
Gentle nudge on this PR, I'd like to get it in to try the full process before drop day tomorrow

recidiviz/staging/cron.yaml
colincadams: Okay then should it be pushed automatically by Travis?

deploy_production.sh
colincadams: ack, filed #657

README.md
```diff
@@ -116,19 +116,31 @@ described in [running tests](#running-tests)). In practice, this is not particul
...
-`gcloud app deploy`. This will upload the full project to the cloud and enable it as the current version in your environment.
+### Deployment
+
+### Deploying a scraper
+To deploy a scraper, simply edit the [cron](./cron.yaml).  The release engineer
+that week should check for all of the newly added scrapers and manually test them.
+
+The release engineer oncall should go through the following steps:
+
+#### On Monday Morning
```
colincadams: re: frequent releases. Agreed we should use this process for more frequent releases, so I would just rename these headers to *Push to Staging* and *Push to Prod* so that the docs are still relevant. Then in the text you can say 'typically on Monday morning' if preferred.

arian487: Done.

PR #630, opened 2019-01-15
Remove pylint dependency tree

zacharylawrence:
@jessex It seems like everyone here in NYC doesn't find the pylint dependency tree very useful (just noisy). Thoughts on disabling it?


zacharylawrence:
> I agree! Let's disable it. But would you mind commenting out the lines instead of deleting them? I find that the most useful approach to maintaining bespoke configuration like this.

Actually the only thing necessary seems to be disabling rule `RP0401` (aka `External Dependencies`). If you think it makes sense, we can go ahead and leave this imports as is and just disable this rule?

PR #634, opened 2019-01-15
Create get_initial_params method instead of individual methods for each param

colincadams:
This also allows for custom headers to be set in params.


recidiviz/ingest/us_fl_hendry/us_fl_hendry_scraper.py
```diff
@@ -95,14 +95,11 @@ def populate_data(self, content, params, ingest_info):
...
-        task_type = params.get('task_type', self.get_initial_task_type())
-        endpoint = params.get('endpoint', self.get_initial_task_type())
-
-        if self.is_initial_task(task_type):
+        if self.is_initial_task(params['task_type']):
             return [self._get_search_page()]
 
         tasks = []
-        tasks.extend(self._get_next_page(content, endpoint))
+        tasks.extend(self._get_next_page(content, params.get('endpoint')))
```
arian487: Can we just do params['endpoint'] here to stay consistent?

colincadams: Yep, this region always sets `endpoint` so that will work!

colincadams:
`create_scraper` doesn't override any of these methods afaict

colincadams:
We could add the new method to it however, if we think that would be useful? I was thinking of creating a separate PR that removes `set_initial_vars` as it isn't clear to me how it is useful. Then I can remove that from the scraper creation docs and instead mention this new method to override initial params.

PR #635, opened 2019-01-15
Add main logic for writing historical snapshots to database

zdg2102:


recidiviz/persistence/database/database.py
```diff
@@ -171,6 +181,215 @@ def update_booking(session, booking_id, **kwargs):
...
+
+    # Traverse all relationships on the record tree and save historical
+    # snapshots where needed
+    #
+    # Some entities in the record tree can be reached by more than one
+    # relationship path, so we need to track which ones have already been
+    # explored
+    unexplored_entities = set([person])
+    explored_entities = set()
+    while len(unexplored_entities) > 0: # pylint: disable=len-as-condition
```
jessex: I think we can just do `while unexplored_entities`

recidiviz/persistence/database/database.py
```diff
@@ -171,6 +181,215 @@ def update_booking(session, booking_id, **kwargs):
         .update(_convert_enums_to_strings(kwargs))
 
 
+def _save_record_tree(session, person):
```
jessex: It looks like we're doing a good ol' fashioned depth-first-search here. I think adding in a graph library and then plugging in our logic for defining the graph as an entity and its tree of relationships would probably not save us very much in the way of lines of code, but it would have the benefit of deferring search logic to a library dedicated for that purpose, letting us focus just on the essential complexity of defining snapshots here. I imagine the tradeoff is probably not worth it but wondering if you have that any consideration?

recidiviz/persistence/database/database.py
```diff
@@ -171,6 +181,215 @@ def update_booking(session, booking_id, **kwargs):
...
+
+
+def _get_most_recent_historical_snapshot(session, entity,
+                                         historical_table_class):
+    """Returns the most recent historical snapshot corresponding to |entity| if
+    one exists, otherwise returns (None).
+    """
+
+    key_column_name = _get_primary_key_column_name(entity.__table__)
+    master_entity_id = getattr(entity, key_column_name)
```
jessex: I don't think passing this in as an argument would make the signature of this method any worse.

recidiviz/tests/persistence/database/database_test.py
```diff
@@ -163,3 +171,316 @@ def test_readPeopleWithOpenBookings(self):
         expected_people = [database_utils.convert_person(p) for p in
                            [person_match_full_name]]
         self.assertCountEqual(people, expected_people)
+
```
jessex: Fantastic test coverage in here 👏

recidiviz/persistence/database/database.py
zdg2102: I hadn't really thought about it. Since most of the complexity here is around finding the relationships and determining what kind of relationships they are, and the actual graph exploration is a very straightforward one, I think the logic we'd be able to eliminate by using a library would end up getting offset by the logic we'd have to add to provide the relationships to the library in whatever format it requires.

recidiviz/persistence/database/database.py

recidiviz/persistence/database/database.py
arian487: Having read the logic below, I agree with zach I think the actual graph search portion of the code is pretty negligible.  Its not even a real DFS I don't think set.pop guarantees any order from the pops (not that a DFS/BFS wouldn't work.

recidiviz/persistence/database/database.py
```diff
@@ -171,6 +181,215 @@ def update_booking(session, booking_id, **kwargs):
...
+    # All historical snapshot changes should be given the same timestamp
+    snapshot_time = datetime.now()
+
+    # Traverse all relationships on the record tree and save historical
+    # snapshots where needed
+    #
+    # Some entities in the record tree can be reached by more than one
+    # relationship path, so we need to track which ones have already been
+    # explored
+    unexplored_entities = set([person])
```
arian487: Discussed offline, but the fact that this is working tells me that SQLAlchemy objects are hashable by default, so its worth verifying it isn't an expensive hash function.  I can imagine a universe where the hash function traverses the entity relationships which would become a very expensive operation for us and would definitely disqualify this approach IMO.

recidiviz/persistence/database/database.py
```diff
@@ -171,6 +181,215 @@ def update_booking(session, booking_id, **kwargs):
...
+    """Returns (True) if all fields on |entity| are equal to the corresponding
+    fields on |historical_snapshot|.
+
+    NOTE: This method *only* compares columns which are present on both the
+    master and historical tables. Any column that is only present on one table
+    will be ignored.
+    """
+
+    for column_property_name in _get_shared_column_property_names(
+            entity.__class__, historical_snapshot.__class__):
```
arian487: does type(entity) not work here instead of accessing __class__?

recidiviz/persistence/database/database.py
```diff
@@ -171,6 +181,215 @@ def update_booking(session, booking_id, **kwargs):
...
+    snapshot_time = datetime.now()
+
+    # Traverse all relationships on the record tree and save historical
+    # snapshots where needed
+    #
+    # Some entities in the record tree can be reached by more than one
+    # relationship path, so we need to track which ones have already been
+    # explored
+    unexplored_entities = set([person])
+    explored_entities = set()
```
arian487: I think the name is a bit confusing here because we have entities objects that are different from these database objects.  I know you use the word a lot everywhere and it might be a bit annoying, but I'd be in favour of changing this name to something that doesn't clash with our entities name.

recidiviz/tests/persistence/database/database_test.py
```diff
@@ -163,3 +171,316 @@ def test_readPeopleWithOpenBookings(self):
...
+    def testWriteRecordTree_noExistingSnapshots_createsSnapshots(self):
+        act_session = Session()
+
+        person = entities.Person.new_with_defaults(
+            region=_REGION, race=Race.OTHER)
+
+        booking = entities.Booking.new_with_defaults(
+            custody_status=CustodyStatus.IN_CUSTODY,
+            last_seen_time=_LAST_SEEN_TIME)
+        person.bookings = [booking]
```
zacharylawrence: If you want, you can do this in one line like:
```
person = entities.Person.new_with_defaults(
  region=_REGION,
  race=Race.OTHER,
  bookings=[
    entities.Booking.new_with_defaults(
      custody_status=CustodyStatus.IN_CUSTODY,
      last_seen_time=_LAST_SEEN_TIME
    )
  ]
)
```

And so on for the fields below

recidiviz/tests/persistence/database/database_test.py
```diff
@@ -163,3 +171,316 @@ def test_readPeopleWithOpenBookings(self):
...
+
+        persisted_person = database.write_person(act_session, person)
+        act_session.commit()
+
+        person_id = persisted_person.person_id
+        booking_id = persisted_person.bookings[0].booking_id
+        charge_1_id = persisted_person.bookings[0].charges[0].charge_id
+        charge_2_id = persisted_person.bookings[0].charges[1].charge_id
+
+        act_session.close()
```
zacharylawrence: Is there a specific reason you need 2 sessions here?

zdg2102:
@jessex So the two main assumptions are the two listed in `_update_historical_snapshots`: the ORM class naming and the shared name for the master key.

I'm not particularly worried about the ORM class naming, because I doubt anyone is going to see the whole schema file full of Thing and ThingHistory classes and then add something that violates that convention.

The shared name for the master key is essentially untestable, because it's about what the columns "mean", rather than anything we can check for in the code.

For everything else, we don't require the column and field names to match, and the code will work properly regardless of whatever names they have. We do require the field names on master and historical to match, but we already have `schema_test.py` to cover that constraint.

The one somewhat weird test strategy I can think of would be: add a test case that writes the historical snapshots for a complete record tree, "complete" here meaning it includes every possible master and historical entity type. This would ensure all the conversions were correct for every table. Then add a helper validation that walks the record tree, checks every relationship, and throws an error if an empty relationship is found. This would force anyone adding a new entity type to the record tree to update this test to include that new entity type. What do you think of that approach?

recidiviz/persistence/database/database.py
```diff
@@ -171,6 +181,215 @@ def update_booking(session, booking_id, **kwargs):
...
+    historical_table_class = getattr(schema, historical_table_class_name)
+    key_column_name = _get_primary_key_column_name(entity.__table__)
+    master_entity_id = getattr(entity, key_column_name)
+
+    most_recent_historical_snapshot = _get_most_recent_historical_snapshot(
+        session, entity, historical_table_class)
+
+    # Historical table only needs to be updated if entity has no existing
+    # snapshots (because it's new) or entity does not match current snapshot
+    if most_recent_historical_snapshot is None or \
```
zacharylawrence: Thoughts on flipping this around and treating it as a guard clause? This will let us avoid all the indentation and might let us juggle less state in our minds.

```
if not most_recent_historical_snapshot and \
    _does_entity_match_historical_snapshot(...):
  return
```

recidiviz/persistence/database/database.py
```diff
@@ -171,6 +181,215 @@ def update_booking(session, booking_id, **kwargs):
...
+    equal. If this assumption is violated, callers of this method will not
+    behave as expected.
+    """
+
+    # table.primary_key returns a PrimaryKeyConstraint containing a single
+    # column
+    return table.primary_key.columns.values()[0].name
+
+
+def _does_entity_match_historical_snapshot(entity, historical_snapshot):
```
zacharylawrence: Should this method live on the schema ORM object?

recidiviz/persistence/database/database.py
```diff
@@ -171,6 +181,215 @@ def update_booking(session, booking_id, **kwargs):
...
+            entity.__class__, historical_snapshot.__class__):
+        entity_value = getattr(entity, column_property_name)
+        historical_value = getattr(historical_snapshot, column_property_name)
+        if entity_value != historical_value:
+            return False
+
+    return True
+
+
+def _copy_entity_fields_to_historical_snapshot(entity, historical_snapshot):
```
zacharylawrence: Would it also make sense for this to live on the ORM object?

recidiviz/tests/persistence/database/database_test.py
zdg2102: I personally find that level of nesting a little harder to read versus just constructing them all in sequence

recidiviz/tests/persistence/database/database_test.py
zdg2102: I split the sessions into arrange/act/assert across all these test cases. I wanted to be sure that nothing was carried over and there were no dirty entities from one session affecting a later one, since in the real process this test is meant to be emulating, these will be separate sessions probably occurring on separate days

recidiviz/persistence/database/database.py
zdg2102: I couldn't figure out from the SQLAlchemy source what the actual hashing process being used here was, so I updated this traversal to explicitly track the entities by object ID (that way we're guaranteed to only be hashing ints)

recidiviz/persistence/database/database.py
zdg2102: Replaced all uses of `thing.__class__` with `type(thing)`

recidiviz/persistence/database/database.py
zdg2102: Makes sense. Done

recidiviz/persistence/database/database.py
zdg2102: I'd rather leave this method here for now just because I don't think it's an inherent part of the ORM object that's really relevant to expose everywhere else the ORM object is used. If it does end up being needed elsewhere though it would probably make sense to move there

recidiviz/persistence/database/database.py
zdg2102: This one seems to me more like the responsibility of the database (reading and writing between tables) rather than the responsibility of the individual ORM models

recidiviz/persistence/database/database.py
zdg2102: This PR probably isn't the right venue to solve the larger issue, but I think the problem is that "entities" is too easily overloaded of a word to be used to properly identify our persistence layer models. I think it would be a lot clearer if those models were renamed, and then we could use the word "entity" as a more generic term like I'm using it here.

I can change it here if you think it's too confusing, but I'd be more in favor of solving what seems to me to be the root problem.

@zacharylawrence and @terinpatelwilson were having some discussion on this recently, although I don't know if they came to any conclusion

jessex:
@zdg2102 your response makes sense, thanks. I think your test case idea is sound, and it aligns philosophically with some of the other tests we have in the codebase, designed to force engineers adding new scrapers or new fields for example to acknowledge core behavior that requires their presence.

zdg2102:
@jessex okay, sounds good. Would you mind if I add that new test in a separate PR, since I think this one is already fairly long and ugly?

recidiviz/persistence/database/database.py
arian487: Awesome, thanks!

recidiviz/persistence/database/database.py
arian487: Fair points, lets leave this as is in that case.

recidiviz/persistence/database/database.py
```diff
@@ -171,10 +181,227 @@ def update_booking(session, booking_id, **kwargs):
...
+    #
+    # Some entities in the record tree can be reached by more than one
+    # relationship path, so we need to track which ones have already been
+    # explored
+    #
+    # As we're not concerned here with any kind of equality between objects
+    # other than their position in the record tree, exploration is tracked by
+    # object identity
+    unexplored_entities = {id(person): person}
+    explored_entities = {}
```
zacharylawrence: @zdg2102 As discussed offline, could you also add a comment why we're using a List instead of a Set here?

recidiviz/persistence/database/database.py
```diff
@@ -171,10 +181,227 @@ def update_booking(session, booking_id, **kwargs):
...
+    while unexplored_entities:
+        entity_object_id = next(iter(unexplored_entities.keys()))
```
colincadams: nit: I think these lines could just be `entity_object_id, entity = unexpolored_entities.popitem()`

recidiviz/tests/persistence/database/database_test.py
```diff
@@ -163,3 +171,316 @@ def test_readPeopleWithOpenBookings(self):
...
+            race=Race.EXTERNAL_UNKNOWN)
+
+        # Ingested booking has new last_seen_time but this is ignored, as it
+        # is not included on the historical table.
+        ingested_booking = entities.Booking.new_with_defaults(
+            booking_id=existing_booking_id,
+            custody_status=CustodyStatus.IN_CUSTODY,
+            facility=_FACILITY,
+            last_seen_time=updated_last_seen_time)
+        ingested_person.bookings = [ingested_booking]
```
colincadams: since we are modifying the charges associated with this booking, wouldn't we expect a new snapshot of the booking as well?

recidiviz/persistence/database/database.py
zdg2102: Discussed at length offline with @zacharylawrence and decided to just track them in a list instead. Since the graph exploration only covers master, not historical, it should never be larger than ~100 entities in the extreme case, and probably much fewer most of the time. Because of that, the increase in readability in not making the reader keep track of how and why we're using object IDs is probably worth the efficiency cost in scanning over lists. (We can always switch back to ID maps if this turns out to be a bottleneck.)

recidiviz/persistence/database/database.py

Issue #598, opened 2019-01-11
Regions should have a way of telling Infer release not to release


arian487:
Some websites you can only query people booked over a time period (one month for example) so infer release would mark them released if they are there for more than one month.  Regions need a way of telling the releaser to mark them in a different state (MISSING_FROM_SITE for example) if missing does not necessarily mean release.

Snapshot scraping would need to be prioritized for these states.

terinpw:
Closed by #639

PR #639, opened 2019-01-16
Allow regions to specify what a person being removed from a website means

terinpw:


terinpw:
Would we rather the region value be a boolean instead of an enum? That way people can set a field to True/False instead of having to write `UNKNOWN_SIGNFICANCE`

terinpw:
Closes #598

recidiviz/ingest/infer_release.py
```diff
@@ -22,34 +22,41 @@
...
     return '', HTTPStatus.OK
 
 
-def _get_jail_regions():
-    return [region_code for region_code, region_dict in
-            get_supported_regions(full_manifest=True).items() if
-            region_dict['agency_type'] == 'jail']
+def _get_release_reason(region: Region):
+    removed_from_website = region.removed_from_website
+    if removed_from_website is RemovedFromWebsite.RELEASED:
```
arian487: I think we want == here.  Is is a special keyword that checks for object equality.  I think it might still work but using == is more correct.

recidiviz/ingest/infer_release.py
```diff
@@ -22,34 +22,41 @@
...
+        return ReleaseReason.INFERRED_RELEASE
+    if removed_from_website is RemovedFromWebsite.UNKNOWN_SIGNIFICANCE:
```
jessex: Is this extra layer of indirection necessary? I'm guessing this is in place because we don't want to couple region config to the set of available release reasons? With only a couple of possible values it feels like overkill but I could see it evolving into a good protection.

terinpw: Yeah, I made it because we have a bunch of ReleaseReasons that don't correspond to anything we want scraper writers to specify. If we just let them set ReleaseReason in the region config, I'd be worried someone would put something that's not `INFERRED_RELEASE` or `REMOVED_FROM_WEBSITE` or that it would be confusing for someone to understand which ones they need to use. 

But maybe the combination of good documentation & strict code reviews would make this fine. What do you think? I marginally prefer making the interface for scraper writers simpler (even if it means we have this conversion here), but I also hear the point of not overcomplicating our code.

terinpw: Ok sg!

recidiviz/ingest/infer_release.py
terinpw: `is` works fine for enums, but chatted offline and using `==` just for consistency in the codebase.

PR #640, opened 2019-01-16
Add error counters to persistence layer

arian487:


recidiviz/persistence/persistence.py
```diff
@@ -133,7 +141,10 @@ def write(ingest_info, metadata):
             logging.info('Successfully wrote to the database')
             session.commit()
             persisted = True
-        except Exception:
+        except Exception as e:
+            # Record the error ype that happened and increment the error counter
```
jessex: ype --> type

recidiviz/persistence/persistence.py
```diff
@@ -133,7 +141,10 @@ def write(ingest_info, metadata):
...
+            mtags[monitoring.TagKey.ERROR] = type(e).__name__
+            measurements.measure_int_put(m_errors, 1)
```
jessex: @colincadams would know better than I, but I think that in OpenCensus a `SumAggregation` that always increments by 1 is the same as a `CountAggregation`, so maybe we should just change to the latter?

recidiviz/persistence/persistence.py
arian487: Done.

recidiviz/persistence/persistence.py
arian487: Chatted offline with Colin and SumAggregation is what we want, I'll let him explain it better though :)

jessex: I love being told why I'm wrong. Lay it on me.

colincadams: Turns out I didn't know what I was talking about :grin: Either `SumAggregation` or `CountAggregation` will work here. I prefer `SumAggregation` as the underlying measure is measuring the number of errors, so for the view it seems logical to add up the values of the measure instead of counting the number of values. But in practice either will work.

PR #641, opened 2019-01-16
HTML5BaseScraper

### LINKED ISSUE #633: figure out how to handle tables lxml can't parse
Alachua, FL (http://oldweb.circuit8.org/inmatelist.php) doesn't close some of their html tags.

```html
<table border=1><tr>
<td><b>Name
    <td><b>Booking Date/Time
        <td><b>Hold
            <tr>
                <td>LAST, FIRST
                    <td>Nov 30
                        <td>N
                            <td>N</table>
```

lxml parses it like this:

```html
<table border="1">
    <tr>
        <td><b>Name</b></td>
        <td><b>Booking Date/Time</b></td>
        <td><b>Hold
            <tr>
                <td>LAST, FIRST</td>
                <td>Nov 30</td>
                <td>N</td>
            </tr>
        </b></td>
    </tr>
</table>
```

I will preprocess the html content. 🤮

jamwalla:
fixes #633


jessex:
High level question: could we not just use this new scraper's method for parsing HTML content in the base scraper itself? Would it break any existing scrapers if we applied that change directly in base scraper?

jamwalla:
It wouldn't break any old scrapers. We talked about explicitly making this behavior separate so that other scrapers don't have to parse content twice if they don't need to, but if that's an acceptable overhead I do think this is more correct.

arian487:
I don't like the idea of processing the HTML twice unless we have to, hence breaking it out in this fashion.

arian487:
Yea I think parsing HTML can be costly, especially if the websites are large and complex so I don't feel its a premature optimization to have this be optional functionality

jamwalla:
Cool, thanks! I opened #643 to track performance testing.

PR #642, opened 2019-01-16
Create a TaskParams object and convert constants to enums

colincadams:


colincadams:
Right now this just makes the changes in the base files and fixes `us_co_mesa` as a sanity check. Wanted to send it out to get thoughts before I change other regions or tests.

recidiviz/ingest/task_params.py
```diff
@@ -0,0 +1,93 @@
...
+                converted[key] = constants.ScrapeType(value)
+            elif key == 'task_type':
+                converted[key] = constants.TaskType(value)
+            elif key == 'scraper_start_time':
+                converted[key] = dateparser.parse(value)
+            else:
+                converted[key] = value
+        return cls(**converted)
+
+    @classmethod
```
zacharylawrence: Could this be `@staticmethod`?

recidiviz/ingest/base_scraper.py
```diff
@@ -112,23 +112,22 @@ def _generic_scrape(self, params):
...
-            }
-        task_type = params.get('task_type')
-        endpoint = params.get('endpoint')
+        if not params.task_type:
+            params = TaskParams.merge(
+                self._get_default_initial_params(),
+                self.get_initial_params(),
+                params,
+            )
+        assert params.scraper_start_time is not None
```
jessex: Sorry, I know I ask this once per month, but are we okay with using `assert` in production code or not?

zacharylawrence: Thoughts on making this a required field on the TaskParams attr so that it must be constructed with a valid value?

arian487: Personally I'm against it.

recidiviz/ingest/task_params.py
```diff
@@ -0,0 +1,93 @@
...
+
+import attr
+import dateparser
+
+from recidiviz.ingest import constants, scraper_utils
+from recidiviz.ingest.models.ingest_info import IngestInfo
+
+
+@attr.s(frozen=True, auto_attribs=True)
+class TaskParams:
```
jessex: Personally, I'd like to see a basic docstring explaining each of the fields on this class. I think there's enough ambiguity/similarity between field names that most future engineers will need to read through usages of the class to find what the fields mean, which is non-ideal.

recidiviz/ingest/constants.py
```diff
@@ -16,19 +16,21 @@
 # =============================================================================
 
 """Constants file used by everyone."""
+import enum
 
 # These constants tell the generic scraper what functions to perform
-INITIAL_TASK = 0x1
-SCRAPE_DATA = 0x2
-GET_MORE_TASKS = 0x4
+class TaskType(enum.Flag):
```
zacharylawrence: Are these enum changes logically distinct from the rest of this PR and would it make sense to break them out into a seperate change? I'm fine if you think this isn't worth the effort and want to just leave them together.

recidiviz/ingest/base_scraper.py
zacharylawrence: > Personally I'm against it.

+1. If this is something worth error handling then we should do it properly.

recidiviz/ingest/task_params.py
```diff
@@ -0,0 +1,93 @@
...
+                converted[key] = constants.TaskType(value)
+            elif key == 'scraper_start_time':
+                converted[key] = dateparser.parse(value)
+            else:
+                converted[key] = value
+        return cls(**converted)
+
+    @classmethod
+    def evolve(cls, params, **kwds):
+        """Convenience so that other modules don't need to import attr."""
```
jessex: 🎉

recidiviz/ingest/base_scraper.py
jamwalla: This could also (sort of) be addressed by making the field required instead of optional

recidiviz/ingest/task_params.py
```diff
@@ -0,0 +1,93 @@
...
+    post_data: Optional[dict] = None
+    json: Optional[dict] = None
+
+    content: Optional[str] = None
+    ingest_info: Optional[IngestInfo] = None
+
+    def to_serializable(self):
+        for key, value in attr.asdict(self, filter=_not_none).items():
+            # map enum fields to their strings
+            if key in {'scrape_type', 'task_type'}:
```
zacharylawrence: Instead of listing each field out, you could just check if the key is of `Enum` type?

recidiviz/ingest/task_params.py
```diff
@@ -0,0 +1,93 @@
...
+        """
+        new_params = cls()
+        for other in others:
+            # Get the attributes of `other` that are set
+            other_attribs = attr.asdict(other, filter=_not_none)
+            new_params = attr.evolve(new_params, **other_attribs)
+        return new_params
+
+
+def _not_none(_attrib, value):
```
zacharylawrence: This LGTM, but just noting that a lambda would work too.

recidiviz/ingest/task_params.py
```diff
@@ -0,0 +1,93 @@
...
+    endpoint: Optional[str] = None
+
+    headers: Optional[dict] = None
+    post_data: Optional[dict] = None
+    json: Optional[dict] = None
+
+    content: Optional[str] = None
+    ingest_info: Optional[IngestInfo] = None
+
+    def to_serializable(self):
```
zacharylawrence: @jessex Should we consider pickling this class instead of writing our own custom serialization method? Alternatively we could consider replacing this class with a proto?

jessex: Mm, good question. I breezed over that. I think a proto might be overkill here, though it does have precedent in the codebase. Pickling feels like the Pythonic approach for serializing straight-forward nested data to me. It doesn't have precedent directly in our codebase yet, but this is a good place to introduce it.

colincadams: The only thing I don't love about pickling is that then the serialized version isn't human readable. This isn't a big deal, but I do think it's a convenience especially since we can view the payloads of running tasks in the Cloud Tasks UI:
image

Using `asdict` when serializing prevents us from accidentally skipping a field when serializing, or forgetting to add it to the serialize method when you add a new field, which is typically my biggest concern with writing your own serialization. So I don't mind this that much.

I would think that with type annotations there could be a good library to make classes serializable and deserializable but I couldn't find a good one as an alternative.

recidiviz/ingest/base_scraper.py
colincadams: I didn't want to make this a required field because there are places in the code where I want to be able to build a `TaskParams` without a `scraper_start_time`, for example when defining default initial params.

recidiviz/ingest/constants.py
colincadams: Once we add these we need to handle the serialize/deserialize issues, which I thought was simpler to handle once we had the `attr` based `TaskParams` class.

recidiviz/ingest/task_params.py
colincadams: Done!

recidiviz/ingest/task_params.py
jessex: The fact that we don't have to manually add in fields for serialization as we come up with new ones is a good argument, but one that also applies to Pickling. So it seems like the trade-off in consideration is the convenience/readability of json versus the built-in simplicity of Pickle. I think the readability is also important in other operational areas, such as logging and potentially debugging/troubleshooting scenarios.

recidiviz/ingest/constants.py
zacharylawrence: SGTM!

recidiviz/ingest/task_params.py
zacharylawrence: This is awesome! LGTM but it might be nice to consider creating an annotation in another change so we can easily add this to our other attr classes!

recidiviz/ingest/base_scraper.py
zacharylawrence: As discussed offline, we could consider making a few of these fields required and passing around a `Builder` here. Then when we call `.build()` an exception would be thrown if a required field was never set.

colincadams: I chatted with @zacharylawrence and we decided to split this into two different objects, one that the scraper writer uses and a wrapper that includes some extra information and that is put on the queue.

This allowed us to get around the `scraper_start_time` issue as it is required in the wrapper object.

recidiviz/ingest/task_params.py
colincadams: I'm now using `cattrs` which handles converting the objects and enums to primitives and back for us. The only custom piece is telling it how we want the datetime serialized and deserialized. What do you all think of this approach?

recidiviz/ingest/constants.py
colincadams: reviving this conversation. We could do the TaskParams object change first, then the enums, just not in the other order. Given the size of this change, would that be preferred? Or since both changes will still have to touch all of the scraper files would we rather do it all at once?

recidiviz/ingest/scraper.py
```diff
@@ -119,11 +124,10 @@ def start_scrape(self, scrape_type):
...
             return
 
-        params = {"scrape_type": scrape_type,
-                  "content": docket_item,
-                  "scraper_start_time": self.get_now_as_str()}
-
-        self.add_task(self.get_initial_task(), params)
+        self.add_task(self.get_initial_task(),
+                      QueueRequest(scrape_type=scrape_type,
+                                  scraper_start_time=datetime.now(),
```
jessex: Are these indented correctly or meant to be one space further in to align? Same with the QueueRequest instantiation on line 225 below.

recidiviz/ingest/task_params.py
```diff
@@ -0,0 +1,75 @@
...
+                                datetime.datetime.isoformat)
+cattr.register_structure_hook(
+    datetime.datetime,
+    lambda serialized, desired_type: desired_type.fromisoformat(serialized))
+
+@attr.s(frozen=True)
+class NextTask:
+    """Describes a scrape task to be performed"""
+
+    # The type of task, this controls which meth
```
jessex: which meth?

recidiviz/ingest/task_params.py
jessex: I think this is good approach. I hadn't seen `cattrs` before but it seems very convenient.

recidiviz/ingest/base_scraper.py
```diff
@@ -89,21 +90,21 @@ def _fetch_content(self, endpoint, response_type, headers=None,
                                post_data=post_data, json_data=json_data)
         if page == -1:
             return -1
-        if response_type == constants.HTML_RESPONSE_TYPE:
+        if response_type == constants.ResponseType.HTML:
```
zacharylawrence: Nit: Consider importing just `ResponseType`? I don't mind either way.

recidiviz/ingest/task_params.py
```diff
@@ -0,0 +1,93 @@
...
+    scraper_start_time: datetime.datetime = attr.ib()
+
+    # The information need ed to perform the scrape task
+    next_task: Task = attr.ib()
+
+    # In the case that a person's information is spread across multiple pages
+    # this is used to pass it along to the next page.
+    ingest_info: Optional[IngestInfo] = attr.ib(default=None)
+
+    def to_serializable(self):
```
arian487: Can you add a comment and say that this function will throw an error if we have an ingest info object that is not none since it is not serializable.  It can be misleading since it exists here, but it is currently broken functionality (which is fine for now)

recidiviz/ingest/scraper.py
colincadams: Fixed

recidiviz/ingest/task_params.py
colincadams: Fixed

recidiviz/ingest/run_scraper.py
```diff
@@ -73,7 +76,8 @@ def add_task(self, task_name, params):
     logging.info('***')
     fn = getattr(self, task_name)
     try:
-        fn(params)
+        fn(QueueRequest.from_serializable(json.loads(
```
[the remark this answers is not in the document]
colincadams: Added a comment for this. It simply ensures the serialization / deserialization is tested locally as well instead of just in production

recidiviz/ingest/task_params.py
colincadams: Fixed

recidiviz/ingest/base_scraper.py
colincadams: Leaving this for now

recidiviz/ingest/task_params.py
colincadams: Fixed

recidiviz/ingest/base_scraper.py
colincadams: These objects are now split

recidiviz/ingest/constants.py
colincadams: I know this PR is huge, but if possible I would like to land it all at once to unblock writing of child scrapers

colincadams:
This should be good to review now.

recidiviz/ingest/task_params.py
```diff
@@ -0,0 +1,94 @@
...
+        attr.ib(default=constants.ResponseType.HTML)
+    # Any http headers to be sent in the request
+    headers: Optional[Dict[str, str]] = attr.ib(default=None)
+    # Any post data to send in the request (forces a POST request)
+    post_data: Optional[Dict[str, Optional[str]]] = attr.ib(default=None)
+    # Any json data to be sent in the request (forces a POST request)
+    json: Optional[Dict[str, Optional[str]]] = attr.ib(default=None)
+
+    # Can be used by the child scraper to store extra information they need
+    custom: Dict[Any, Any] = attr.ib(default={})
```
zacharylawrence: You should use `factory=list` here instead of assigning the default to a mutable object: http://www.attrs.org/en/stable/examples.html#defaults

recidiviz/ingest/task_params.py
```diff
@@ -0,0 +1,94 @@
...
+from recidiviz.ingest.models.ingest_info import IngestInfo
+
+cattr.register_unstructure_hook(datetime.datetime,
+                                datetime.datetime.isoformat)
+cattr.register_structure_hook(
+    datetime.datetime,
+    lambda serialized, desired_type: desired_type.fromisoformat(serialized))
+
+@attr.s(frozen=True)
+class Task:
```
zacharylawrence: Just noting that it may be helpful for this to extend `BuildableAttr` if we could see uses for using the builder pattern in child scrapers. I think this is fine as is until then.

recidiviz/ingest/task_params.py
colincadams: Thanks good catch!

PR #644, opened 2019-01-16
Iml vendor scraper and us_nj_bergen

ohinds:


recidiviz/common/constants/bond.py
```diff
@@ -52,6 +52,7 @@ def _get_default_map():
     'NO BOND': BondType.NO_BOND,
     'NO BOND ALLOWED': BondType.NO_BOND,
     'NONE SET': BondType.NO_BOND,
+    'RELEASED ON OWN RECOGNIZANCE': BondType.NO_BOND,
```
jessex: Just for edification, this is what ROR below refers to as well?

recidiviz/ingest/scraper.py
```diff
@@ -224,21 +224,27 @@ def resume_scrape(self, scrape_type):
         self.add_task(self.get_initial_task(), params)
 
     @staticmethod
-    def fetch_page(url, post_data=None, json_data=None):
+    def fetch_page(url, post_data=None, json_data=None, extra_headers=None):
```
jessex: I think there's a merge conflict here. `scraper.fetch_page` now has `headers` as its second argument, which you should be able to use.

recidiviz/common/constants/charge.py
```diff
@@ -77,9 +78,15 @@ def _get_default_map():
     '1ST': ChargeDegree.FIRST,
     '2ND': ChargeDegree.SECOND,
     '3RD': ChargeDegree.THIRD,
+    '4TH': ChargeDegree.FORTH,
```
arian487: Fourth?

recidiviz/ingest/vendors/iml/iml_scraper.py
```diff
@@ -0,0 +1,349 @@
...
+            os.path.dirname(__file__), key_mapping_file)
+
+        bond_key_mapping_file = 'iml_bond.yaml'
+        self.bond_yaml_file = os.path.join(
+            os.path.dirname(__file__), bond_key_mapping_file)
+
+        super(ImlScraper, self).__init__(region_name)
+
+        self._base_endpoint = self.get_region().base_url
+        self._num_people_per_search = 30
```
jessex: Let's extract this number out to a constant in the file for visibility.

recidiviz/ingest/vendors/iml/iml_scraper.py
```diff
@@ -0,0 +1,349 @@
...
+        """Parses the number of people out of the content.
+
+        Args:
+            content: An lxml html tree.
+
+        Returns:
+            The total number of people to be scraped.
+
+        """
+        matches = re.findall(b'([0-9]+?) results', html.tostring(content))
```
jessex: There's no CSS selector or anything to look for? I could envision, e.g. some odd SEO-related update, adding "X results" somewhere in the skeleton of the page.

recidiviz/ingest/vendors/iml/iml_scraper.py
```diff
@@ -0,0 +1,349 @@
...
+        if len(matches) != 2:
```
jessex: Why do we expect two results? What's the second one? This relates to my comment one line up about a more resilient check.

recidiviz/ingest/vendors/iml/iml_scraper.py
```diff
@@ -0,0 +1,349 @@
...
+from recidiviz.ingest.extractor.html_data_extractor import HtmlDataExtractor
+
+
+class ImlScraper(BaseScraper):
+    """Scraper for counties using Iml."""
+
+    def __init__(self, region_name):
+        key_mapping_file = 'iml.yaml'
+        self.yaml_file = os.path.join(
+            os.path.dirname(__file__), key_mapping_file)
```
arian487: I would bake in the option to have a vendor scraper pass in its own yaml file.  I don't know if its possible, but I can envision that some regions can have different fields populated even amongst vendor scrapers (this is true for archonix for example).  You can use jailtracker and us_vt as a model

recidiviz/ingest/vendors/iml/iml_scraper.py
arian487: Add this as a class constant instead of a member instance variable.

recidiviz/ingest/vendors/iml/iml_scraper.py
```diff
@@ -0,0 +1,349 @@
...
+
+        Returns:
+            a dictionary with data needed to get the next page of search
+            results.
+
+        """
+
+        # Get the last result index on the previous page.
+        # pylint:disable=anomalous-backslash-in-string
+        matches = re.findall(b'Showing\s*[0-9]+\s*to\s*([0-9]+)\s*of',
```
jessex: Same comments as above about looking for specific elements or classes instead of doing a big regex search, and the number of matches thereof.

recidiviz/ingest/vendors/iml/iml_scraper.py
arian487: I don't like the idea of parsing the content back to raw html, html parsing can be an expensive operation and that is somewthing I'd like to avoid when at scale.  I'd rather have a flag that you can pass down through the params that tells fetch content to return us parsed html or not.

recidiviz/ingest/vendors/iml/iml_scraper.py
```diff
@@ -0,0 +1,349 @@
...
+        Args:
+            content: An lxml html tree.
+
+        Returns:
+            The total number of people to be scraped.
+
+        """
+        matches = re.findall(b'([0-9]+?) results', html.tostring(content))
+        if len(matches) != 2:
+            logging.error('Did not find exactly two matches when parsing the '
```
arian487: Can you add a comment here on why we always expect two?  Does this always hold true?

recidiviz/ingest/vendors/iml/iml_scraper.py
```diff
@@ -0,0 +1,349 @@
...
+
+            sysID = int(ids[1])
+            imgSysID = int(ids[2])
+
+            # Get the person IDs for the DB.
+            td_as = row.xpath('td/a')
+
+            if len(td_as) != 4:
+                logging.error('Did not find exactly 4 numbers in the row '
+                              'click function call, as we should have.')
```
jessex: Is "row click function call" accurate here, or a copy+paste boo boo?

recidiviz/ingest/vendors/iml/iml_scraper.py
```diff
@@ -0,0 +1,349 @@
...
+
+        Args:
+            content: An lxml html tree.
+            params: dict of parameters passed from the last scrape session.
+
+        Returns:
+            A list of params containing endpoint and task_type at minimum.
+        """
+        task_type = params.get('task_type', self.get_initial_task_type())
+        params_list = []
```
jessex: This is an example of Hungarian Notation and it's generally frowned upon, i.e. we shouldn't embed types in the names of our variables if we can avoid it. In this case, something like `task_params` may be more appropriate.

recidiviz/ingest/vendors/iml/iml_scraper.py
```diff
@@ -0,0 +1,349 @@
...
+        Args:
+            content: An lxml html tree.
+            params: dict of parameters passed from the last scrape session.
+            ingest_info: The IngestInfo object to populate
+        """
+        data_extractor = HtmlDataExtractor(self.yaml_file)
+        ingest_info = data_extractor.extract_and_populate_data(content,
+                                                               ingest_info)
+        if len(ingest_info.person) != 1:
+            logging.error("Data extractor didn't find exactly one person as "
```
jessex: In other cases of error-level log statements in this PR, we short-circuit. But here we silently ignore the additional persons. Should we be ignoring them instead of short-circuiting? Or, at least, should we include in the log statement that we'll ignore the extra persons?

recidiviz/ingest/vendors/iml/iml_scraper.py
```diff
@@ -0,0 +1,349 @@
...
+                          "it should have")
+
+        person = ingest_info.person[0]
```
jessex: What if `ingest_info.person` is empty?

recidiviz/ingest/vendors/iml/iml_scraper.py
```diff
@@ -0,0 +1,349 @@
...
+        person.person_id = params['person_id']
+        person.booking[0].booking_id = params['booking_id']
```
jessex: Is `person.booking` guaranteed to be non-empty?

recidiviz/ingest/vendors/iml/iml_scraper.py
```diff
@@ -0,0 +1,349 @@
...
+
+        if person.race.startswith('HISPANIC'):
+            person.ethnicity = 'HISPANIC'
+            person.race = ' '.join(person.race.split()[1:])
+
+        # Find the bonds and take their data
+        bond_tab = [t for t in content.xpath('//table')
```
jessex: This list comprehension is guaranteed to be non-empty? There's definitely a table with the text "Bond Information"?

recidiviz/ingest/vendors/iml/iml_scraper.py
jamwalla: This is handled now in #637 and #650 if the request has `params.['response_type'] = constants.TEXT_RESPONSE_TYPE`

jamwalla: could we use the `<form/>` elements at the top of the page instead? Seems like we could use the `nextSearch` number until it reaches `lastSearch`

recidiviz/ingest/vendors/iml/iml_scraper.py
```diff
@@ -0,0 +1,349 @@
...
+                                                               ingest_info)
+        if len(ingest_info.person) != 1:
+            logging.error("Data extractor didn't find exactly one person as "
+                          "it should have")
+
+        person = ingest_info.person[0]
+        person.person_id = params['person_id']
+        person.booking[0].booking_id = params['booking_id']
+
+        if person.race.startswith('HISPANIC'):
```
jamwalla: Do you think this is something to try to handle in the converter? I could see either way.

recidiviz/ingest/vendors/iml/iml_scraper.py
jamwalla: nit- does it work to replace `html.tostring(t)` with `t.text_content()`?

recidiviz/ingest/vendors/iml/iml_scraper.py
```diff
@@ -0,0 +1,349 @@
...
+        bond_ingest_info = bond_extractor.extract_and_populate_data(bond_tab,
+                                                                    None)
+
+        # Connect bonds and charges, and fill in an ICE hold when we
+        # see it in the charge name.
+        for charge in person.booking[0].charge:
+            if charge.name == 'IMMIGRATION DETAINEE':
+                person.booking[0].hold = 'ICE'
+
+            if not bond_ingest_info.person:
```
jamwalla: this check should be above the for loop i think

recidiviz/ingest/vendors/iml/iml_scraper.py
```diff
@@ -0,0 +1,349 @@
...
+
+        if person.race.startswith('HISPANIC'):
+            person.ethnicity = 'HISPANIC'
+            person.race = ' '.join(person.race.split()[1:])
+
+        # Find the bonds and take their data
+        bond_tab = [t for t in content.xpath('//table')
+                    if b'Bond Information' in html.tostring(t)][0]
+        bond_extractor = HtmlDataExtractor(self.bond_yaml_file)
+        bond_ingest_info = bond_extractor.extract_and_populate_data(bond_tab,
```
jamwalla: is it worth checking also if, after looping through the charges, there are any bonds from the table that did not match to any charge?

recidiviz/persistence/database/migrations/versions/3bf06134321e_add_chargedegree_forth.py
```diff
@@ -0,0 +1,45 @@
+"""add_ChargeDegree_FORTH
```
zdg2102: Can we break this out? I'd prefer to have the enum change and migration go in as its own PR first, rather than being lumped in with all these other changes

recidiviz/common/constants/bond.py
ohinds: yup!

recidiviz/ingest/scraper.py
ohinds: Great, I'll pull that in.

recidiviz/ingest/vendors/iml/iml_scraper.py
ohinds: Done

recidiviz/ingest/vendors/iml/iml_scraper.py
ohinds: This is moot given the rework to use xpaths instead of res.

recidiviz/ingest/vendors/iml/iml_scraper.py
ohinds: Same as above.

recidiviz/ingest/vendors/iml/iml_scraper.py
ohinds: Copy/paste error. Fixed.

recidiviz/common/constants/charge.py
ohinds: You canadian.

recidiviz/ingest/vendors/iml/iml_scraper.py
ohinds: I totally agree about hungarian notation in general, but I don't think this fits neatly into that. 'type' here doesn't refer to the datatype, but rather to the variety of task that's to be gotten. I guess I could call it task_variety, but that seems like just replacing this with a synonym.

recidiviz/ingest/vendors/iml/iml_scraper.py
ohinds: Yup, I just missed the return. Fixed.

recidiviz/ingest/vendors/iml/iml_scraper.py
ohinds: Same as above.

recidiviz/ingest/vendors/iml/iml_scraper.py
ohinds: I thought it was, but turns out no. Check added.

recidiviz/ingest/vendors/iml/iml_scraper.py
ohinds: Yup, done.

recidiviz/ingest/vendors/iml/iml_scraper.py
```diff
@@ -0,0 +1,349 @@
...
+        # Connect bonds and charges, and fill in an ICE hold when we
+        # see it in the charge name.
+        for charge in person.booking[0].charge:
+            if charge.name == 'IMMIGRATION DETAINEE':
+                person.booking[0].hold = 'ICE'
+
+            if not bond_ingest_info.person:
+                continue
+
+            this_charge = [c for c in
```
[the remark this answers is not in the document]
ohinds: So for this one I left the list comprehension because I want to test if we got more than one bond for a single charge. That seem ok?

recidiviz/ingest/vendors/iml/iml_scraper.py
```diff
@@ -0,0 +1,349 @@
...
+                           if c.case_number == charge.case_number]
+
+            if not this_charge:
+                continue
+
+            charge.bond = this_charge[0].bond
+            charge.judge_name = this_charge[0].judge_name
+
+            # Transfer the bond type to charge status, if we should.
+            _CHARGE_STATUSES = [
```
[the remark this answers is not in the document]
ohinds: This was so I could have the minimal universe of replacements that I've seen right here. I'll document that, unless you'd rather I include all possibilities we could handle?

recidiviz/ingest/vendors/iml/iml_scraper.py
```diff
@@ -0,0 +1,349 @@
...
+            'flow_action': 'searchbyname',
+            'quantity': self._num_people_per_search,
+            'systemUser_includereleasedinmate': 'N',
+            'systemUser_includereleasedinmate2': 'N',
+        }
+
+        return data
+
+    def _fetch_content(self, endpoint, post_data=None, json_data=None):
+        """Returns the page content. This is overriden from BaseScraper
```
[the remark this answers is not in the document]
ohinds: Done

recidiviz/persistence/database/migrations/versions/3bf06134321e_add_chargedegree_forth.py
```diff
@@ -0,0 +1,45 @@
...
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = '3bf06134321e'
+down_revision = '328448d2b4d2'
+branch_labels = None
+depends_on = None
+
+# Without 'FORTH'
```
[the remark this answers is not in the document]
ohinds: Done

recidiviz/ingest/vendors/iml/iml_scraper.py
ohinds: Could be? The check would need to be more robust than this, I imagine.

recidiviz/ingest/vendors/iml/iml_scraper.py
ohinds: Yup.

recidiviz/ingest/vendors/iml/iml_scraper.py
ohinds: This will happen, because some bonds don't have a case number. I'm not sure what to do with those, do you have ideas?

recidiviz/ingest/vendors/iml/iml_scraper.py
```diff
@@ -0,0 +1,349 @@
...
+specific navigation and data extraction. All counties that use IML should
+have their county-specific scrapers inherit from this class.
+
+Vendor-specific notes:
+    - No historical data is kept, only data on current people
+    - An initial request can be made to retrieve a page with all roster entries
+    - The person page contains the current booking info
+
+Background scraping procedure:
+    1. A get without session variables to find the total number of people to
```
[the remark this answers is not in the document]
ohinds: Done

recidiviz/persistence/database/migrations/versions/3bf06134321e_add_chargedegree_forth.py
ohinds: Sure thing.

recidiviz/ingest/vendors/iml/iml_scraper.py
jessex: Mm, that's fair, but I think the coincidence may be misleading. I think renaming to avoid the appearance of HN is worth the effort, personally, but feel free to ignore.

recidiviz/ingest/vendors/iml/iml_scraper.py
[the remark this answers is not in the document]
jessex: That's reasonable!

recidiviz/ingest/vendors/iml/iml_scraper.py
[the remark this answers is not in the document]
jessex: Hm. Personally, it seems that not having an unexpected bug if we miss a possibility in the future and not having to define another collection of statuses outweighs the benefit of this sort of documentation. If that's something we want to document, I'd rather have a comment somewhere saying "We expect these statuses specifically..."

recidiviz/ingest/vendors/iml/iml_scraper.py
ohinds: Done

recidiviz/ingest/vendors/iml/iml_scraper.py
ohinds: #697

recidiviz/ingest/vendors/iml/iml_scraper.py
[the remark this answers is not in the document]
ohinds: Ok, I'm on board. I made this change.

recidiviz/ingest/vendors/iml/iml_scraper.py
jamwalla: Oh, totally. Maybe we could add the log there as a warning and see if we can come back to this? I filed #696 because this is the second vendor we have this problem with.

Issue #596, opened 2019-01-11
Track next opencensus release


colincadams:
Using this issue to track the next opencensus release, which we need for the following reasons.

For metrics we need the following for exporting to stackdriver to work at all:
https://github.com/census-instrumentation/opencensus-python/pull/374
https://github.com/census-instrumentation/opencensus-python/pull/438

For traces we need the following for it to skip metadata requests:
https://github.com/census-instrumentation/opencensus-python/pull/289

Issue #59, opened 2018-04-18
Add metrics to ingest pipeline


cawarren:
From a comment Josh made on PR #46 -

- Number of new inmates, records, and snapshots created per session
- Number of existing inmates, records, and snapshots updated per session
- Number of each field that was updated in a new snapshot per session (would be super handy to help understand the latent volatility in a region's source data)
- Infrastructure metrics related to the task queue itself, e.g. number of retries, average idle time per worker, average wall time per task, etc.

First step will be to add metrics similar to the above to a session (there can be many sessions to a full scrape). Follow-up step will be aggregating up to a scrape.

cawarren:
_This is part of ingest generalization work_

Scope update:
- Read and determine how to use Google Stackdriver for custom metrics in Google App Engine. Some relevant links:
  - [Using Stackdriver logging in App Engine apps](https://cloud.google.com/appengine/articles/logging)
  - [Stackdriver : Using Custom Metrics](https://cloud.google.com/monitoring/custom-metrics/)
- Enable Google Stackdriver support in Cloud Console, set up configuration
- Add `metrics` module under `ingest` package (see the design doc's [Modules section](https://docs.google.com/document/d/11fv-hLwLOFcM5XSE16lYne5hnqFzcQK1vwvTBVVyhn8/edit#heading=h.v6780khuazyt))
- Define metrics if necessary, and provide helper functions for writing metrics from scrapers in the new `metrics` module
- Add metrics for the data types described in the design doc's [Monitoring section](https://docs.google.com/document/d/11fv-hLwLOFcM5XSE16lYne5hnqFzcQK1vwvTBVVyhn8/edit#heading=h.pq2010l3u22n)

colincadams:
Just found this thread. Based on [this discussion in OpenCensus](https://github.com/census-instrumentation/opencensus-go/issues/576), we should be able to handle alerting on cumulative metrics by creating some gauge views of them. Should work for now and hopefully in the future it will be supported fully.

colincadams:
stealing this issue for now to track monitoring

colincadams:
A first step is already in with #449, but it actually functioning is blocked by #596

arian487:
Whenever https://github.com/Recidiviz/recidiviz-data/pull/640 goes in, along with another PR handling error reporting on the ingest side I'm going to consider this done for the purposes of drop day as that is the most important.  We can add these other metrics over time.  Moving this out for now.

jessex:
[...]

PR #645, opened 2019-01-17
upgrade opencesus to git version

colincadams:
The timeline for getting the changes to opencensus out to pypi isn't as clear as we thought it was, do to some other behavior changes in between the prior release and my patches.

For now I think we should just install the latest from their repo and lock to that commit. Then once the next release is out that includes my changes we can move back to the pypi release.

pushes along #59 and #596


Pipfile
```diff
@@ -90,7 +90,7 @@ google-cloud-logging = "*"
 pymysql = "*"
 deepdiff = "*"
 mypy = "*"
-opencensus = "*"
+opencensus = {editable = true, git = "https://github.com/census-instrumentation/opencensus-python.git", ref = "989470edce8fa7a82813e3846f774fd4927b5b07"}
```
jessex: Do we want to put a TODO on this? I think we can comment in a Pipfile.

Issue #129, opened 2018-09-23
Ingest should support ingesting aggregate-level population and admissions statistics


ghost:
Currently ingestion only supports individual-level data. However many jurisdictions publish reports containing aggregated statistics at a variety of time scales. Ingesting these will greatly increased the breadth and quality of the incarceration data we can provide.

Examples:
Daily KY DOC pop: https://corrections.ky.gov/About/researchandstats/pages/dailycount.aspx)
Weekly KY jail pop:https://corrections.ky.gov/About/researchandstats/pages/weeklyjail.aspx)
Monthly TX jail pop: https://www.tcjs.state.tx.us/docs/AbbreRptCurrent.pdf
Yearly FL DOC pop: http://www.dc.state.fl.us/pub/annual/1617/FDC_AR2016-17.pdf

jessex:
The high-level design question, in the context of our current architecture, is how to fit this into the world of ingest-and-calculate? One could create a scraper for aggregate data, write it to some aggregate-centric modeling within the ingest schema, and then write calculation logic which converts those stored aggregates into fully modeled `*Metric` calculations. One could also smush ingest and calculate together, transforming ingested aggregates into `*Metric` calculations on the fly. The latter seems simpler but we should think about it further to see if there's anything we might miss by skipping the end-to-end pipeline, and how we want to structure this in actual code.

arian487:
Closing in favour of https://github.com/Recidiviz/recidiviz-data/issues/804, nearly done!

PR #648, opened 2019-01-17
Add parsing logic for aggregated FL

zacharylawrence:
This is a first pass at adding state level aggregate report PDF parsing to our codebase. The idea behind this change is that a follow-up change will be added that uses this parsed PDF data to write directly to our database.

#129


recidiviz/aggregate_ingest/aggregate_ingest_utils.py
```diff
@@ -0,0 +1,25 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+# =============================================================================
+"""Common utility functions used across aggregated_ingest."""
+import pandas as pd
+
+
+def collapse_header(columns: pd.MultiIndex) -> pd.MultiIndex:
+    """Strip empty information from a multi-index header."""
+    words_to_keep = [filter(lambda word: 'Unnamed' not in word, words)
```
jessex: You're calling that var 'Unnamed' because it doesn't matter what the string value is, right? If so, would it be more Pythonic to call it '_'?

recidiviz/tests/aggregate_ingest/fl/fixtures/__init__.py
```diff
@@ -0,0 +1,16 @@
...
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+# =============================================================================
```
jessex: I just want to make 100% sure that there's no sensitive information in the PDF fixtures we are placing in here unmodified, and that there are no terms of service on their respective hosting sites that prohibit their scraping or their placement in a repository like this?

recidiviz/aggregate_ingest/fl/fl_aggregate_ingest.py
```diff
@@ -0,0 +1,72 @@
...
+    return {
+        'fl_county_adp': _parse_table_1(filename)
+    }
+
+
+def _parse_table_1(filename: str) -> pd.DataFrame:
+    """Parses the FL County ADP - Table 1 in the PDF."""
+    part1 = tabula.read_pdf(
+        filename,
+        pages=[3],
```
jessex: Is there a more dynamic way to select where to start and end in Tabula?

recidiviz/aggregate_ingest/fl/fl_aggregate_ingest.py
```diff
@@ -0,0 +1,72 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+# =============================================================================
+"""Parse the FL Aggregated Statistics PDF."""
+import locale
+from typing import Dict
+
+import pandas as pd
+import tabula
```
jessex: Is this the same [Tabula](https://github.com/chezou/tabula-py) we're using here? I'm curious about whether we'll run into issues with the introduction of a JVM runtime dependency.

recidiviz/aggregate_ingest/aggregate_ingest_utils.py
```diff
@@ -0,0 +1,25 @@
+# Recidiviz - a platform for tracking granular recidivism metrics in real time
```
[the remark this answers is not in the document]
zacharylawrence: I like this approach and think it makes a lot of sense to me overall! This seems like a large refactor, so I made #652 to track it. For now, I went ahead and refactored this change into `recidiviz/ingest/aggregate/`.

zacharylawrence:
Note that writing the result of the parsed PDF is done in the seperate PR #677.

recidiviz/aggregate_ingest/aggregate_ingest_utils.py
jessex: Ah my bad. I very much misread that lambda.

recidiviz/aggregate_ingest/fl/fl_aggregate_ingest.py
jessex: Yeah there's less structure to latch onto in a raw PDF. I believe these are really consistent, changing very slowly. But I'm worried about tables that may be of variable length and wind up spilling into a new page on occasion. This could happen if the row scheme is something that changes, like a table oriented around the kinds of releases or charges instead of county. It could also happen if there's a rendering mistake and a table goes too far horizontally and breaks to a newline. 

I suppose we can wait to see how this performs on historical reports to generate bugs if they exist.

recidiviz/tests/ingest/fixtures.py
```diff
@@ -57,3 +56,18 @@ def as_dict(region_directory, filename):
     """
     contents = as_string(region_directory, filename)
     return json.loads(contents)
+
+
+# TODO(647): Refactor this to be usable outside the `ingest` module
```
ohinds: If you say #647 instead, it automatically puts a mention in the issue that can help you track the TODOs to be resolved when the issue is closed.

zacharylawrence: Awesome! Done.

recidiviz/aggregate_ingest/fl/fl_aggregate_ingest.py
zacharylawrence: Looking through the historical PDFs for 7 states, I don't see any cases of malformed tables. As we're writing this parsing logic, I think we should design for cases that fail loudly when a PDF can't be correctly parsed. In this case, future changes will parse the pages 4+ which would fail if this table leaked onto that page.

I think it makes sense to move forward with hard coding the pages here. I've made #684 to track adding guards to ensure that each PDF matches the expected requirements.

recidiviz/tests/aggregate_ingest/fl/fixtures/__init__.py
ohinds: Chris gives a thumbs up to checking these in.

zacharylawrence: Awesome, thanks for checking @ohinds!

PR #653, opened 2019-01-17
Adds admission reason to Booking and move `INFERRED_RELEASE` and `REMOVED_FROM_SOURCE` into CustodyStatus

### LINKED ISSUE #536: There should be an `admission_reason` field in booking
Let's start with an enum with values:

From Escape
New Commitment
Parole Violation
Probation Violation
Transfer

terinpw:
Closes #536


recidiviz/common/constants/booking.py
```diff
@@ -21,6 +21,18 @@
 from recidiviz.common.constants.mappable_enum import MappableEnum
 
 
+class AdmissionReason(MappableEnum):
+    FROM_ESCAPE = enum_strings.admission_reason_from_escape
```
zdg2102: nit: can we just call this `ESCAPE`?

recidiviz/common/constants/enum_canonical_strings.py
```diff
@@ -58,23 +58,29 @@
 
 # Booking
 
+admission_reason_from_escape = 'FROM_ESCAPE'
```
zdg2102: Same here

recidiviz/ingest/models/ingest_info.proto
```diff
@@ -48,6 +48,7 @@ message Person {
 message Booking {
 	optional string booking_id = 1;
 	optional string admission_date = 2;
+    optional string admission_reason = 14;
```
zdg2102: nit: this doesn't seem properly indented

recidiviz/common/constants/booking.py
terinpw: @ohinds to comment -- took it directly from issue #536, but that would work for me.

recidiviz/ingest/models/ingest_info.proto
terinpw: was a tab/space issue that looked fine on my machine. Updated file to use spaces as that's what our codebase does.

Issue #585, opened 2019-01-10
add INFERRED_DROP to droppable fields


jamwalla:
@zdg2102 @terinpw once we make all the decisions about #573, we will probably want a way of distinguishing between objects whose 'dropped' status was scraped vs. inferred

terinpw:
For Booking we've already determined we do the following when a booking doesn't show up on a scrape. 
```
release_reason: INFERRED_RELEASE
release_date: date of last scrape
release_date_inferred: True 
```

But what field/fields should we do for: Hold, Arrest, Charge, Sentence, Bond

ohinds:
I think just leave all the fields in hold, arrest, charge, and sentence the way they were. We have no further information, so just leave it?

jamwalla:
To be clear, we're talking about what an inferred drop would look like for each of those objects. E.g. if there are fewer charges on a subsequent scrape, do we change the charge status? mark the charge as 'drop_inferred'? etc.

terinpw:
Chatted with Oliver offline and to recap our conversation his points were:

When a booking is resolved:
- We only mark the booking INFERRED_RELEASE (nothing happens to the children beneath it b/c we can't assume anything about Charges/Holds/Bonds/ or Sentences)

When an entity is removed from a booking in successive scrapes we should update following fields on the entities:

Charge: status changes to `INFERRED_DROPPED`
Hold: status changes to `INFERRED_DROP`
Bond: status changes to `UNKNOWN_REMOVED_FROM_SOURCE`
Sentences: status changes to `UNKNOWN_REMOVED_FROM_SOURCE`

jamwalla:
cc @zdg2102

PR #660, opened 2019-01-17
Add SentenceStatus field and create generic UNKNOWN_REMOVED_FROM_SOURCE enum value

terinpw:


recidiviz/common/constants/enum_canonical_strings.py
```diff
@@ -37,6 +37,7 @@
 # default value for enums that are not provided (which should be represented
 # with None/NULL).
 external_unknown = 'EXTERNAL_UNKNOWN'
+unknown_removed_from_source = 'UNKNOWN_REMOVED_FROM_SOURCE'
```
zdg2102: Might be worth adding a comment clarifying what this is, like `external_unknown` has

recidiviz/common/constants/enum_canonical_strings.py
```diff
@@ -105,6 +106,12 @@
 bond_status_active = 'ACTIVE'
 bond_status_posted = 'POSTED'
 
+# Sentence
+
+sentence_status_commuted = 'COMMUTED'
+sentence_status_completed = 'COMPLETED'
```
zdg2102: Probably shouldn't address this in this PR, but should we remove `SENTENCE_COMPLETED` from `ChargeStatus` and instead create a dummy sentence with `SentenceStatus.COMPLETED`?

recidiviz/common/constants/enum_canonical_strings.py
terinpw: Added!

recidiviz/common/constants/enum_canonical_strings.py
terinpw: Filed issue #663

terinpw:
helps with #585

PR #665, opened 2019-01-17
Update enum values for BondType, BondStatus, and ChargeClass

zdg2102:
Also update converter to correctly handle cases where BondType and BondStatus have overlapping information


recidiviz/persistence/converter/converter_utils.py
```diff
@@ -158,17 +159,24 @@ def parse_days(time_string):
...
+    is not a numeric value, checks if the text value contains information about
+    bond type or bond status.
 
-    if 'DENIED' in amount.upper():
-        return None, BondType.BOND_DENIED
+    Returns:
+        Tuple of bond amount, bond type, and bond status.
+    """
+    bond_type = BOND_TYPE_MAP.get(amount.upper(), None)
+    bond_status = BOND_STATUS_MAP.get(amount.upper(), BondStatus.INFERRED_SET)
```
jamwalla: Oliver and I agreed that `BondStatus.(INFERRED_)SET` shouldn't accompany `BondType.NO_BOND`. We should enforce this in the converter.

I *think* that might be something we have to ask all or most scrapers to tell us - if we see 'no bond', an empty bond field, '$0.00', etc., does the website mean `BondStatus.DENIED` or `BondStatus.NOT_REQUIRED`?

zdg2102: I'm now using `UNKNOWN_FOUND_IN_SOURCE` everywhere that `BondType` is `NO_BOND` and we can't explicitly set `BondStatus`, and I'm reserving `INFERRED_SET` only for the case where the only value provided is a numeric value (where we also infer `BondType.CASH`).

We can get a final weigh-in on this approach from @ohinds

PR #677, opened 2019-01-17
Add fl_aggregate_ingest write logic

zacharylawrence:
This is a first pass at writing parsed Aggregated PDF data into our Database. In future changes, parsing the PDF then writing the resulting DataFrame into the Database will be done by a Cloud Function.

#129


recidiviz/persistence/database/database.py
```diff
@@ -171,6 +175,32 @@ def update_booking(session, booking_id, **kwargs):
...
+
+def _write_df_only_successful_rows(
+        table: DeclarativeMeta, df: pd.DataFrame) -> None:
+    """If the dataframe can't be written all at once (eg. some rows already
+    exist in the database) then we write only the rows that we can."""
+    for i in range(len(df)):
+        try:
+            df.iloc[i:i + 1].to_sql(table.__tablename__, recidiviz.db_engine,
+                                    if_exists='append', index=False)
+        except IntegrityError:
```
jessex: Let's log the rows that we have already written.

recidiviz/persistence/database/database.py
```diff
@@ -171,6 +175,32 @@ def update_booking(session, booking_id, **kwargs):
         .update(_convert_enums_to_strings(kwargs))
 
 
+def write_df(table: DeclarativeMeta, df: pd.DataFrame) -> None:
+    """
+    Writes the |df| to the |table|.
```
jessex: Should we document that this behaves the same as the below method? It tries to write everything at once but if anything in there fails, it will run the method below and write all of the rows it can, one by one.

recidiviz/persistence/database/schema.py
```diff
@@ -550,3 +550,13 @@ class ChargeHistory(Base):
     case_number = Column(String(255))
     next_court_date = Column(Date)
     judge_name = Column(String(255))
+
+
+class FlCountyAdp(Base):
+    __tablename__ = 'fl_county_adp'
```
jessex: @ohinds is there value in having all of the ADP states in the same table for easy querying across states? Or are they not consistent enough, or are we planning to organize them for cross-state access in BigQuery, etc.?

recidiviz/tests/persistence/database/database_test.py
```diff
@@ -34,6 +38,9 @@
 _EXTERNAL_ID = 'external_id'
 _BIRTHDATE = datetime.date(year=2012, month=1, day=2)
 
+# TODO(676): Set this based on Cloud Function metadata
```
ohinds: '#' here too

PR #688, opened 2019-01-18
Define schema for state-provided aggregate statistics

zacharylawrence:
#129
Co-authored-by: arian487


zacharylawrence:
@ohinds @jessex This is a first pass at defining the schema we'll use to store state-provided aggregate statistics. The intention is to build further aggregated bigquery views on this data.

recidiviz/persistence/database/schema.py
```diff
@@ -587,3 +590,205 @@ class ChargeHistory(Base, DatabaseEntity):
...
+class HiFacilityAggregates(Base, _AggregateTableMixin):
+    """HI state-provided aggregate statistics."""
+    __tablename__ = 'hi_facility_aggregates'
+
+    facility_name = Column(String(255))
+
+    design_bed_capacity = Column(Integer)
+    operation_bed_capacity = Column(Integer)
+
+    total_population = Column(Integer)
```
zacharylawrence: Note that here we elected to store the `total_population` despite being able to calculate it from other "more raw" aggregated metrics we also store (eg. `male_population` + `female_population`). We decided not to store these "aggregated aggregates" for other fields. (eg. we don't store `held_for_other_jurisdiction_population` which would be `held_for_other_jurisdiction_male_population` + `held_for_other_jurisdiction_female_population`.

@jessex @ohinds Any thoughts on this?

recidiviz/persistence/database/schema.py
```diff
@@ -587,3 +590,205 @@ class ChargeHistory(Base, DatabaseEntity):
...
+
+    # TODO(#689): Ensure that `fips` also supports facility level fips
+    # TODO(#689): Consider adding fips_type to denote county vs facility
+    fips = Column(String(255), primary_key=True)
+    report_date = Column(DateTime, primary_key=True)
+    report_granularity = Enum(*report_granularity_values,
+                              name='report_granularity', primary_key=True)
+
+
+class FlCountyAggregates(Base, _AggregateTableMixin):
```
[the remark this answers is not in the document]
jessex: I think the class name should be singular. Each instance corresponds to a single aggregate value.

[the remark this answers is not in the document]
zacharylawrence: Yeah, that seems inline with the rest of our tables too (eg. `Charge`). Updated!

recidiviz/persistence/database/schema.py
```diff
@@ -587,3 +590,205 @@ class ChargeHistory(Base, DatabaseEntity):
...
+    """FL state-provided aggregate statistics."""
+    __tablename__ = 'fl_county_aggregates'
+
+    county_name = Column(String(255))
+    county_population = Column(Integer)
+    average_daily_population = Column(Integer)
+    date_reported = Column(DateTime)
+
+
+class FlFacilityPretrialAggregate(Base, _AggregateTableMixin):
```
zacharylawrence: @jessex @ohinds Thoughts on this table name specifically? Should we specify `Pretrial` here or is just `FlCountyAggregate` and `FlFacilityAggregate` what we want here?

jessex: I would cut out Pretrial and have everything follow a pattern of `[State][Unit]Aggregate` where Unit is either Facility or County or whatever the jurisdictional unit of measurement is. I would use `pretrial` in the field names of the aggregate.

recidiviz/persistence/database/schema.py
jessex: I think we should have a consistent strategy for under what conditions we will sum up ingested aggregates prior to persistence. Possibilities:
* No additional aggregation at all, only store what you parse. Do any post-processing in BigQuery.
* Additional aggregation for anything of worth so it can be pivoted directly from SQL to BigQuery.
* No additional aggregation in general, but only in some special cases. We just need to define what those cases could be and document them for posterity. It sounds like in this case, the rationale is that a "total population" metric is critical to all aggregate types and that if we don't have it, we should roll it up if possible. If that's the rationale, then I think it's sound but think we should document it.

arian487: Agreed, we'll add a comment for this I think in general we shouldn't do any extra aggregation in these base tables, but Zach and I agreed total_population felt like a special case that we were willing to break the rule for.

arian487: I've added a comment at the top

recidiviz/persistence/database/schema.py
arian487: Done.

PR #691, opened 2019-01-22
Update us_pa_greene scraper from ArchonixScraper to JailTrackerScraper

KareemG:
Greene County PA updated their inmate lookup tool, and are now using JailTracker. This was linked to on http://www.co.greene.pa.us/secured/gc2/depts/lo/prison/prison.htm under "Inmate Lookup".

P.S. I'm new here, this is my first PR! Am eager to receive feedback/pointers.


recidiviz/ingest/us_pa_greene/us_pa_greene_scraper.py
```diff
@@ -15,19 +15,41 @@
...
+            'SUMMARY': ChargeClass.INFRACTION,
+
+            # Charge Statuses
+            'SENTENCED BY COMMONWEALTH': ChargeStatus.SENTENCED,
+
+            # Bond Types
+            'CASH/SURETY': BondType.UNSECURED,
+            'STRAIGHT': BondType.CASH,
+            'NO BOND PERMITTED (UNBONDABLE CHARGE)': BondType.NO_BOND,
+            'PERCENTAGE': None,
```
KareemG: Not sure what this is just yet. Could be a surety or a [10 Percent Bond](https://definitions.uslegal.com/t/ten-percent-bond/).

KareemG:
Looking for advice regarding tests before I make my own custom tests/fixtures. It looks like tests/ingest/us_vt is a valid test setup for most JailTracker scrapers - should this be refactored?

jamwalla:
This looks great! We should do some refactoring, but we'd like to have separate tests for each child scraper, even if they look pretty much the same because the vendor is the same. If the us_vt test works here, great; there's also a test template that you can use by running `python3 create_scraper.py greene pa jail -tests_only`.

KareemG:
Added tests! I'm not sure how to treat PERCENTAGE. There isn't a lot of data that points to anything from their JailTracker table. Have kept it as None for now.

recidiviz/ingest/us_pa_greene/us_pa_greene_scraper.py
arian487: @ohinds any idea what percentage is?

jamwalla:
@KareemG, we changed our deploy process in a way that broke travis on other forks of the project. Can you try patching #718 onto this and see if we can get travis to run?

jamwalla:
Cool, looks like travis is happy. I went with a different approach in #718 which is now merged into master, can you merge from master again? then we should be able to commit this. Like Arian said, really great find!

recidiviz/ingest/us_pa_greene/us_pa_greene_scraper.py
jamwalla: Let's use `BondType.CASH` for now.

jamwalla:
Looks great, I'm going to merge this in! Awesome work.

PR #692, opened 2019-01-22
Make IngestInfo fields optional and add type annotations to populate_data

colincadams:


recidiviz/ingest/us_fl_hendry/us_fl_hendry_scraper.py
```diff
@@ -74,22 +76,23 @@ def populate_data(self, content, params, ingest_info):
...
+                            charge.name = charge_search.group(2)\
+                                .replace('\n', '')\
+                                .replace('\t', '')\
+                                .replace(u'\xa0', u'')\
+                                .strip()
 
-                        if charge.name == '':
-                            charge.name = None
+                            if charge.name == '':
+                                charge.name = ''
```
colincadams: Here I had to set `''` instead of `None` to keep mypy happy. Should we instead make all of the fields in `IngestInfo` optional?

zacharylawrence: For what it's worth, that's what we do in the [Entities classes](https://github.com/Recidiviz/recidiviz-data/blob/2fc0ec40a90bdcfa6f2b19454745896a85f34e97/recidiviz/persistence/entities.py).

PR #700, opened 2019-01-22
Add 2nd table to FL aggregate parser

zacharylawrence:
This PR is a continuation of #648 by adding parsing logic for the 2nd table. This completes all FL aggregate scrape logic because the 3rd and final table will not be parsed (since it contains no county level data).

#129


recidiviz/ingest/aggregate/regions/fl/fl_aggregate_ingest.py
```diff
@@ -80,6 +89,57 @@ def _parse_table_1(filename: str) -> pd.DataFrame:
     return result
 
 
+def _parse_table_2(filename: str) -> pd.DataFrame:
```
arian487: I'd be more descriptive in the naming here, maybe 'parse_facility_pretrial' table

arian487: Similarly I'd change the name above to be a bit more descriptive too.

recidiviz/ingest/aggregate/regions/fl/fl_aggregate_ingest.py
```diff
@@ -80,6 +89,57 @@ def _parse_table_1(filename: str) -> pd.DataFrame:
...
+        pages=[6],
+        pandas_options={
+            'skiprows': [0, 1, 2],
+            'usecols': [0, 2, 3, 4, 5],  # Column 1 contains no data
+            'names': column_names,
+            'skipfooter': 2  # The last 2 rows are the totals
+        })
+    result = part1.append(part2, ignore_index=True)
+
+    result = result[['facility_name', 'average_daily_population',
```
jessex: Are you going to apply the change from #703 here, too, after that's merged and before this one?

zacharylawrence: Yup, good idea. Updated.

PR #704, opened 2019-01-22
Add GA aggregate scraper

zacharylawrence:
#129


recidiviz/ingest/aggregate/regions/ga/ga_aggregate_ingest.py
```diff
@@ -0,0 +1,105 @@
...
+        'Number of Inmates Awaiting Trial in Jail [% of Total]',
+        'Number of Inmates Serving County Sentence [Number]',
+        'Number of Inmates Serving County Sentence [% of Total]',
+        'Number of Other Inmates [Number]',
+        'Number of Other Inmates [% of Total]'
+    ]
+
+    result = tabula.read_pdf(
+        filename,
+
```
arian487: I dont' think we need a newline in between each keyword?

zacharylawrence: Good idea. Done.

PR #706, opened 2019-01-22
support cookies natively in base_scraper

colincadams:


recidiviz/ingest/base_scraper.py
```diff
@@ -74,23 +75,32 @@ def get_initial_task_method(self):
...
         """Returns the page content.
 
         Args:
             endpoint: the endpoint to make a request to.
             headers: dict of headers to send in the request.
+            cookies: dict of cookies to send in the request.
             post_data: dict of parameters to pass into the html request.
 
         Returns:
             Returns the content of the page or -1.
```
jessex: We should update this doc since we're now returning a tuple of this and cookies from the response.

colincadams: Fixed

Issue #675, opened 2019-01-17
finish writing scraper tests


jamwalla:
- [ ] us_ar_van_buren
- [ ] us_mo_stone
- [ ] us_mt_gallatin
- [ ] us_nc_guilford

and/or
- [ ] archonix
- [ ] jailtracker
- [ ] newworld

jamwalla:
We should follow the approach in #743 and write base test classes for vendors if the tests for child scrapers are redundant. This way each child scraper will be tested separately, but the test files they're working over will be centralized.

jamwalla:
@eitalse started on DCN

Issue #527, opened 2019-01-07
JSON data extractor


jamwalla:
- [x] expect only `key_mappings`
- [ ] process keys with lists of strings as `multi_key_mappings`

jamwalla:
It seems like we could need `multi_key`-ish behavior from weird data like a list of charge names that should be zipped with a list of statute codes. We wrote this code but didn't commit it yet..
```python
if lookup_key in self.multi_keys:
    assert isinstance(v, list)
    class_to_set, ingest_key = self.multi_keys[lookup_key].split('.')
    self._set_or_create_object(ingest_info, class_to_set, ingest_key, v)
```

jamwalla:
Closing this since in #707 we decided a list of strings should create a new object for each string (eg a list of charge names).

PR #707, opened 2019-01-22
EagleAdvantage vendor scraper

jamwalla:


recidiviz/ingest/vendors/eagle_advantage/eagle_advantage_scraper.py
```diff
@@ -0,0 +1,127 @@
...
+                else:
+                    logging.error("Charge information did not match expected "
+                                  "pattern: %s", charge.name)
+                    continue
+
+                charge.charge_id = c_id
+                charge.statute = code
+                charge.name = off1 + off2
+                charge.charge_class = c_class
+                # charge.create_sentence()
```
jessex: Is this to be done later?

jamwalla: yes, coming soon! This looks like a general 'notes' field so we'll do what we can with it

recidiviz/ingest/extractor/json_data_extractor.py
```diff
@@ -79,7 +79,7 @@ def _set_value_if_key_exists(self, lookup_key, value, ingest_info):
 
     def _extract_list(self, content, ingest_info, current_key=None):
         for value in content:
-            if not isinstance(value, list) and not isinstance(value, dict):
-                logging.error('JSON value was not an object or array: %s',
-                              value)
-            self._extract(value, ingest_info, current_key)
+            if isinstance(value, (list, dict)):
```
jamwalla: @arian487 - we originally talked about lists of strings in JSON as meaning columns for `multi_key_mappings` (#527). I think this might be a more common case, though: there is a list of strings containing charge names, and we should create a charge for each string. What do you think about modifying the JsonDataExtractor this way?

recidiviz/ingest/vendors/eagle_advantage/eagle_advantage.yaml
```diff
@@ -0,0 +1,20 @@
...
+  Inmates.BookedStatus: booking.custody_status
+  Inmates.Charges: charge.name
+  Inmates.ReleaseDate: booking.release_date
+  Inmates.Bond: bond.amount
+  Inmates.Address: person.place_of_residence
+
+keys_to_ignore:
+  - Inmates.DaysInJail
+  - Inmates.Location # Cell, Hall, Bed
+  - Inmates.History # List of previous book/release dates
```
jamwalla: #708

recidiviz/ingest/models/ingest_info.py
```diff
@@ -335,6 +336,8 @@ def to_repr(obj):
 
 
 def restricted_setattr(self, last_field, name, value):
+    if isinstance(value, str) and (value == '' or value.isspace()):
```
jamwalla: @zacharylawrence we talked about what the converter should do if it gets empty strings from IngestInfo, and had last settled on ignoring the error from `normalize` and converting empty strings to `None`. On second thought I think it might be better handled in IngestInfo. wdyt?

jamwalla:
I think this is ready for review, with the caveat that the text now going into `Charge.charge_notes` could tell us lots of stuff about charges/bonds/sentences and I'm only handling the few things I saw consistently in the two child scrapers. @jessex @ohinds @arian487 can you take a look? @ohinds if you esp. could double check the CourtType enum mappings. Thanks!!

recidiviz/ingest/models/ingest_info.py
zacharylawrence: Yeah, I like this look! Probably makes sense to also through the exception in the Data Converter though?

jamwalla: Cool! It's still handled in converter_utils, I undid that change.

recidiviz/ingest/us_tn_mcminn/us_tn_mcminn_scraper.py
```diff
@@ -0,0 +1,64 @@
...
+                                                                    task,
+                                                                    ingest_info)
+        if scraped_data:
+            ingest_info = scraped_data.ingest_info
+
+        for person in ingest_info.people:
+            for charge in person.bookings[0].charges:
+                if not charge.charge_notes:
+                    continue
+                if 'NOLLE' in charge.charge_notes or \
```
jessex: What does this come from exactly? And how full are these notes? Could this text show up as a false positive somehow?

recidiviz/tests/ingest/us_tn_mcminn/fixtures/page.json
```diff
@@ -0,0 +1,60 @@
...
+      "Bond": "1234.56",
+      "Address": "SPRINGFIELD",
+      "Location": "Cell:XX, Hall:XX, Bed:XX",
+      "History": [
+        "book:1/1/2009,release:1/1/2018"
+      ]
+    }
+  ],
+  "TotalCount": 21
+}
```
jessex: Missing a newline here.

recidiviz/ingest/vendors/eagle_advantage/eagle_advantage_scraper.py
```diff
@@ -0,0 +1,122 @@
...
+        agency_service_port"""
+        super(EagleAdvantageScraper, self).__init__(region_name)
+        self.yaml_file = os.path.join(
+            os.path.dirname(__file__), 'eagle_advantage.yaml')
+        self.agency_service_ip = agency_service_ip
+        self.agency_service_port = agency_service_port
+
+    def populate_data(self, content, task: Task,
+                      ingest_info: IngestInfo) -> Optional[ScrapedData]:
+        """NOTE: Child scrapers should likely write their own populate_data
```
jessex: And in our experience so far, there's enough variability from county to county that we shouldn't handle it at the vendor level?

recidiviz/ingest/us_tn_mcminn/us_tn_mcminn_scraper.py
jamwalla: There's a `Charges` field with a bunch of key/value pairs, like code, court type, etc. There's one called `sent` that has free text that seems to get entered manually; Oliver suggested trying to parse what we can and storing the free text. Here's a doc with all the values of this field currently on the two websites (team only): https://docs.google.com/document/d/1laMb5U4JRnWmc4f1bDdntPMH3cR6jSD8tYv1G6-YwG0/edit?usp=sharing

jamwalla: Ahh, sorry, I thought you'd have the right permissions from the drive folder. Can you see it now?

jessex: Thanks, @jamwalla! Looking at that document was like taking a brief glimpse into the abyss, what with the ample jargon and spelling mistakes and inconsistency in ordering. Looking forward to eventually building an engine to understand general-purpose criminal justice notes!

Anyway, that's helpful to see, and I'm on board with storing the free text. I wonder in general with free text fields, though, how much we should try to parse/structure in production workflows versus leaving it as task at analysis time? Doing piecemeal parsing of only a couple of symbols seems like it could be brittle if we don't have really high confidence in the incidence patterns of those symbols.

jamwalla: Ugly, right!? :P

Agreed, it's hard to know just from one scrape what this field will look like, and since it looks sort of handwritten it could change a lot too. @ohinds suggested trying to get some benefit out of the field at scrape time, but I removed it for now. Oliver, do you think we should a) include the analysis in this change, b) look over the charge_notes for a longer period of scraping time and then decide, c) punt this until someone goes back and looks at this manually, d) something else?

recidiviz/ingest/us_tn_mcminn/us_tn_mcminn_scraper.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def __init__(self):
+        super(UsTnMcminnScraper,
+              self).__init__('us_tn_mcminn',
+                             agency_service_ip='mcminnsheriff.serveftp.net',
+                             agency_service_port=9000)
+
+    def get_enum_overrides(self):
+        return {
+            'O': ChargeClass.OTHER,
```
arian487: Can this be in the top level enum overrides (or even the global map?)

recidiviz/ingest/us_tn_mcminn/us_tn_mcminn_scraper.py
```diff
@@ -0,0 +1,39 @@
...
+            'GENERALSESSIONS': CourtType.DISTRICT,
+            # JUVENILE seems to mostly refer to child support charges
+            'JUVENILE': CourtType.DISTRICT
```
arian487: Global map?

recidiviz/ingest/vendors/eagle_advantage/eagle_advantage_scraper.py
```diff
@@ -0,0 +1,122 @@
...
+from recidiviz.ingest.task_params import Task, ScrapedData
+
+_ENDPOINT = 'http://offendermiddleservice.offenderindex.com/api/Values'
+_PAGE_SIZE = 10
+
+
+class EagleAdvantageScraper(BaseScraper):
+    """Vendor scraper implementation for Eagle Advantage."""
+
+    @abstractmethod
```
ohinds: Is this just to prevent accidental direct instantiation of the vendor scraper? Or is there some other reason this is abstract?

recidiviz/ingest/vendors/eagle_advantage/eagle_advantage_scraper.py
```diff
@@ -0,0 +1,122 @@
...
+                    c_id, code, off1, off2, c_class, sent, court, counts = \
+                        matches.groups()
+                else:
+                    logging.error("Charge information did not match expected "
+                                  "pattern: %s", charge.name)
+                    continue
+
+                charge.charge_id = c_id
+                charge.statute = code
+                charge.name = off1 + off2
```
ohinds: Do we want to delimit that manually somehow? Like a ':' or ' - '?

recidiviz/ingest/vendors/eagle_advantage/eagle_advantage_scraper.py
```diff
@@ -0,0 +1,122 @@
...
+                charge.charge_class = c_class
+                charge.charge_notes = sent
+                charge.court_type = court
+                charge.number_of_counts = counts
+
+        return ScrapedData(ingest_info)
+
+    def get_more_tasks(self, content, task: Task) -> List[Task]:
+        total_count = int(content['TotalCount'])
+        total_pages = math.ceil(total_count / 10)
```
ohinds: / _PAGE_SIZE?

ohinds: Also, this is integer division, right? so if we had 73 people, wouldn't we have:
73 / 10 --> 7 pages instead of 8?

recidiviz/ingest/vendors/eagle_advantage/eagle_advantage_scraper.py
```diff
@@ -0,0 +1,122 @@
...
+        return [
+            Task(endpoint=self._get_next_page_url(page_number),
+                 task_type=constants.TaskType.SCRAPE_DATA,
+                 response_type=constants.ResponseType.JSON)
+            for page_number in range(1, total_pages + 1)
+        ]
+
+    def _get_next_page_url(self, page_number: int) -> str:
+        post_data = {
+            'three': 1,
```
ohinds: lol. wtf.

recidiviz/ingest/vendors/eagle_advantage/eagle_advantage_scraper.py
```diff
@@ -0,0 +1,122 @@
...
+            'pageSize': _PAGE_SIZE
+        }
+        # Weirdly, the website doesn't accept POST requests, so encode manually.
+        endpoint = '{}?{}'.format(_ENDPOINT, urlencode(post_data))
+        return endpoint
+
+    def get_initial_task(self) -> Task:
+        endpoint = self._get_next_page_url(1)
+        return Task(endpoint=endpoint,
+                    task_type=constants.TaskType.GET_MORE_TASKS,
```
ohinds: Should this be INITIAL_AND_MORE?

recidiviz/tests/ingest/us_tn_mcminn/us_tn_mcminn_scraper_test.py
```diff
@@ -0,0 +1,106 @@
+# Recidiviz - a platform for tracking granular recidivism metrics in real time
```
ohinds: Nothing you should change here, but I'm noticing more how child vendor tests tend to share a good deal of code. We should probably think about creating some vendor-specific utils so we can simplify some of these.

recidiviz/ingest/vendors/eagle_advantage/eagle_advantage_scraper.py
jessex: It's wrapped in a `ceiling` call, so it will go up to the next integer if it doesn't divide evenly, i.e. `ceil(73 / 10) = 8`.

recidiviz/ingest/us_tn_mcminn/us_tn_mcminn_scraper.py
jamwalla: done

recidiviz/ingest/vendors/eagle_advantage/eagle_advantage_scraper.py
jamwalla: nope, this was leftover from a worse approach. removed, thanks!

recidiviz/ingest/vendors/eagle_advantage/eagle_advantage_scraper.py
jamwalla: nope, they'll just concatenate the two strings on the site. for example, `off1:LEGEND DRUGS, POSSESS WITHOUT PRESCRIPTI,off2:ON`

recidiviz/ingest/vendors/eagle_advantage/eagle_advantage_scraper.py
jamwalla: yep, python3 division produces a float!

recidiviz/ingest/vendors/eagle_advantage/eagle_advantage_scraper.py
jamwalla: no idea XD

recidiviz/ingest/vendors/eagle_advantage/eagle_advantage_scraper.py
jamwalla: yep!

recidiviz/ingest/vendors/eagle_advantage/eagle_advantage_scraper.py
```diff
@@ -0,0 +1,120 @@
...
+        """NOTE: Child scrapers may need write their own populate_data
+        function that calls this one and then processes charge.charge_notes.
+        The field may have information about charges, bonds, or sentences;
+        the child scraper should attempt to process the more common values."""
+        extractor = JsonDataExtractor(key_mapping_file=self.yaml_file)
+        ingest_info = extractor.extract_and_populate_data(content, ingest_info)
+
+        for person in ingest_info.people:
+            if len(person.bookings) != 1:
+                logging.error("Person did not have exactly 1 booking as "
```
zacharylawrence: Consider `continue` after this, otherwise indexing into `person.bookings[0]` will cause an exception.

recidiviz/ingest/vendors/eagle_advantage/eagle_advantage_scraper.py
```diff
@@ -0,0 +1,120 @@
...
+                              "expected")
+            for charge in person.bookings[0].charges:
+                regex: Any = re.compile(r'Id:(.*?),code:(.*?),off1:(.*?),'
+                                        r'off2:(.*?),type:(.*?),sent:(.*?),'
+                                        r'court:(.*?),counts:(\d?)', re.DOTALL)
+
+                matches = re.match(regex, charge.name)
+                if matches:
```
zacharylawrence: Nit: You could optionally consider inverting this so you can drop both the `if` and the `else`.

Example:
```
if not matches:
  logging.error(...)
  continue

matches = re.match(regex, charge.name)
```

In my opinion, this kind of guard clause is easier to read since it reduces nesting, but feel free to ignore this comment if you disagree.

jamwalla:
#745

recidiviz/ingest/us_tn_mcminn/us_tn_mcminn_scraper.py
ohinds: I think (c) is good. Can you file an issue and assign it to me?

recidiviz/ingest/vendors/eagle_advantage/eagle_advantage_scraper.py
ohinds: Whooooaaaaaa. Mind blown.

recidiviz/tests/ingest/us_tn_mcminn/us_tn_mcminn_scraper_test.py
jamwalla: yeah! Terin did this in #743, I made a note of this approach in #675.

recidiviz/ingest/us_tn_mcminn/us_tn_mcminn_scraper.py
jamwalla: Created #754

PR #709, opened 2019-01-22
Add TX aggregate scraper

zacharylawrence:
#129


recidiviz/tests/ingest/aggregate/regions/ga/__init__.py
```diff
@@ -0,0 +1,16 @@
+# Recidiviz - a platform for tracking granular recidivism metrics in real time
```
jessex: I think the package for all of these tests is incorrect. It's `ga` instead of `tx`.

zacharylawrence: Ah good eye! Updated.

PR #710, opened 2019-01-23
DCN vendor scraper and one child

arian487:
An open question is the enum overrides.  There were some weird bond types that I wasn't exactly sure what to do with, thoughts?


recidiviz/ingest/vendors/dcn/dcn_scraper.py
```diff
@@ -0,0 +1,176 @@
...
+            data[compression_key] = scraper_utils.decompress_string(
+                data[compression_key])
+
+    def get_enum_overrides(self):
+        return {
+            # This is set on bond type sometimes.
+            'Sentenced': BondType.NO_BOND,
+            'Declined': BondType.CASH,
+            'CASH OR SURETY BOND': BondType.CASH,
+            'Purge': None,
```
jessex: [According to this source](https://www.avvo.com/legal-answers/my-daughter-is-in-jail-under-a-purge-bond--how-can-2241151.html) and others, a "Purge bond" is one where the entire bond must be paid in full before release, I think. A specific example I encountered was someone held in jail for failing to pay child support, with a purge bond set to the full amount still outstanding. So I think this would map to `CASH`. @ohinds?

recidiviz/ingest/vendors/dcn/dcn_scraper.py
```diff
@@ -0,0 +1,176 @@
...
+        compression_key = '__VIEWSTATE'
+        if data and compression_key in data:
+            data[compression_key] = scraper_utils.decompress_string(
+                data[compression_key])
+
+    def get_enum_overrides(self):
+        return {
+            # This is set on bond type sometimes.
+            'Sentenced': BondType.NO_BOND,
+            'Declined': BondType.CASH,
```
jessex: Does this maybe map to `BondStatus.DENIED` instead?

recidiviz/ingest/vendors/dcn/dcn_scraper.py
```diff
@@ -0,0 +1,176 @@
...
+        """
+        compression_key = '__VIEWSTATE'
+        if data and compression_key in data:
+            data[compression_key] = scraper_utils.decompress_string(
+                data[compression_key])
+
+    def get_enum_overrides(self):
+        return {
+            # This is set on bond type sometimes.
+            'Sentenced': BondType.NO_BOND,
```
jessex: This may be divorced enough from our context to not be relevant, i.e. these people may be talking about sentencing versus bail in general and maybe this is something more specific, but [this thread at least](https://www.avvo.com/legal-answers/what-does-the-bail-type--sentenced--mean-when-look-2182246.html) has people saying that this simply means the person has already been sentenced and is therefore no longer eligible for bail. If you look at some of the bookings where you saw this, does that seem consistent?

recidiviz/ingest/vendors/dcn/dcn_scraper.py
```diff
@@ -0,0 +1,176 @@
...
+        super(DcnScraper, self).__init__(region_name)
+
+        self.yaml = yaml_file or os.path.join(
+            os.path.dirname(__file__), 'dcn_scraper.yaml')
+        self._required_session_vars = ['__VIEWSTATE']
+
+        self._keys_re = r"keys':\[(.*)\]},"
+        self._people_re = r"DCN\/inmate-details(.*?)\\\'>"
+        self._callback_params = "c0:KV|691;[{}];GB|22;12|PAGERONCLICK5|PSP-1;"
+        self._base_endpoint = ('https://jail.desotosheriff.org/DCN/'
```
jamwalla: this should probably go in the child scraper instead

recidiviz/ingest/vendors/dcn/dcn_scraper.py
arian487: Spoke with oliver offline, we agreed they are just overloading charge status in this case, i've changed the PR to reflect that.

recidiviz/ingest/vendors/dcn/dcn_scraper.py
arian487: The problem was that no bond for us has always meant that the amount field was empty or 0.  In this case, the site had declined under type, but the amount field was not none so I was confused.  Spoke with oliver we agreed to map it to no bond, but not wipe out the amount field.

recidiviz/ingest/vendors/dcn/dcn_scraper.py
arian487: You are correct, mapping it to cash.

PR #714, opened 2019-01-23
Populate data should return whether to persist

colincadams:


recidiviz/ingest/base_scraper.py
```diff
@@ -170,50 +170,50 @@ def _generic_scrape(self, request: QueueRequest):
...
+                    ingest_info=ingest_info_to_send,
                 ))
-        # If we don't have any more tasks to scrape, we are at a leaf node and
-        # we are ready to populate the database.
-        else:
-            # Something is wrong if we get here but don't have ingest_info
-            # to work with.
-            if not ingest_info:
-                raise ValueError(
-                    'IngestInfo must be populated if there are no more tasks')
```
jamwalla: Should we keep this error if `scraped_data.persist`?

colincadams: sure sure!

PR #717, opened 2019-01-23
change run scraper to use queue

colincadams:
The old implementation started the next task immediately upon it being added instead of managing a queue. This caused issues for `SCRAPE_DATA_AND_MORE` tasks as the next task would start running before the current one was able to persist its data, so none of the data was printed unless the entire scrape was allowed to finish.


recidiviz/ingest/run_scraper.py
```diff
@@ -19,7 +19,7 @@
 
 usage: run_scraper.py [-h] --region REGION [--num_tasks NUM_TASKS]
                       [--sleep_between_requests SLEEP_BETWEEN_REQUESTS]
-                      [--run_forever RUN_FOREVER] [--fail_fast FAIL_FAST]
+                      [--run_forever RUN_FOREVER] [--no_fail_fast]
```
jamwalla: 😍

recidiviz/ingest/run_scraper.py
```diff
@@ -33,106 +33,103 @@
...
-        '--fail_fast', required=False, default=str(fail_fast),
-        help='Stop running after an error, default is {}'.format(fail_fast)
+        '--no_fail_fast', required=False, dest='fail_fast',
+        action='store_false', help='Continue running after an error'
     )
     parser.add_argument(
-        '--log', required=False, default=logging_level,
-        help='Set the logging level, default is '
-             '{}'.format(logging.getLevelName(logging_level))
+        '--log', required=False, default=logging.INFO,
```
jamwalla: I think the help message will tell you the default is `30`, which is why I used `logging.getLevelName` before. Is there an easy way to keep that?

recidiviz/ingest/run_scraper.py
```diff
@@ -33,106 +33,103 @@
...
 from recidiviz.ingest import constants
 from recidiviz.ingest.task_params import QueueRequest
 from recidiviz.utils import regions
 
-# Sleep 1 seconds per task
-sleep_between_requests = 1
 
-# Number of people to scrape by default
-num_tasks_left = 5
+# This function acts as a bound method to the scraper instance.  It is
```
jessex: Drop the "is"

recidiviz/ingest/run_scraper.py
```diff
@@ -33,106 +33,103 @@
...
+# overwrites the functionality of add_task to add to the inmemory queue.
```
jessex: "in-memory"

recidiviz/ingest/run_scraper.py
```diff
@@ -33,106 +33,103 @@
...
+def add_task(queue, _self, task_name, request):
+    """Overwritten version of add task which simply runs the task
```
jessex: Is this docstring correct? It looks like it is not ignoring the task queue to me. It's appending to it on line 57 below.

recidiviz/ingest/run_scraper.py
colincadams: Done. I removed this second sentence as it mimicked the docstring.

recidiviz/ingest/run_scraper.py
colincadams: Done, removed but reflected below.

recidiviz/ingest/run_scraper.py
colincadams: Fixed this.

recidiviz/ingest/run_scraper.py
colincadams: Oh good call! What about doing it this way? Then you can always just supply a string on the command line.

jamwalla: thats coool

PR #711, opened 2019-01-23
Remove "implement process_record" from jailtracker_scraper docstring

KareemG:
Small documentation update.

PR #718, opened 2019-01-23
only set up client-secret-staging key before deploy

jamwalla:
possible fix for travis errors in #691 and #711


jamwalla:
@arian487 or @jessex, can you take another look? I think the most straightforward thing is to not do this on pull requests

jamwalla:
great, thanks!

PR #720, opened 2019-01-23
 Add ny aggregate scraper

zacharylawrence:
#129


recidiviz/ingest/aggregate/regions/ny/ny_aggregate_ingest.py
```diff
@@ -0,0 +1,159 @@
...
+
+    # Recombine top_df and bottom_df since it's not the correct table division
+    aligned_df = pd.concat([top_df, bottom_df], ignore_index=True)
+
+    # The second table starts when a new facility is listed
+    table_starts = np.where(aligned_df['FACILITY'].notnull())[0]
+
+    # If only 1 'FACILITY' is found, then it's the last page with only 1 table
+    if len(table_starts) == 1:
+        return [aligned_df]
```
jessex: Should this actually be `(aligned_def)` since we're returning a tuple of `first_df, second_df` below? We can process both the same generally but we're technically returning different types right now.

recidiviz/ingest/aggregate/regions/ny/ny_aggregate_ingest.py
```diff
@@ -0,0 +1,159 @@
...
+    for column_name in set(result.columns) - {'report_date'}:
+        result[column_name] = result[column_name].apply(locale.atoi)
+
+    result['facility_name'] = df['FACILITY'].iloc[0]
+
+    return result
+
+
+def _transpose_df(df: pd.DataFrame) -> pd.DataFrame:
+    """Transpose the Dataframe by setting the ."""
```
jessex: "by setting the ..."

recidiviz/ingest/aggregate/regions/ny/ny_aggregate_ingest.py
```diff
@@ -0,0 +1,159 @@
...
+    second_df = aligned_df[second_start:]
+
+    return first_df, second_df
+
+
+def _format_df(df: pd.DataFrame) -> pd.DataFrame:
+    """Format the DataFrame to match the schema."""
+    result = _transpose_df(df)
+
+    result = aggregate_ingest_utils.rename_columns_and_select(result, {
```
jessex: I see we're not capturing the `census` rows. What exactly do they mean in the PDFs?

recidiviz/ingest/aggregate/regions/ny/ny_aggregate_ingest.py
zacharylawrence: Yeah, good point. Updated this to return a `Tuple[pd.DataFrame]`.

recidiviz/ingest/aggregate/regions/ny/ny_aggregate_ingest.py
zacharylawrence: The `Census` row seems to be `In House` - `Boarded In` + `Boarded Out` meaning something like: `People assigned to this facility` - `People held in other facilities` + `People this facility is holding for other facilities`. Talking with @ohinds it seems like we want to keep this field since it's an ADP we care about. I went ahead and added it here. Thanks for pointing this out!!

recidiviz/ingest/aggregate/regions/ny/ny_aggregate_ingest.py
zacharylawrence: Updated!

PR #722, opened 2019-01-24
Add additional database_test

zacharylawrence:


zacharylawrence:
This is added ing #728 instead

PR #725, opened 2019-01-24
make endpoint required in Task for scraper

terinpw:


recidiviz/ingest/us_ny/us_ny_scraper.py
```diff
@@ -264,7 +264,7 @@ def _get_store_person_task(self, content) -> Task:
             A dict containing the params necessary to store the person info.
         """
         return Task(
-            endpoint=None,
+            endpoint='',
```
zacharylawrence: Consider adding a quick comment why we need to set this to empty string?

PR #729, opened 2019-01-24
Us Nc Buncombe

ohinds:


recidiviz/ingest/vendors/superion/superion_scraper.py
```diff
@@ -184,6 +180,21 @@ def populate_data(self, content, task: Task,
                             bond.bond_type = bond.amount
                             bond.amount = None
 
+                # Bond type is listed as 'PAID' when status is posted.
+                if bond.bond_type == 'PAID':
+                    bond.status = 'POSTED'
```
jamwalla: `BondStatus.POSTED.value`, just in case?

region_manifest.yaml
```diff
@@ -148,6 +148,14 @@ regions:
     names_file: us_ny_names.csv
     scraper_package: us_fl_hendry
     timezone: America/New_York
+  us_nc_buncombe:
+    agency_name: $agency
```
jamwalla: we should fix this in create_scraper if it's usually not necessary (I keep just putting "x county Sheriff's Office" but that seems silly)

recidiviz/common/constants/person.py
```diff
@@ -75,6 +75,7 @@ def _get_default_map():
     'INDIAN': Race.AMERICAN_INDIAN_ALASKAN_NATIVE,
     'ASIAN': Race.ASIAN,
     'ASIAN / PACIFIC ISLANDER': Race.ASIAN,
+    'ASIAN / PACIFIC ISLANDER/': Race.ASIAN,
```
jessex: I wonder if instead of this, and taken to its extreme adding all kinds of off-by-one variants to these maps, we should just strip punctuation from the beginnings and ends of the strings we run through the converter?

recidiviz/ingest/us_nc_buncombe/us_nc_buncombe_scraper.py
```diff
@@ -0,0 +1,36 @@
...
+    """Scraper implementation for us_nc_buncombe."""
+    def __init__(self):
+        super(UsNcBuncombeScraper, self).__init__('us_nc_buncombe')
+
+    def get_enum_overrides(self):
+        return {
+            'DOM-VIO': BondType.NO_BOND,
+            'INCLUDED W OTHER': BondType.EXTERNAL_UNKNOWN,
+            'SECURED FTA PRIOR': BondType.NO_BOND,
+            'PURGE PAYMENT': BondType.CASH,
```
jessex: Now that PURGE has shown up in multiple counties (and will presumably show up in more), should we promote it to the global map and out of overrides?

recidiviz/ingest/us_nc_buncombe/us_nc_buncombe.yaml
```diff
@@ -0,0 +1,2 @@
+key_mappings:
```
jessex: Do we need this empty mappings file? us_nc_guilford skips it entirely.

arian487: Yup, this is not needed.

recidiviz/ingest/us_nc_buncombe/us_nc_buncombe_scraper.py
arian487: I agree, I imagine if we ever see purge anywhere else, it'll always map to cash?

recidiviz/ingest/vendors/superion/superion_scraper.py
ohinds: Done

region_manifest.yaml
ohinds: I actually kinda like it. And I feel like it could be useful.

recidiviz/common/constants/person.py
ohinds: Totally. #733

ohinds:
UsNcBumcome as a child of Superion

PR #731, opened 2019-01-24
Add period fields to master entities

zdg2102:


zdg2102:
Discussed offline, taking a different approach and dropping this one

PR #734, opened 2019-01-25
Switch to docker container and app engine flex

colincadams:
This is necessary to be able to run tabula (backed by java) in our environment.


app.yaml
```diff
@@ -1,7 +1,10 @@
-runtime: python37
-instance_class: F4
-
-entrypoint: gunicorn -c gunicorn.conf.py -b :$PORT recidiviz.server:app
-
+runtime: custom
+env: flex
+resources:
```
jessex: These resource settings are per-node, right?

recidiviz/__init__.py
```diff
@@ -37,9 +37,9 @@
     cloudsql_instance_id = secrets.get_secret('cloudsql_instance_id')
 
     sqlalchemy_url = \
-        ('postgresql://{db_user}:{db_password}@{db_host}/{'
-         'db_name}?host=/cloudsql/{cloudsql_instance_id}') \
-            .format(db_user=db_user, db_password=db_password, db_host=db_host,
```
jessex: Will db_host be used elsewhere or should it be removed from `env` entirely?

colincadams: I think we can get rid of it

app.yaml
colincadams: Correct this is per instance. It determines the machine type that is used for each instance.
https://cloud.google.com/appengine/docs/flexible/python/reference/app-yaml#resource-settings

Scaling is done automatically with flex, although we can modify the way it scales or override it altogether. We are planning to watch how it does at scaling automatically and override if necessary.

arian487: Thats correct, instances are auto managed this is just setting the VM size.  The nice thing about Flex, is that it lets us scale the memory and cpu independently, as opposed to setting an instance size which scales both.  In our case we are memory constrained, but not CPU constrained so we were able to choose the smallest CPU size, but up our RAM as needed.

PR #735, opened 2019-01-25
Cloud function

arian487:
Add cloud function to send an HTTP request to the cloud_function endpoint.  This PR adds a util as well for making auth'd requests to IAP-protected webapps.  The permissions and cloud functions have already been set up on prod and staging.


recidiviz/cloud_functions/cloud_functions.py
```diff
@@ -0,0 +1,72 @@
...
+from recidiviz.utils.params import get_value
+
+cloud_functions_blueprint = Blueprint('cloud_functions', __name__)
+
+
+class StateAggregateError(Exception):
+    """Errors thrown in the state aggregate endpoint"""
+
+
+@cloud_functions_blueprint.route('/state_aggregate')
```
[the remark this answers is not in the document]
zacharylawrence: Good point, I think it's good practice to prefix our endpoints with `v1/` and increment that every time we make a breaking version update: https://apiguide.readthedocs.io/en/latest/build_and_publish/use_RESTful_urls.html#versioning

recidiviz/cloud_functions/cloud_functions.py
```diff
@@ -0,0 +1,72 @@
...
+
+    # Providing a stream buffer to tabula reader does not work because it
+    # tries to load the file into the local filesystem, since appengine is a
+    # read only filesystem (except for the tmpdir) we download the file into
+    # the local tmpdir and pass that in.
+    tmpdir_path = os.path.join(tempfile.gettempdir(), filename)
+    fs.get(path, tmpdir_path)
+    logging.info('Successfully downloaded file from gcs')
+
+    parser(tmpdir_path, datetime.datetime.now())
```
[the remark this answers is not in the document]
jessex: I was just about to suggest allowing us to pass the date in as a query parameter for the same reasons.

recidiviz/cloud_functions/cloud_functions.py
```diff
@@ -0,0 +1,72 @@
...
+    parser = state_to_parser[state]
+    fs = gcsfs.GCSFileSystem(project=os.environ.get('GCP_PROJECT'))
+
+    # Providing a stream buffer to tabula reader does not work because it
+    # tries to load the file into the local filesystem, since appengine is a
+    # read only filesystem (except for the tmpdir) we download the file into
+    # the local tmpdir and pass that in.
+    tmpdir_path = os.path.join(tempfile.gettempdir(), filename)
+    fs.get(path, tmpdir_path)
+    logging.info('Successfully downloaded file from gcs')
```
jessex: Let's place the filename in the log message, too.

recidiviz/server.py
```diff
@@ -59,6 +60,7 @@
 app.register_blueprint(worker, url_prefix='/scraper')
 app.register_blueprint(actions, url_prefix='/ingest')
 app.register_blueprint(infer_release_blueprint, url_prefix='/infer_release')
+app.register_blueprint(cloud_functions_blueprint, url_prefix='/cloud_function')
```
jessex: I commented about this elsewhere, but I think the URL paths for the API, such as it is, are pretty inconsistent at the moment. Instead of naming them in the current manner, I'd like us to take the opportunity to provide consistent paths that are relevant to the domain. For example, something like the following:
```
/ingest/v1/aggregates
/ingest/v1/records
/ingest/v1/scraper/{action}
/ingest/v1/infer_release
```

recidiviz/cloud_functions/main.py
```diff
@@ -0,0 +1,42 @@
...
+"""This file contains all of the relevant cloud functions"""
+
+import logging
+import os
+
+from cloud_function_utils import make_iap_request
+
+CLOUD_FUNCTION_URL = ('http://{}.appspot.com/cloud_function/state_aggregate?'
+                      'bucket={}&state={}&filename={}')
+CLIENT_ID = ('984160736970-flbivauv2l7sccjsppe34p7436l6890m.apps.'
```
jessex: It seems like this should be a secret, no? If so, do we need to rotate it now that it's been in version control?

recidiviz/cloud_functions/cloud_functions.py
```diff
@@ -0,0 +1,72 @@
...
+from recidiviz.ingest.aggregate.regions.fl import fl_aggregate_ingest
+from recidiviz.ingest.aggregate.regions.ga import ga_aggregate_ingest
+from recidiviz.ingest.aggregate.regions.tx import tx_aggregate_ingest
+from recidiviz.utils.auth import authenticate_request
+from recidiviz.utils.params import get_value
+
+cloud_functions_blueprint = Blueprint('cloud_functions', __name__)
+
+
+class StateAggregateError(Exception):
```
jessex: Should this go in the new `recidiviz.ingest.errors`?

recidiviz/cloud_functions/cloud_function_utils.py
```diff
@@ -0,0 +1,117 @@
...
+    Returns:
+      The page body, or raises an exception if the page couldn't be retrieved.
+    """
+    # Figure out what environment we're running in and get some preliminary
+    # information about the service account.
+    bootstrap_credentials, _ = google.auth.default(
+        scopes=[IAM_SCOPE])
+    if isinstance(bootstrap_credentials,
+                  google.oauth2.credentials.Credentials):
+        raise Exception('make_iap_request is only supported for service '
```
jessex: Under what conditions will this happen? If a user tries to make the API request directly through the browser? If so, then is that something we ever want to enable for manual uploads?

recidiviz/cloud_functions/cloud_function_utils.py
```diff
@@ -0,0 +1,117 @@
...
+                        'accounts.')
+
+    # For service account's using the Compute Engine metadata service,
```
jessex: accounts

recidiviz/cloud_functions/cloud_function_utils.py
```diff
@@ -0,0 +1,117 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+# =============================================================================
+
+"""This file contains all of the relevant helpers for cloud functions."""
```
jessex: Was this completely artisan and hand-made, or copied in part from Google Cloud docs somewhere? Regardless, are there any docs on the authorization requirements in play here that we can point to from this docstring?

recidiviz/cloud_functions/cloud_function_utils.py
```diff
@@ -0,0 +1,117 @@
...
+    # service_account_credentials gives us a JWT signed by the service
+    # account. Next, we use that to obtain an OpenID Connect token,
+    # which is a JWT signed by Google.
+    google_open_id_connect_token = get_google_open_id_connect_token(
+        service_account_credentials)
+
+    # Fetch the Identity-Aware Proxy-protected URL, including an
+    # Authorization header containing "Bearer " followed by a
+    # Google-issued OpenID Connect token for the service account.
+    resp = requests.request(
```
jessex: Can we just call this `response`? The shortening feels unnecessary.

recidiviz/cloud_functions/cloud_function_utils.py
```diff
@@ -0,0 +1,117 @@
...
+        method, url,
+        headers={'Authorization': 'Bearer {}'.format(
+            google_open_id_connect_token)}, **kwargs)
+    if resp.status_code == 403:
+        raise Exception('Service account {} does not have permission to '
+                        'access the IAP-protected application.'.format(
+                            signer_email))
+    elif resp.status_code != 200:
```
jessex: There's nothing else in the 2XX- or 3XX-space that's allowable? I'm guessing no since we are receiving a response payload, but want to be sure this isn't unnecessarily restrictive.

recidiviz/cloud_functions/cloud_functions.py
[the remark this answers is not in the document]
jessex: Agreed! I provided an example proposal in a comment below. That includes the version component as the first path segment _after_ a general `ingest` prefix, but I could see it going either way. At the very beginning may make more sense.

recidiviz/cloud_functions/cloud_function_utils.py
```diff
@@ -0,0 +1,115 @@
...
+
+import google.auth
+import google.auth.app_engine
+import google.auth.compute_engine.credentials
+import google.auth.iam
+from google.auth.transport.requests import Request
+import google.oauth2.credentials
+import google.oauth2.service_account
+
+IAM_SCOPE = 'https://www.googleapis.com/auth/iam'
```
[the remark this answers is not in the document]
arian487: Done.

recidiviz/cloud_functions/cloud_function_utils.py
```diff
@@ -0,0 +1,115 @@
...
+    service_account_jwt = (
+        service_account_credentials._make_authorization_grant_assertion())
+    request = google.auth.transport.requests.Request()
+    body = {
+        'assertion': service_account_jwt,
+        'grant_type': google.oauth2._client._JWT_GRANT_TYPE,
+    }
+    token_response = google.oauth2._client._token_endpoint_request(
+        request, OAUTH_TOKEN_URI, body)
+    return token_response['id_token']
```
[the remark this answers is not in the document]
arian487: Done.

recidiviz/cloud_functions/cloud_function_utils.py
```diff
@@ -0,0 +1,115 @@
...
+                            signer_email))
+    elif resp.status_code != 200:
+        raise Exception(
+            'Bad response from application: {!r} / {!r} / {!r}'.format(
+                resp.status_code, resp.headers, resp.text))
+    else:
+        return resp
+
+
+def get_google_open_id_connect_token(service_account_credentials):
```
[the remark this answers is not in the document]
arian487: I think it can be used in isolation, so I made it not private.

recidiviz/cloud_functions/cloud_functions.py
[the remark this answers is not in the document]
arian487: IMO that type of logic should live entirely in the parser.  It represents the 'report_date' and these cloud functions should not know or care about what report date is.  I think in the next iteration, parser should take only the file and then it can extract the date in any way it wants.

requirements.txt
```diff
@@ -30,6 +30,7 @@ enum34==1.1.6
 eventlet==0.24.1 ; extra == 'eventlet'
 flask==1.0.2
 funcsigs==1.0.2
+gcsfs==0.2.0
```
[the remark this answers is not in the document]
arian487: Nope, removed.

recidiviz/cloud_functions/cloud_functions.py
```diff
@@ -0,0 +1,72 @@
...
+    """Errors thrown in the state aggregate endpoint"""
+
+
+@cloud_functions_blueprint.route('/state_aggregate')
+@authenticate_request
+def state_aggregate():
+
+    state_to_parser = {
```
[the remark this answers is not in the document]
arian487: Yea, I feel comfortable with it living in code since it references a specific method and there will only be 7 of them.

recidiviz/cloud_functions/cloud_functions.py
arian487: Done.

recidiviz/server.py
arian487: I completely agree, but IMO lets do that in a separate PR all in one go, I've made an issue:
https://github.com/Recidiviz/recidiviz-data/issues/739

recidiviz/cloud_functions/cloud_functions.py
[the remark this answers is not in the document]
arian487: I've opened an issue, I don't think that refactor belongs in this PR:
https://github.com/Recidiviz/recidiviz-data/issues/739

recidiviz/cloud_functions/cloud_function_utils.py
arian487: Done.

recidiviz/cloud_functions/cloud_functions.py
jessex: Ah, I see. That makes sense. But Cloud Functions handles linking automatically, e.g. we have tons of imports to pip packages but also to our own code elsewhere in the repository, and Cloud Functions bundles everything without having to be statically told where it all lives?

recidiviz/cloud_functions/cloud_function_utils.py
arian487: Its mostly copied from docs, I removed some parts that were not relevant for us.  I've added the doc in the docstring.

recidiviz/cloud_functions/cloud_function_utils.py
arian487: Done.

recidiviz/cloud_functions/cloud_function_utils.py
arian487: Nope, I think this is fine.

recidiviz/cloud_functions/cloud_functions.py
arian487: Cloud functions bundles only starting from the current directory.  The idea is to have a requirements in a sub-directory like we have here, that tells cloud functions how to package and deploy this small unit of work.  In this case, the unit of work does very little, and so only relies on requests.  It doesn't see the imports that happen in its parents because the parents directories are not deployed at all.

requirements.txt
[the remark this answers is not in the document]
arian487: We actually did need it, added back in.

jessex:
What's the latest state here, i.e. with testing and deployment? Is this ready for final review?

arian487:
Yes, this is ready for final review.  I'm writing another PR now that includes some tests for this but IMO we should get this in because it is actually whats deployed in prod (I wanted to test the E2E of the cloud function).

recidiviz/cloud_functions/main.py
```diff
@@ -0,0 +1,47 @@
...
+                       'bucket={}&state={}&filename={}')
+_CLIENT_ID = {
+    'recidiviz-staging': ('984160736970-flbivauv2l7sccjsppe34p7436l6890m.apps.'
+                          'googleusercontent.com'),
+    'recidiviz-123': ('688733534196-uol4tvqcb345md66joje9gfgm26ufqj6.apps.'
+                      'googleusercontent.com')
+}
+
+
+def parse_state_aggregate(data, _):
```
jessex: Could we get a docstring on this explaining the parameters and the nature of when/how it's called? Also, as @zacharylawrence noted elsewhere, type annotations on the function signatures throughout.

recidiviz/cloud_functions/cloud_function_utils.py
```diff
@@ -0,0 +1,122 @@
...
+              ('GET', 'OPTIONS', 'HEAD', 'POST', 'PUT', 'PATCH', 'DELETE')
+      **kwargs: Any of the parameters defined for the request function:
+                https://github.com/requests/requests/blob/master/requests/api.py
+
+    Returns:
+      The page body, or raises an exception if the page couldn't be retrieved.
+    """
+    # Figure out what environment we're running in and get some preliminary
+    # information about the service account.
+    bootstrap_credentials, _ = google.auth.default(
```
zacharylawrence: Could this be one line?

recidiviz/cloud_functions/cloud_function_utils.py
```diff
@@ -0,0 +1,122 @@
...
+import google.auth.compute_engine.credentials
+import google.auth.iam
+from google.auth.transport.requests import Request
+import google.oauth2.credentials
+import google.oauth2.service_account
+
+_IAM_SCOPE = 'https://www.googleapis.com/auth/iam'
+_OAUTH_TOKEN_URI = 'https://www.googleapis.com/oauth2/v4/token'
+
+# pylint: disable=protected-access
```
zacharylawrence: Is there a reason we need this pylint override?

recidiviz/cloud_functions/main.py
```diff
@@ -0,0 +1,54 @@
...
+    """This function is triggered when a file is dropped into the state
+    aggregate bucket and makes a request to parse and write the data to the
+    aggregate table database.
+
+    data: A cloud storage object that holds name information and other metadata
```
jessex: Can we document what the `_` is? Between the underscore'd/ignored parameter and the fact that the function is triggered automatically from outside the artifact, I think it's worth saying what that is.

zacharylawrence:
> Only one last doc request from me. Amazing work, @arian487!

Agreed, nice job finding out how all the auth stuff works!

recidiviz/cloud_functions/main.py
arian487: Done.

recidiviz/cloud_functions/cloud_function_utils.py
arian487: Done

recidiviz/cloud_functions/cloud_function_utils.py
arian487: Yea the code I copied to do this all has this private members.  I looked at other sources it looks like everyone does the same thing, why they dont just make them public members I have no idea haha.

recidiviz/cloud_functions/main.py
arian487: Done.

PR #736, opened 2019-01-25
update entity matching to allow some people to have external ids and some to not

### LINKED ISSUE #716: entity match if not all entities have external_id
@terinpw found a site with historical booking info where only the current booking has a booking ID. We should match properly in this case, but right now we assume that all entities either do or don't have ids

terinpw:
closes #716


recidiviz/persistence/entity_matching.py
```diff
@@ -48,13 +48,38 @@ def match_entities(
...
-    if _has_external_ids(ingested_people):
-        db_people = database.read_people_by_external_ids(session, region,
-                                                         ingested_people)
-    else:
-        db_people = database.read_people_with_open_bookings(session, region,
-                                                            ingested_people)
+    with_external_ids = []
+    without_external_ids = []
 
+    for ingested_person in ingested_people:
```
jamwalla: optionally:

`without_external_ids, with_external_ids =`[`more_itertools.partition`](https://more-itertools.readthedocs.io/en/latest/api.html#more_itertools.partition)`(ingested_people, _has_external_id)` where `_has_external_id = lambda p: bool(p.external_id)`.

Either way, also delete `_has_external_ids` if we're not using it anywhere else.

terinpw: Chatted offline, b/c partition returns a filter, doesn't make the code much more readable.

PR #737, opened 2019-01-25
Travis should use pipenv

colincadams:
This allows us to get rid of requirements.txt, as docker doesn't use it either


colincadams:
@arian487 do the cloud functions use our top-level `requirements.txt`?

arian487:
Nope, the cloud functions use their own requirements.txt in the directory

colincadams:
@arian487 after our discussion yesterday we are good to go on this correct? can I get an LGTM? 😁

Issue #719, opened 2019-01-23
schema additions


jamwalla:
From #707:
 - Add column Charge.charge_notes
 - Add enum value CourtType.CIVIL
 - Add enum value ChargeStatus.REVOKED

From #723:
 - Add column Bond.BOND_AGENT (BOND_AGENCY? BONDSMAN?)

ohinds:
Ah right, you weren't around when jasmine got back to me about surety. Apparently this means different things in different places, so we're going to call that 'secured'. I do like the idea of adding a bond agent field though.

zdg2102:
`charge_notes` and `bond_agent` were added in #747

`CourtType.CIVIL` was added in #753

`ChargeStatus.REVOKED` is part of community supervision, so it'll be covered by #400

And the changes to `BondType` and `BondStatus` will be covered by #741

PR #749, opened 2019-01-25
SmartCOP (CTS) vendor scraper

jamwalla:
3 children: us_fl_bradford, us_fl_columbia, us_ga_gwinnett

~~1 new enum: BondType.SURETY (TBD, #719)~~ mapping 'SURETY' -> BondType.SECURED instead


recidiviz/ingest/us_fl_bradford/us_fl_bradford_scraper.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def __init__(self):
+        super(UsFlBradfordScraper, self).__init__('us_fl_bradford')
+
+    def get_initial_task(self) -> Task:
+        return self.get_preview_task()
+
+    def get_enum_overrides(self):
+        return {
+            'C': ChargeDegree.EXTERNAL_UNKNOWN,
```
jessex: Curious: are these meant to stay as unknown indefinitely, or are we trying to figure out what they correspond to?

recidiviz/ingest/us_fl_columbia/us_fl_columbia_scraper.py
```diff
@@ -0,0 +1,43 @@
...
+class UsFlColumbiaScraper(SmartCopScraper):
+    """Scraper implementation for us_fl_columbia."""
+    def __init__(self):
+        super(UsFlColumbiaScraper, self).__init__('us_fl_columbia')
+
+    def get_initial_task(self):
+        return self.get_preview_task()
+
+
+    def get_enum_overrides(self):
```
jessex: If these maps are the same across all counties for the vendor, shouldn't the map go in the vendor?

recidiviz/ingest/us_fl_columbia/us_fl_columbia_scraper.py
```diff
@@ -0,0 +1,43 @@
...
+        return {
+            'C': ChargeDegree.EXTERNAL_UNKNOWN,
+            'F': ChargeDegree.EXTERNAL_UNKNOWN,
+            'S': ChargeDegree.EXTERNAL_UNKNOWN,
+            'T': ChargeDegree.EXTERNAL_UNKNOWN,
+            'N': ChargeDegree.EXTERNAL_UNKNOWN,
+            'I': Race.AMERICAN_INDIAN_ALASKAN_NATIVE,
+            'U': Race.EXTERNAL_UNKNOWN,
+            '': None
```
jessex: Does an empty string not usually map to None?

jamwalla:
i think this is ready for review. @ohinds, do you know what charge classes ('Degree' or 'Level') the following are? C, F, L, M, N, S, T

thanks!!

recidiviz/ingest/us_fl_columbia/us_fl_columbia_scraper.py
jamwalla: fixed in #707, there should never be empty strings on IngestInfo.

recidiviz/ingest/us_fl_bradford/us_fl_bradford_scraper.py
jamwalla: We should figure out what they mean! I moved them to the vendor scraper since they seem to be common across TN and GA.

queue.yaml
```diff
@@ -234,6 +258,18 @@ queue:
     task_age_limit: 3d
     task_retry_limit: 5
 
+# us-fl-columbia-scraper - Columbia County, Florida
```
jessex: We've got Columbia County in this changeset twice now. Lines 225 and 261.

jamwalla: oops, thanks!!

recidiviz/ingest/vendors/smart_cop/smart_cop_scraper.py
```diff
@@ -0,0 +1,140 @@
...
+        task.json['RecordsLoaded'] = str(int(last_index) + returned)
+        return [task]
+
+    def get_initial_task(self) -> Task:
+        endpoint = '{}/{}'.format(self.region.base_url, 'AddMoreResults')
+        return Task(endpoint=endpoint, json=_SEARCH_PARAMS,
+                    response_type=ResponseType.JSON,
+                    task_type=TaskType.SCRAPE_DATA_AND_MORE)
+
+    def get_preview_task(self) -> Task:
```
ohinds: Do we know what the limit is? What if a jail is under this limit now and we use the preview task, but then it grows and we can't use it anymore?

jamwalla: This is a really good point. I changed this so all scrapers first try the preview task, since it fails gracefully. If it does fail, we can fall back on the search instead.

jamwalla:
> C, F, L, M, N, S, T

C, F, N are both
L, S, T are only a degree
M is only a level

jamwalla:
Talked to Oliver, I'm going to merge this now and we'll fill in the overrides when we find out what things mean (#806).

PR #750, opened 2019-01-26
split app.yaml for staging and prod

colincadams:


prod.yaml
```diff
@@ -0,0 +1,11 @@
...
+env: flex
+
+resources:
+  cpu: 1
+  memory_gb: 0.5
+  disk_size_gb: 10
+beta_settings:
+  cloud_sql_instances: recidiviz-123:us-east4:prod-data
+env_variables:
+  RECIDIVIZ_ENV: production
```
jessex: I see this is the same value in both environments. Is that correct?

colincadams: For now yes because we want [`in_prod`](https://github.com/Recidiviz/recidiviz-data/blob/master/recidiviz/utils/environment.py#L35) to be true. But we could change it in the future.

Issue #733, opened 2019-01-25
Consider stripping punctuation and special characters from strings before doing enum mapping.


ohinds:
To prevent things like:
    'ASIAN / PACIFIC ISLANDER': Race.ASIAN,
    'ASIAN / PACIFIC ISLANDER/': Race.ASIAN,

KareemG:
I've run into a similar case, where the normalize function (seen in #756) isn't enough . For example:

'NO BOND - MUST SEE PRETRIAL': BondType.NO_BOND,
'NO BOND - SENTENCED': BondType.NO_BOND,

A solution I was thinking about was refactoring from a 1-1 enum_mapping to enum_predicates (or maybe support both?). So maybe something like:
```
_DATA_MASSAGER = {
    ChargeStatus.PENDING: [
        lambda x : x == 'OPEN',
    ],
    BondType.CASH: [
        lambda x : x == 'FULL CASH',
    ],
    BondType.NO_BOND: [
        lambda x : x.startswith('NO BOND'),
    ]
}
```

I was thinking this might also simplify the manual charge info manipulations seen in us_in_vigo, but perhaps it needs more use cases.

PR #756, opened 2019-01-28
common_utils.normalize

jamwalla:
move `normalize` from converter_utils to common_utils, so that it can be called by enum `from_str`s. When we figure out #733, this is a centralized place to do that.


recidiviz/common/common_utils.py
```diff
@@ -21,9 +21,19 @@
...
 def is_generated_id(id_str: Optional[str]) -> bool:
     return id_str is not None and id_str.endswith(_GENERATED_ID_SUFFIX)
+
+
+def normalize(s: str) -> str:
+    """Normalizes whitespace within the provided string by converting all groups
+    of whitespaces into ' ', and uppercases the string."""
+    if s is None or s == '' or s.isspace():
+        raise ValueError(
+            'function normalize should never be called with None or an empty '
```
zacharylawrence: Nit: Thoughts on something like "Cannot normalize None or Empty String" since then it might fit on one line? Not a big deal and I'm fine with it as is if you like that more.

PR #758, opened 2019-01-28
add convenience method to check if enum parses

### LINKED ISSUE #695: There should be a simple way to check if a scraped string will be a mapped to a valid enum value
Currently, the only way to check this is to wrap a call to an enum object's `from_str()` method in a try and then catch the EnumParsingError. However, this doesn't check if the value is in the jurisdiciton's enum_overrides. There should be a unified, simple way to do a complete check.

### LINKED ISSUE #757: MappableEnum.from_str should require an overrides map
so we don't have bugs like #755.

This is related to #695; scraper writers shouldn't have to call from_str.

colincadams:
closes #695, closes #757


recidiviz/common/constants/mappable_enum.py
```diff
@@ -55,6 +54,21 @@ def from_str(cls,
...
+    def can_parse(
+            cls, label: str,
+            override_map: Dict[str, Optional['MappableEnum']]) -> bool:
+        """Checks if the given string will parse into this enum.
+
+        Convenience method to be used by a child scraper to tell if a given
+        string should be used for this field.
+        """
+        try:
+            cls.parse(label, override_map or {})
```
zacharylawrence: As discussed offline, consider making override_map required

colincadams: Done (I had changed the prototype but forgot to remove this default)

PR #760, opened 2019-01-28
Updates necessary for net data vendor

terinpw:
- Handle ResponseType.Raw in BaseScraper
- Dynamically find key numbers from the html itslef.


recidiviz/ingest/base_scraper.py
```diff
@@ -122,6 +122,8 @@ def _fetch_content(self, endpoint, response_type, headers=None,
                 return -1, None
         if response_type is constants.ResponseType.TEXT:
             return response.text, cookies
+        if response_type == constants.ResponseType.RAW:
```
zacharylawrence: I think you should use `is` when comparing enums, since enum members are singletons: https://docs.python.org/3/library/enum.html#enum-members-aka-instances

recidiviz/ingest/vendors/net_data/net_data_scraper.py
```diff
@@ -149,6 +146,21 @@ def _get_person_tasks(content, task: Task) -> List[Task]:
     return task_list
 
 
+def _generate_url(endpoint: str, action: str) -> str:
+    # URLs for NetData always assume the form 'http://...../bok/xxx.xx'. Here we
+    # remove everything after the 'bok/' to create a new base for our url.
+    base = endpoint.rsplit('/', 1)[0]
+
+    # Actions can come in the form 'action', 'bok/action', or '/bok/action'. In
+    # all cases, we want normalized_action = 'action.
```
zacharylawrence: `'action` -> `'action'`?

PR #761, opened 2019-01-28
add last-in-first-out option to run_scraper

terinpw:


recidiviz/ingest/run_scraper.py
```diff
@@ -131,6 +134,12 @@ def _create_parser():
         '--log', required=False, default='INFO', type=logging.getLevelName,
         help='Set the logging level'
     )
+    parser.add_argument(
+        'lifo', requred=False, default=False, type=bool, action='store_true',
```
colincadams: nit: `default=False` and `type=bool` aren't needed when you use `store_true` as far as I know

PR #762, opened 2019-01-28
UsNcAlamance

ohinds:
A superion scraper.


recidiviz/ingest/us_nc_buncombe/us_nc_buncombe.yaml
```diff
@@ -0,0 +1,2 @@
+key_mappings:
```
jamwalla: I wonder if create_scraper shouldn't make these by default? or if I'm only noticing them when they're empty. anyways, this is for a different region so definitely shouldn't be in this pr, but while you're at it let's remove the alamance yaml as well

recidiviz/ingest/vendors/superion/superion_scraper.py
```diff
@@ -196,6 +199,35 @@ def populate_data(self, content, task: Task,
...
+                # Fill in charge class, if it's in the name
+                if charge.name:
+                    for charge_class in ChargeClass:
+                        # Skip searching for the string 'OTHER' in
+                        # charge name, as it's probably unrealted to
+                        # the charge class, even if it's there.
+                        if charge_class is ChargeClass.OTHER:
+                            continue
+
+                        if charge_class.value in charge.name:
```
jamwalla: maybe call `normalize` on charge.name (once I land #756), unless you're sure it's always uppercase and space-delimited

recidiviz/ingest/vendors/superion/superion_scraper.py
```diff
@@ -196,6 +199,35 @@ def populate_data(self, content, task: Task,
...
+                            charge.charge_class = charge_class.value
+                            break
+
+                # Transfer the bond type to charge status, if we
+                # detect that the bond type field contains a charge
+                # status enum value.
+                try:
+                    if charge.bond and charge.bond.bond_type:
+                        _ = ChargeStatus.from_str(charge.bond.bond_type)
```
jamwalla: use ChargeStatus.can_parse instead (thanks colin! #758)

PR #763, opened 2019-01-28
Texas parse our report date

arian487:
Adds functionality to parse date for texas


recidiviz/ingest/aggregate/regions/tx/tx_aggregate_ingest.py
```diff
@@ -15,32 +15,47 @@
...
 
 import recidiviz.common.constants.enum_canonical_strings as enum_strings
 from recidiviz.ingest.aggregate import aggregate_ingest_utils
 from recidiviz.persistence.database.schema import TxCountyAggregate
 
+DATE_PARSE_ANCHOR = 'Abbreviated Population Report for'
 
-def parse(filename: str, date_scraped: datetime.datetime) \
+
+def parse(filename: str) \
```
zacharylawrence: Does this fit on one line now?

recidiviz/ingest/aggregate/regions/tx/tx_aggregate_ingest.py
```diff
@@ -15,32 +15,47 @@
...
-    table['report_date'] = date_scraped
+    table['report_date'] = _parse_date(filename)
     table['report_granularity'] = enum_strings.monthly_granularity
 
     return {
         TxCountyAggregate: table
     }
 
 
+def _parse_date(filename: str):
```
zacharylawrence: Could you add the return type here too?

recidiviz/ingest/aggregate/regions/tx/tx_aggregate_ingest.py
```diff
@@ -15,32 +15,47 @@
...
+    with open(filename, 'rb') as f:
+        pdf = PdfFileReader(f)
+        page = pdf.getPage(0)
+        text = page.extractText()
+        lines = text.split('\n')
+        for index, line in enumerate(lines):
+            if DATE_PARSE_ANCHOR in line:
+                # The date is on the next line if anchor is present on the line
+                return dateparser.parse(lines[index+1]).date()
```
zacharylawrence: Consider raising an `AggregateDateParsingError` if we don't find the anchor? I think mypy will complain about this.

recidiviz/ingest/aggregate/regions/tx/tx_aggregate_ingest.py
arian487: Done.

recidiviz/ingest/aggregate/regions/tx/tx_aggregate_ingest.py
arian487: Done.

recidiviz/ingest/aggregate/regions/tx/tx_aggregate_ingest.py
```diff
@@ -15,32 +15,49 @@
...
-    table['report_date'] = date_scraped
+    table['report_date'] = _parse_date(filename)
     table['report_granularity'] = enum_strings.monthly_granularity
 
     return {
         TxCountyAggregate: table
     }
 
 
+def _parse_date(filename: str) -> datetime.date:
```
jessex: I'm surprised that Texas and Georgia have exactly the same structure. Why is that?

recidiviz/ingest/aggregate/regions/hi/hi_aggregate_ingest.py
```diff
@@ -51,20 +51,29 @@
...
     }
 
 
+def _parse_date(filename: str) -> datetime.date:
+    # PyPD2 can't extract text properly from hawaii, so we use the filename
+    # instead.
+    prefix_to_remove = 'Pop-Reports-EOM-'
+    filename = filename.split('/')[-1]
+    date_str = filename.strip('.pdf').strip(prefix_to_remove)
+    return dateparser.parse(date_str).date()
```
jessex: Should we proactively catch parse errors and propagate an `AggregateDateParsingError`?

recidiviz/ingest/aggregate/regions/tx/tx_aggregate_ingest.py
arian487: What do you mean?  The structure is actually different, its just the logic that is similar.  We discussed offline with @zacharylawrence  and we felt that the way to go for now is a dedicated function per scraper that looks for key words/phrases to find the date.  All encompassing regexs are a lot less readable, especially if they have to be able to parse out any date format in any arbitrary text.  This will result in somewhat duplicated code for a couple regions but its not too bad because its only 7 states in total for now.

recidiviz/ingest/aggregate/regions/hi/hi_aggregate_ingest.py
arian487: I think thats sensible, added it for all 3.

recidiviz/ingest/aggregate/regions/tx/tx_aggregate_ingest.py
jessex: Unless I'm crazy or Github is doing something odd, the logic for the two functions is exactly the same. They're both this:
```
def _parse_date(filename: str) -> datetime.date:
    with open(filename, 'rb') as f:
        try:
            pdf = PdfFileReader(f)
            page = pdf.getPage(0)
            text = page.extractText()
            lines = text.split('\n')
        except Exception as e:
            raise AggregateDateParsingError(str(e))
        for index, line in enumerate(lines):
            if DATE_PARSE_ANCHOR in line:
                # The date is on the next line if anchor is present on the line
                return dateparser.parse(lines[index+1]).date()
        raise AggregateDateParsingError('Could not extract date')
```

arian487: Ah I see what you're saying now, so the logic is the same, both of them have the date on a newline preceded by some string indicating the date was coming up next (DATE_PARSE_ANCHOR).  I guess I could factor this out into utils or something but it feels weird to have this live in utils, _parse_date_with_anchor would the name but the logic is really specific, the date has to be on a newline.  This logic only works for these two states I think so I guess I wasn't sure what a good home would be for this function.  

Thoughts?

jessex: Eh, you're fine! I don't think we need to DRY this up.

PR #772, opened 2019-01-29
Add us_ky_woodford scraper

KareemG:
Scraper for http://www.woodfordcountydetention.com/InmateRoster/WCDC_inmatelist.html (JailTracker).


recidiviz/ingest/us_ky_woodford/us_ky_woodford_scraper.py
```diff
@@ -0,0 +1,64 @@
...
+    def get_enum_overrides(self):
+        return {
+            # Charge Classes
+            'T': None,                              # "in transit"
+            'V': ChargeClass.INFRACTION,            # e.g. traffic violation
+            'C': ChargeClass.FELONY,                # e.g. controlled substance
+
+            # Charge Statuses
+            'DISMISSED': ChargeStatus.DROPPED,
+            'FUGITIVE': None,
```
KareemG: Not sure where this fits.

recidiviz/ingest/us_ky_woodford/us_ky_woodford_scraper.py
```diff
@@ -0,0 +1,64 @@
...
+            'GUILTY': ChargeStatus.CONVICTED,
+            'OPEN': None,                           # TODO(616): bond -> pending
+            'WAIVED': ChargeStatus.DROPPED,
+            'BLANKET BOND': None,
```
KareemG: What to do in this case, where one or two charges seem to cover the bonds for a bunch?

recidiviz/ingest/us_ky_woodford/us_ky_woodford_scraper.py
```diff
@@ -0,0 +1,64 @@
...
+            'ENHANCED': None,
```
KareemG: e.g. enhanced charge? "persistent felony offender ii"

recidiviz/ingest/us_ky_woodford/us_ky_woodford_scraper.py
```diff
@@ -0,0 +1,64 @@
...
+
+            # Court Types
+            'FEDERAL': None,
```
KareemG: Not sure if SUPERIOR, website has entries ingesting DISTRICT type.

recidiviz/ingest/us_ky_woodford/us_ky_woodford_scraper.py
arian487: @ohinds  will know best

recidiviz/ingest/us_ky_woodford/us_ky_woodford_scraper.py
arian487: So the relationship is one bond to many charges.  Generally each charge has a single bond, but some sites will aggregate the bonds for a bunch of charges into a single bond amount.  In this case, you can create one bond with the relevant value, and point many charges to that one bond and it will be reflected correctly in the database.

Can you show an example of what a blanket bond looks like on a site?

recidiviz/ingest/us_ky_woodford/us_ky_woodford_scraper.py
```diff
@@ -0,0 +1,64 @@
...
+
+            # Bond Types
+            'FULL CASH': BondType.CASH,
+            'NO BOND - OTHER': BondType.NO_BOND,
+            'NO BOND - MUST SEE PRETRIAL': BondType.NO_BOND,
+            'NO BOND - SENTENCED': BondType.NO_BOND,
+            'SURETY': BondType.UNSECURED,
```
arian487: Similar to the above comment, I think you can update the global map here.

recidiviz/ingest/us_ky_woodford/us_ky_woodford_scraper.py
KareemG: Here's an example of a set of blanket bonds with some having an amount of 0 (presumably because they relate to some other blanket bond...):

> [{'BondAmount': 25000.0,
  'BondType': 'BLANKET BOND',
  'Status': 'OPEN - ACTIVE'
   ...},
 {'BondAmount': 0.0,
  'BondType': 'BLANKET BOND',
  'Status': 'OPEN - ACTIVE'
   ...},
 {'BondAmount': 0.0,
  'BondType': 'BLANKET BOND',
  'Status': 'OPEN - ACTIVE'
   ...},
 {'BondAmount': 10000.0,
  'BondType': 'BLANKET BOND',
  'Status': 'OPEN - ACTIVE'
   ...},
 {'BondAmount': 10000.0,
  'BondType': 'BLANKET BOND',
  'Status': 'OPEN - ACTIVE'
   ...}]

recidiviz/ingest/us_ky_woodford/us_ky_woodford_scraper.py
ohinds: I would guess PRETRIAL?

recidiviz/ingest/us_ky_woodford/us_ky_woodford_scraper.py
ohinds: Blanket bond means exactly what @arian487 said. There are several charges that all share a bond. So I think this should get mapped to 'CASH'. The trick is to group the appropriate charges with the right 'blanket'. Is there more data in the JSON that would allow you to do that? Like some sort of ID or case number?

recidiviz/ingest/us_ky_woodford/us_ky_woodford_scraper.py
ohinds: FEDERAL --> DISTRICT, and also create a hold on the booking with Federal as the jurisdiction.

recidiviz/ingest/us_ky_woodford/us_ky_woodford_scraper.py
```diff
@@ -0,0 +1,64 @@
...
+            'BLANKET BOND': None,
+            'ENHANCED': None,
+
+            # Court Types
+            'FEDERAL': None,
+
+            # Bond Types
+            'FULL CASH': BondType.CASH,
+            'NO BOND - OTHER': BondType.NO_BOND,
+            'NO BOND - MUST SEE PRETRIAL': BondType.NO_BOND,
```
ohinds: I thinking this is PENDING

recidiviz/ingest/us_ky_woodford/us_ky_woodford_scraper.py
ohinds: We recently found that this is SECURED. But we might have fixed that already.

recidiviz/ingest/us_ky_woodford/us_ky_woodford_scraper.py
```diff
@@ -0,0 +1,64 @@
...
+            'NO BOND - SENTENCED': BondType.NO_BOND,
```
ohinds: This is fine, but could you also change the charge status to SENTENCED here?

PR #774, opened 2019-01-29
Zuercher vendor and us_ga_douglas

colincadams:


recidiviz/ingest/vendors/zuercher/zuercher.yaml
```diff
@@ -0,0 +1,15 @@
+key_mappings:
+  records.name: person.full_name
+  records.dob: person.age
+  records.sex: person.gender
+  records.cell_block: booking.facility
```
jamwalla: this is probably more specific than "facility", like a location in the facility - but I couldn't find one easily. Do you know what values this has?

recidiviz/ingest/vendors/zuercher/zuercher_scraper.py
```diff
@@ -0,0 +1,291 @@
...
+    def _preprocess_json(content):
+        records = content['records']
+        for record in records:
+            hold_reasons_text = record.pop('hold_reasons')
+            if hold_reasons_text:
+                hrs = hold_reasons_text.split('<br />')
+                record['charges'] = [hr.strip() for hr in hrs if hr]
+        return content
+
+    def _postprocess_charge(self, booking: _Booking, charge: _Charge):
```
colincadams: @jamwalla lmk if you think it's possible to munge this and get it to work with the data extractor (see test fixture for examples). I thought about munging the semicolons and such but it seemed like since the format varies based on the key it would be just as or more cumbersome. Maybe you have some creative ideas? 😁

recidiviz/ingest/us_ga_douglas/us_ga_douglas_scraper.py
```diff
@@ -0,0 +1,46 @@
...
+    """Scraper implementation for us_ga_douglas."""
+    def __init__(self):
+        super(UsGaDouglasScraper, self).__init__(
+            region_name='us_ga_douglas', jurisdiction_name='Douglas County, GA')
+
+    def get_enum_overrides(self):
+        return {
+            **super(UsGaDouglasScraper, self).get_enum_overrides(),
+            # Races
+            'NOT SPECIFIED': None,
```
colincadams: @ohinds I wasn't sure what the proper value was for some of these enums. lmk what you think

recidiviz/ingest/vendors/zuercher/zuercher_scraper.py
```diff
@@ -0,0 +1,291 @@
...
+        text, *rest = map(str.strip, text.split('('))
+        for item in rest:
+            item = item.rstrip(')')
+            if ChargeStatus.can_parse(item, self.get_enum_overrides()):
+                charge.status = item
+            elif ChargeClass.can_parse(item, self.get_enum_overrides()):
+                charge.charge_class = item
+        return text
+
+    def get_enum_overrides(self):
```
jamwalla: this shouldnt be necessary

recidiviz/ingest/vendors/zuercher/zuercher_scraper.py
jamwalla: ahh, I think this is probably the best approach, you'd end up doing just as much work preprocessing but this is pretty readable

recidiviz/common/constants/person.py
```diff
@@ -75,6 +75,7 @@ def _get_default_map():
     'NATIVE AM': Race.AMERICAN_INDIAN_ALASKAN_NATIVE,
     'INDIAN': Race.AMERICAN_INDIAN_ALASKAN_NATIVE,
     'ASIAN': Race.ASIAN,
+    'ASIAN/PACIFIC ISLANDER': Race.ASIAN,
```
jessex: This is another case where #733 would be helpful. This map shouldn't need all 3 of these combinations and the assuredly more to come.

recidiviz/ingest/vendors/zuercher/zuercher.yaml
jessex: Holding my breath for the other values, but yeah this is almost certainly sub-facility. There's potentially an entirely entity we could add to our schema related to the physical holding location and conditions, if it were valuable enough to track.

recidiviz/ingest/vendors/zuercher/zuercher_scraper.py
```diff
@@ -0,0 +1,291 @@
...
+import pytz
+
+from recidiviz.common.constants.charge import ChargeClass, ChargeStatus
+from recidiviz.ingest import constants
+from recidiviz.ingest.base_scraper import BaseScraper
+from recidiviz.ingest.extractor.json_data_extractor import JsonDataExtractor
+from recidiviz.ingest.models.ingest_info import IngestInfo, _Booking, _Charge
+from recidiviz.ingest.task_params import ScrapedData, Task
+
+BATCH_SIZE = 100
```
jessex: This should be private, no?

recidiviz/ingest/vendors/zuercher/zuercher_scraper.py
```diff
@@ -0,0 +1,291 @@
...
+
+class ZuercherScraper(BaseScraper):
+    """Scraper for counties using Zuercher."""
+    yaml = os.path.join(os.path.dirname(__file__), 'zuercher.yaml')
+
+    CHARGE_KEY = 'Charge:'
```
jessex: Curious what the pattern is on the website for when a colon is appended or not. Is it just pure inconsistency or is there reason to this?

recidiviz/ingest/vendors/zuercher/zuercher_scraper.py
```diff
@@ -0,0 +1,291 @@
...
+        )
+
+    def get_more_tasks(self, content, task: Task) -> List[Task]:
+        start = 0
+        if task.json:
+            start = task.json['paging']['start'] + BATCH_SIZE
+
+        task_type = constants.TaskType.SCRAPE_DATA
+        if self.is_initial_task(task.task_type) or \
+                start + BATCH_SIZE - 1 < content['total_record_count']:
```
jessex: Stylistic nit but I think `start + BATCH_SIZE <= content['total_record_count']` reads more obviously than future developers needing to explicitly think about the `- 1` when they see this line.

recidiviz/ingest/vendors/zuercher/zuercher_scraper.py
```diff
@@ -0,0 +1,291 @@
...
+                      ingest_info: IngestInfo) -> Optional[ScrapedData]:
+        content = self._preprocess_json(content)
+        data_extractor = JsonDataExtractor(self.yaml)
+        ingest_info = data_extractor.extract_and_populate_data(
+            content, ingest_info)
+        for person in ingest_info.people:
+            for booking in person.bookings:
+                for charge in booking.charges:
+                    self._postprocess_charge(booking, charge)
+        # print(ingest_info)
```
jessex: Remember to drop this before merging.

colincadams: done.

recidiviz/ingest/vendors/zuercher/zuercher_scraper.py
colincadams: good catch! fixed

recidiviz/ingest/vendors/zuercher/zuercher_scraper.py
colincadams: It seems quite random to me - the author might have a reason but I couldn't figure it out. At first I thought it was that keys that were valid to start a line had colons, but that was proven wrong by quite a few.

recidiviz/ingest/vendors/zuercher/zuercher_scraper.py
colincadams: fixed

recidiviz/ingest/vendors/zuercher/zuercher_scraper.py
colincadams: removed

recidiviz/ingest/vendors/zuercher/zuercher.yaml
colincadams: This region doesn't ever set the cell_block field, although there are some valid values listed in the response from the `init` request. All of those start with `DCSO` which makes me assume they just have one jail facility at the sheriff's office. I can remove the key for the vendor overall and then take another look with the next region as some of them set it.

recidiviz/ingest/vendors/zuercher/zuercher.yaml
```diff
@@ -0,0 +1,15 @@
...
+  records.sex: person.gender
+  records.release_date: booking.release_date
+  records.held_for_agency: hold.jurisdiction_name
+  records.race: person.race
+  records.arrest_date: booking.admission_date
+  records.charges: charge.name
+
+keys_to_ignore:
+  - records.cell_block
+  - records.is_juvenile
```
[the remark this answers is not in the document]
colincadams: Where should it be tracked? I figured since we are putting age into the `ingest_info` we can always tell if someone is a juvenile. Or is it worth adding a separate field to see if they explicitly tell us they are a juvenile?

recidiviz/ingest/vendors/zuercher/zuercher_scraper.py
```diff
@@ -0,0 +1,292 @@
...
+    # Middle of line keys
+    ARREST_DATE_KEY = 'Arrest Date'
+    BOND_KEY = 'Bond'
+    SET_BY_KEY = 'Set By'
+
+    # Suffixes
+    COUNTS_SUFFIX = 'Counts'
+    DEGREE_SUFFIX = 'Degree'
+
+    def __init__(self, region_name, jurisdiction_name):
```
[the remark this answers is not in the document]
colincadams: Done. I like that better!

PR #780, opened 2019-01-30
Support running multiple regions in run_scraper

terinpw:


terinpw:
> Can you update the help string for the `--region` flag with this new functionality?
> 
> Otherwise LGTM :shipit:

Good point! Done

PR #781, opened 2019-01-30
update ingest_info entities to all be public

terinpw:
While writing child scrapers, there have been several instances where creating objects directly would be much clearer/concise rather than relying on explicit `create_child()` methods on parent objects.


recidiviz/tests/ingest/models/ingest_info_test.py
```diff
@@ -56,20 +56,20 @@ def _verify_fields(proto, ingest_info_source, ignore=None):
...
-        sentence = charge.create_sentence()
-        bond = charge.create_bond()
-        _verify_fields(Person, person, person_fields_ignore)
-        _verify_fields(Booking, booking, booking_fields_ignore)
-        _verify_fields(Charge, charge, charge_fields_ignore)
-        _verify_fields(Hold, hold)
-        _verify_fields(Arrest, arrest)
-        _verify_fields(Sentence, sentence)
-        _verify_fields(Bond, bond)
+        # person = IngestInfo().create_person()
```
zacharylawrence: Should these be commented out?

PR #788, opened 2019-01-30
All brooks jeffrey regions in al

terinpw:


region_manifest.yaml
```diff
@@ -339,3 +339,75 @@ regions:
     base_url: https://apps.alamance-nc.com/p2c
     scraper_package: us_nc_alamance
     timezone: America/New_York
+us_al_autauga:
+    agency_name: Autagua County Sheriff's Office
```
jamwalla: Autauga

PR #791, opened 2019-01-31
Create endpoint to scrape websites and find their PDFs

arian487:
Currently only added texas.  The logic uses the modified timestamp from the historical bucket and passes that in with the If-Modified-Since header which will abort a download if the file on the server is newer.


arian487:
This is now ready for review

recidiviz/ingest/aggregate/regions/tx/tx_aggregate_site_scraper.py
```diff
@@ -0,0 +1,40 @@
...
+"""Scrapes the texas aggregate site and finds pdfs to download."""
+from typing import Set
+from lxml import html
+import requests
+
+STATE_AGGREGATE_URL = 'https://www.tcjs.state.tx.us/index.php?linkID=326'
+PDF_URL_TEMPLATE = 'https://www.tcjs.state.tx.us/docs/AbbreviatedPopReports/{}'
+
+
+# TODO: fill this in with logic to scrape and return the URLs.
```
jessex: This TODO can be removed now, right?

recidiviz/ingest/aggregate/scrape_aggregate_reports.py
```diff
@@ -0,0 +1,108 @@
...
+    state = get_value('state', request.args)
+    urls = state_to_scraper[state]()
+    gcp_project = metadata.project_id()
+    historical_bucket = HISTORICAL_BUCKET.format(gcp_project)
+    upload_bucket = UPLOAD_BUCKET.format(gcp_project)
+    fs = gcsfs.GCSFileSystem(project=gcp_project)
+    logging.info('Scraping all pdfs for %s', state)
+
+    for url in urls:
+        pdf_name = url.split('/')[-1].lower()
```
jessex: This assumes there can be no `/` in the PDF filename itself, but delimiters vary by environment. I assume this is relying on the fact that Google Cloud's filesystem does _not_ allow slashes in filenames?

recidiviz/ingest/aggregate/scrape_aggregate_reports.py
```diff
@@ -0,0 +1,108 @@
...
+            logging.info(
+                'Skipping %s because the file already exists', url)
+
+    return '', HTTPStatus.OK
+
+
+def _get_file_to_upload(
+        path: str, fs: gcsfs.GCSFileSystem, url: str, pdf_name: str) \
+        -> Optional[str]:
+    """This function checks first whether it needs to download, and then
```
jessex: The logic for avoiding duplication relies on the state website never renaming older PDFs, right? Otherwise, if a PDF is renamed then we will say it doesn't exist yet and return it as a file to upload, and it will get added anew to the bucket and parsed immediately. Now, depending on how the report date is parsed for that state, it's possible that the report date would remain the same (e.g. if it was not parsed from the filename or parsed from a part of the filename that was unchanged) and then when we tried to persist data into the schema it would fail the key constraint. But in other cases, it would throw duplicate data into the database. Does that all sound right? If so, have you given any thought to these edge cases yet?

recidiviz/ingest/aggregate/regions/ga/ga_aggregate_site_scraper.py
```diff
@@ -0,0 +1,26 @@
...
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+# =============================================================================
+
+"""Scrapes the georgia aggregate site and finds pdfs to download."""
+from typing import List
+
+STATE_AGGREGATE_URL = 'https://www.dca.ga.gov/node/3811/documents/2086'
+
+
+# TODO: fill this in with logic to scrape and return the URLs.
```
jessex: Are we planning to fill these in in this PR or later on?

recidiviz/ingest/aggregate/regions/tx/tx_aggregate_site_scraper.py
arian487: Done.

recidiviz/ingest/aggregate/scrape_aggregate_reports.py
arian487: Yea it does rely on that.  None of the state aggregate sites have that pattern so personally I am ok with a solution that may break but probably won't since these states have kept up their pattern for years (decades some of them).  Also even if the filename has a / in it, it doesn't prevent us from downloading the file.  We still use the URL as is, we just get the name wrong (but we get it consistently wrong so it still processes the file the same).

recidiviz/ingest/aggregate/regions/ga/ga_aggregate_site_scraper.py
jessex: Cool. Do you mind putting the issue number on each TODO to be thorough? 🙏

recidiviz/ingest/aggregate/scrape_aggregate_reports.py
jessex: Oh snap! That all sounds like exactly the desired behavior. Well done :)

recidiviz/ingest/aggregate/regions/ga/ga_aggregate_site_scraper.py
arian487: Done.

recidiviz/ingest/aggregate/regions/fl/fl_aggregate_site_scraper.py
```diff
@@ -0,0 +1,26 @@
+# Recidiviz - a platform for tracking granular recidivism metrics in real time
+# Copyright (C) 2018 Recidiviz, Inc.
```
zacharylawrence: @ohinds suggests updating this on new files to `2019`

recidiviz/ingest/aggregate/regions/tx/tx_aggregate_site_scraper.py
```diff
@@ -0,0 +1,39 @@
...
+PDF_URL_TEMPLATE = 'https://www.tcjs.state.tx.us/docs/AbbreviatedPopReports/{}'
+
+
+def get_urls_to_download() -> Set[str]:
+    """Scrapes the aggregate report pdfs from the state website"""
+    page = requests.get(STATE_AGGREGATE_URL).text
+    html_tree = html.fromstring(page)
+    links = html_tree.xpath('//a/@href')
+
+    ret_links = set()
```
zacharylawrence: What does `ret` mean here?

recidiviz/ingest/aggregate/scrape_aggregate_reports.py
zacharylawrence: > it would fail the key constraint

Right, to add to what @arian487 was saying: if we try to write a row with a conflicting `UniqueConstraint`, the db write will fail but we will continue writing the rest of the rows (and call the PDF parse as a whole successful).

You can see how this happens here: [database.py#L273](https://github.com/Recidiviz/recidiviz-data/blob/2e9d805658415e63ee5c6636c277d1981c303a3e/recidiviz/persistence/database/database.py#L273)

recidiviz/tests/ingest/aggregate/regions/tx/tx_aggregate_site_scraper_test.py
```diff
@@ -0,0 +1,43 @@
...
+    @patch.object(requests, 'get')
+    def testGetAllUrls(self, mockget):
+        mockresponse = Mock()
+        mockget.return_value = mockresponse
+        mockresponse.text = REPORTS_HTML
+        url1 = ('https://www.tcjs.state.tx.us/docs/AbbreviatedPopReports/'
+                'Abbreviated Pop Rpt June 2017.pdf')
+        url2 = ('https://www.tcjs.state.tx.us/docs/AbbreviatedPopReports/'
+                'Abbreviated Pop Rpt Jan 2016.pdf')
+        expected_urls = set([url1, url2])
```
zacharylawrence: You can define a set literal like: `{url1, url2}`

recidiviz/ingest/aggregate/regions/fl/fl_aggregate_site_scraper.py
arian487: Done.

recidiviz/ingest/aggregate/regions/tx/tx_aggregate_site_scraper.py
arian487: return, I've updated the name to be not so bad haha.

recidiviz/tests/ingest/aggregate/regions/tx/tx_aggregate_site_scraper_test.py
arian487: Done.

recidiviz/ingest/aggregate/scrape_aggregate_reports.py
colincadams: Are there any reports where the date is built in to the directory structure? i.e.:
```
jail-reports
|-- 2017
|   |-- july.pdf
|   `-- december.pdf
|-- 2018
|   |-- january.pdf
|   |-- march.pdf
|   `-- august.pdf
`-- 2019
    `-- january.pdf
```
That seems like a reasonable, not that unlikely file structure that wouldn't work by just using the last part.

Is there a downside to instead just use everything past the domain and replacing `'/'` with another character like `'_'`? i.e. `urlparse(url).path.replace('/', '_')`

arian487: I think this is sensible!  Done.

Issue #698, opened 2019-01-22
Add fips to aggregated report parsers


zacharylawrence:
We currently just set the fips field to `None`

PR #793, opened 2019-01-31
Add fips mapping

zacharylawrence:
#698


recidiviz/persistence/database/schema.py
```diff
@@ -784,7 +784,9 @@ class TxCountyAggregate(Base, _AggregateTableMixin):
     """TX state-provided aggregate statistics."""
     __tablename__ = 'tx_county_aggregate'
     __table_args__ = (
-        UniqueConstraint('fips', 'report_date', 'report_granularity'),
+        UniqueConstraint(
+            'fips', 'county_name', 'report_date', 'report_granularity'
+        ),
     )
 
     county_name = Column(String(255))
```
zacharylawrence: Should this be renamed to `facility_name` since the "County" column in the PDF contains rows like: `McLennan` and `McLennan 1(P)` and `McLennan 2(P)`

jessex:
General question: how often does the FIPS mapping change? I'm guessing existing codes never change, so the mapping only changes when new counties are incorporated (or existing ones are de-corporated, or two are merged which is essentially a combination of the previous two). The failure mode is that a new county is incorporated in one of the states we ingest aggregate data for, and we start failing to get a fips code.

recidiviz/ingest/aggregate/fips.py
```diff
@@ -0,0 +1,73 @@
...
+    old_index = df.index
+    df.index = county_names.apply(_sanitize_county_name)
+
+    df = _fuzzy_join(df, _get_fips_for(state))
+
+    df.index = old_index
+    return df
+
+
+def _get_fips_for(state: us.states) -> pd.DataFrame:
```
jessex: What's the behavior here if the state name is somehow not found? I know it's truly exceptional but just in case there's a human error setting up a new state in the future. It looks like it would be an error on line 49 but I could be mistaken.

recidiviz/ingest/aggregate/fips.py
```diff
@@ -0,0 +1,73 @@
...
+    df1.index = df1.index.map(lambda x: _best_match(x, df2.index))
+    return df1.join(df2)
+
+
+def _best_match(item: Any, iterable: Iterable[Any]) -> Any:
+    """Returns the closest match of |item| in |iterable|."""
+    close_matches = difflib.get_close_matches(item, iterable)
+
+    if not close_matches:
+        raise FipsMergingError('Fail to match: {} in {}'.format(item, iterable))
```
jessex: So this gets thrown if we ingest a county name that doesn't exist _and_ that is sufficiently far from existing county names?

recidiviz/ingest/aggregate/fips.py
```diff
@@ -0,0 +1,73 @@
+# Recidiviz - a platform for tracking granular recidivism metrics in real time
```
jessex: Can we add unit tests for this package?

recidiviz/ingest/aggregate/fips.py
```diff
@@ -0,0 +1,73 @@
...
+
+    df = _fuzzy_join(df, _get_fips_for(state))
+
+    df.index = old_index
+    return df
+
+
+def _get_fips_for(state: us.states) -> pd.DataFrame:
+    """Get the [county_name, fips] df, filtering for the given |state|."""
+    fips = _FIPS[_FIPS.state_code == int(state.fips)]
```
ohinds: you could use state.fips.astype(int) here.

recidiviz/ingest/aggregate/fips.py
ohinds: Have you played around with the cutoff parameter at all? It would be nice to know how wrong it could be before not returning a match.

recidiviz/persistence/database/schema.py
ohinds: Yes!

zacharylawrence:
> Have you played around with the cutoff parameter at all? It would be nice to know how wrong it could be before not returning a match.

Yeah, this is a good point. I tightened this up as far as I could to 0.75 / 1.00 (from the default of 0.6). I think it's reasonable to start with a value as close to 1.00 as possible then relax it as needed since relaxing this value could never introduce mismatches in previous scrapers. We know this because verifying that a match can already be made guarantees that it will always be the best match.

recidiviz/ingest/aggregate/fips.py
jessex: Nope, perfect.

PR #797, opened 2019-01-31
parse pdfs during test run instead of collection

colincadams:


recidiviz/tests/ingest/aggregate/regions/fl/fl_aggregate_ingest_test.py
```diff
@@ -17,29 +17,32 @@
 """Tests for fl_aggregate_ingest.py."""
 import datetime
 from unittest import TestCase
-import numpy as np
 
+import numpy as np
```
colincadams: sorry about the import noise

PR #801, opened 2019-01-31
Set fips on all aggregate ingest

### LINKED ISSUE #698: Add fips to aggregated report parsers

zacharylawrence:
Set fips on aggregate ingest & set fields as non-null in the schema.

fixes #698


recidiviz/ingest/aggregate/regions/fl/fl_aggregate_ingest.py
```diff
@@ -164,6 +167,15 @@ def _to_int(int_str: str) -> Optional[int]:
         return None
 
 
+def _pretend_facility_is_county(facility_name: str):
+    """Format facility_name like a county_name to match each to a fips."""
+    words_before_county_name = ['County', 'Conte', 'CSO', 'Central']
+    for deliminator in words_before_county_name:
```
jessex: Should this be `delimiter`?

recidiviz/ingest/aggregate/regions/hi/hi_aggregate_ingest.py
```diff
@@ -98,6 +126,11 @@ def _parse_table(filename: str) -> pd.DataFrame:
     result = head_count_ending_df.append(contracted_facilities_df,
                                          ignore_index=True)
 
+    result['fips'] = result.facility_name.map(
```
jessex: So these both map to `None` when we encounter a new acronym for the first time, right? Should we add in some sort of non-failing error/alert so we at least know when this occurs?

recidiviz/persistence/database/schema.py
```diff
@@ -624,11 +624,12 @@ def __new__(cls, *_, **__):
 
     # TODO(#689): Ensure that `fips` also supports facility level fips
     # TODO(#689): Consider adding fips_type to denote county vs facility
-    fips = Column(String(255))
+    fips = Column(String(255), nullable=False)
```
ohinds: What is the benefit of string versus int here?

recidiviz/ingest/aggregate/regions/fl/fl_aggregate_ingest.py
zacharylawrence: Yes, good catch!

recidiviz/ingest/aggregate/regions/hi/hi_aggregate_ingest.py
zacharylawrence: Like how the rest of the aggregate scraper work, I think raising a `FipsMergingError` would make sense here and be consistent.

recidiviz/persistence/database/schema.py
jessex: I think preserving the 0 is necessary, and I think it would be overkill to define a custom type for this. Postgres does have zero padding for integers but it involves casting to chars anyway, so I think this is reasonable.

zacharylawrence: @jessex Yeah, that sounds reasonable to me. I've made #821 to track changing a few place where we already introduced FIPS as an int (eg. `fips.csv` has no leading `0`).

PR #802, opened 2019-01-31
Refactor snapshot writer to use context registry

zdg2102:
This is a no-op change. It updates `update_historical_snapshots` to store the snapshots in a context registry, but doesn't change how those snapshots are fetched or used. Those changes will come as separate followups.


arian487:
Can you add some colour in the description to decribe why exactly we made this change?

zdg2102:
@arian487 for every snapshot operation, we need three pieces of information that come from three different places: the new entity, the latest snapshot, and any special values to determine if a new snapshot needs to have its period columns manually adjusted.

Because we want to fetch all the snapshots in bulk, and because figuring out date adjustments requires looking at the whole record tree rather than a single entity, we can't just go through the process of fetching these values for each entity separately. We need to do each of them as a bulk operation for all the entities we're concerned with.

The context registry lets us keep these values organized as we go through each step, keeping them associated with the right entity. Then at the end we can just iterate over the registry and handle each entity in isolation, because each one has all the required context values bundled with it.

recidiviz/persistence/database/database.py
```diff
@@ -162,86 +164,42 @@ def write_person(session, person):
...
 
     Returns:
-        persisted Person schema object
+        persisted (Person) schema object
     """
-    return _save_record_tree(session, database_utils.convert_person(person))
+    persisted_people = _save_record_trees(
+        session, [database_utils.convert_person(person)])
+    # persisted_people will only contain the single person passed in
+    return persisted_people[0]
```
terinpw: nit - let use more_itertools.one() for this.

jessex:
@zdg2102 just a heads up that I want to take a look at both this and #818 later today but haven't had the time yet. 🙌

terinpw:
Also just for record keeping - @zacharylawrence and I are going to walk through this with @zdg2102 on Monday as well.

recidiviz/persistence/database/database.py
```diff
@@ -162,86 +164,42 @@ def write_person(session, person):
...
         session: (Session)
-        person: Person schema object
+        root_people: list of (Person) schema objects
 
     Returns:
-        persisted Person schema object
+        list of persisted (Person) schema objects
     """
 
-    logging.info('Starting merge and flush for record tree')
```
jessex: Can we at least return logging to denote the batch updates about to begin and end?

recidiviz/persistence/database/update_historical_snapshots.py
```diff
@@ -7,113 +19,149 @@
...
+def _execute_action_for_all_entities(root_people, action, *args):
+    """For every entity in every record tree rooted at one of |root_people|,
+    invokes |action|, passing the entity and |*args| as arguments"""
 
-def _get_most_recent_historical_snapshot(session, entity,
-                                         historical_table_class):
-    """Returns the most recent historical snapshot corresponding to |entity| if
-    one exists, otherwise returns (None).
-    """
+    unprocessed = [person for person in root_people]
```
jessex: What's the advantage of this over `unprocessed = root_people.copy()`?

recidiviz/persistence/database/update_historical_snapshots.py
```diff
@@ -7,113 +19,149 @@
...
-        session.merge(most_recent_historical_snapshot)
+    if context.most_recent_snapshot is not None:
+        context.most_recent_snapshot.valid_to = snapshot_time
+        session.merge(context.most_recent_snapshot)
 
-    logging.info(
-        'Finished merging historical snapshots for entity: %s %s',
-        master_table_class_name, master_entity_id)
 
+def _execute_action_for_all_entities(root_people, action, *args):
```
jessex: Let's make sure to add type annotations throughout this module.

recidiviz/persistence/database/update_historical_snapshots.py
```diff
@@ -155,3 +203,96 @@ def _get_shared_column_property_names(entity_class_a, entity_class_b):
...
+
+    def register_entity(self, entity):
+        """Creates (_SnapshotContext) for |entity| and adds it to registry
+
+        Raises (ValueError) if |entity| has already been registered
+        """
+        key = self._key_for_master_entity(entity)
+        if key in self.snapshot_contexts:
+            raise ValueError(
+                'Entity {entity} already registered with key {key}'.format(
```
jessex: I'm sure I just lost track of this in the past couple of months, but our schema is set up such that no two parent entities can refer to the same child entity, right? A primary key won't ever show up in two separate record trees, but only in a single record tree?

recidiviz/persistence/database/update_historical_snapshots.py
```diff
@@ -155,3 +203,96 @@ def _get_shared_column_property_names(entity_class_a, entity_class_b):
...
+        Raises (ValueError) if no master entity has been registered
+        corresponding to |snapshot|
+        """
+        master_entity_key = self._master_key_for_snapshot(snapshot)
+        if master_entity_key not in self.snapshot_contexts:
+            raise ValueError(
+                'No master entity corresponding with snapshot {snapshot} '
+                'registered with key {key}'.format(
+                    snapshot=snapshot, key=master_entity_key))
+        self.snapshot_contexts[master_entity_key].most_recent_snapshot = \
```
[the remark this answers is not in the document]
jessex: Ah I see now that we only ever enter `add_snapshot` for the results of a call to fetch the most recent snapshot for a given entity, so this cannot happen.

recidiviz/persistence/database/update_historical_snapshots.py
```diff
@@ -7,113 +19,149 @@
...
     # Historical table does not need to be updated if entity is not new and
     # its current state matches its most recent historical snapshot
-    if most_recent_historical_snapshot is not None and \
+    if context.most_recent_snapshot is not None and \
             _does_entity_match_historical_snapshot(
-                    entity, most_recent_historical_snapshot):
+                    context.entity, context.most_recent_snapshot):
         return
 
-    logging.info(
```
jessex: I'm assuming we're planning to return this logging once we resolve the TODO at the top of this method? It seems like very useful tracing to have.

recidiviz/persistence/database/update_historical_snapshots.py
zdg2102: Right now that's the case, every record tree is completely separate and nothing can ever be shared. We may end up complicating that slightly if we have to add a separate table for facilities or jurisdictions, which would be shared between record trees. However, if we did that, that relationship would very likely not be included in the ORM model (since there's no benefit to including constant tables that never change when we're fetching data to perform updates), so it wouldn't show up in the tree traversal here anyway

recidiviz/persistence/database/database.py
zdg2102: The equivalent logging has been moved to the start of `update_historical_snapshots` now

recidiviz/persistence/database/update_historical_snapshots.py
zdg2102: Yep. I'm going to clean up all the logging once the basic flow is in place

recidiviz/persistence/database/database.py
```diff
@@ -162,86 +164,42 @@ def write_person(session, person):
...
-
-    # Merge includes all related entities, so this persists the master record
-    # tree
-    person = session.merge(person)
-
-    logging.info('Merge complete')
+    # Merge includes all related entities, so this persists all master entities
+    # in all record trees
+    root_people = [session.merge(root_person) for root_person in root_people]
```
zacharylawrence: Do we need to merge here to force the creation of primary_keys?

recidiviz/persistence/database/update_historical_snapshots.py
```diff
@@ -155,3 +203,96 @@ def _get_shared_column_property_names(entity_class_a, entity_class_b):
...
+    def __init__(self, entity=None, most_recent_snapshot=None,
+                 provided_valid_from=None,
+                 provided_valid_to=None):
+        self.entity = entity
+        self.most_recent_snapshot = most_recent_snapshot
+        self.provided_valid_from = provided_valid_from
+        self.provided_valid_to = provided_valid_to
+
+
+class _SnapshotContextRegistry:
```
zacharylawrence: Thoughts on moving this out into its own file?

recidiviz/persistence/database/update_historical_snapshots.py
```diff
@@ -155,3 +203,96 @@ def _get_shared_column_property_names(entity_class_a, entity_class_b):
...
+        if master_entity_key not in self.snapshot_contexts:
+            raise ValueError(
+                'No master entity corresponding with snapshot {snapshot} '
+                'registered with key {key}'.format(
+                    snapshot=snapshot, key=master_entity_key))
+        self.snapshot_contexts[master_entity_key].most_recent_snapshot = \
+            snapshot
+
+    @staticmethod
+    def _key_for_master_entity(entity):
```
zacharylawrence: As discussed offline, consider using a Dict[Dict] instead of string concatenation

recidiviz/persistence/database/update_historical_snapshots.py
```diff
@@ -155,3 +203,96 @@ def _get_shared_column_property_names(entity_class_a, entity_class_b):
...
+    """Returns ORM class of master table associated with the historical table of
+    |historical_class|
+    """
+    # See module assumption #1
+    master_class_name = historical_class.__name__.replace(
+        _HISTORICAL_TABLE_CLASS_SUFFIX, '')
+    return getattr(schema, master_class_name)
+
+
+class _SnapshotContext:
```
zacharylawrence: Can this be an `attrs` class? Also, thoughts on moving it into the same file as `_SnapshotContextRegistry` or a new file entirely?

recidiviz/persistence/database/update_historical_snapshots.py
```diff
@@ -7,113 +19,149 @@
...
-    return session.query(historical_table_class) \
-        .filter(text(filter_statement)) \
-        .order_by(historical_table_class.valid_from.desc()) \
-        .first()
+    related_entities = []
+
+    for relationship_name in entity.get_relationship_property_names():
+        related = getattr(entity, relationship_name)
+
+        # Relationship can return either a list or a single item
```
zacharylawrence: Is it possible to replace this comment with types?

recidiviz/persistence/database/update_historical_snapshots.py
```diff
@@ -7,113 +19,149 @@
...
-    """
+    unprocessed = [person for person in root_people]
+    processed = []
+    while unprocessed:
+        entity = unprocessed.pop()
+        action(entity, *args)
+        processed.append(entity)
 
-    key_column_name = type(entity).get_primary_key_column_name()
+        unprocessed.extend([related_entity for related_entity
```
zacharylawrence: Could we get by using a `filter` here? http://book.pythontips.com/en/latest/map_filter.html#filter

recidiviz/persistence/database/update_historical_snapshots.py
```diff
@@ -7,113 +19,149 @@
...
+    any required historical snapshot updates.
+
+    If any entity has no existing historical snapshots, an initial snapshot will
+    be created for it.
+
+    If any column of an entity differs from its current snapshot, the current
+    snapshot will be closed and a new snapshot will be opened corresponding to
+    the updated entity.
+
+    If neither of these cases applies, no action will be taken on the entity.
```
terinpw: nit - either in the docstring here or in comments below, can you explicitly lay out the 4 steps we talked about? 

1) Fetching/registering persistence entities
2) Fetching/registering snapshot entities
3) (Not done) Backdating
4) Persisting to the db?

recidiviz/persistence/database/update_historical_snapshots.py
terinpw: +1

recidiviz/persistence/database/database.py
```diff
@@ -141,16 +141,18 @@ def _query_people_and_open_bookings(session, region):
...
         session: (Session)
         people:  List[entities.Person]
+
+    Returns:
+        list of persisted (Person) schema objects
     """
-    for person in people:
-        write_person(session, person)
+    return _save_record_trees(
+        session, [database_utils.convert_person(person) for person in people])
```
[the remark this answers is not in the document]
zacharylawrence: As discussed offline, a list comprehension seems to be prefered here.

recidiviz/persistence/database/update_historical_snapshots.py
zdg2102: I'd rather avoid comments that just describe what the code is doing, since it means there's a danger of them getting out of sync. I also think that the way this function is structured is about as explicit an outline of those steps as I can make it: it just calls `register_entity` for all entities (step 1), calls `fetch_most_recent_snapshots` (step 2), and then calls `update_snapshots` (steps 3 & 4, which will be replaced by 2 function calls, one for each step)

recidiviz/persistence/database/update_historical_snapshots.py
zdg2102: Per same discussion as `map` above, list comprehension seems to be the preferred option here

recidiviz/persistence/database/update_historical_snapshots.py
zdg2102: It would be nice to make this file a little less long, but I'd rather keep this here since no other code should ever use it, so it doesn't make sense to me to make it a top-level entity under the `database` module

recidiviz/persistence/database/update_historical_snapshots.py
zdg2102: As discussed offline, since entities are typed here as `Any`, it wouldn't add any clarity

recidiviz/persistence/database/database.py
zdg2102: Per offline discussion, reordered this to make it clear it's the combination of merge+flush that is required

recidiviz/persistence/database/update_historical_snapshots.py
zdg2102: What's the benefit of using an `attrs` class?

recidiviz/persistence/database/database.py

recidiviz/persistence/database/update_historical_snapshots.py
zdg2102: No advantage. I ended up changing this to be `list(root_people)`, since to me that seems slightly more semantic, in that it's not exactly that I need a copy of `root_people`, but that I need a new list that should just start out with the same values as `root_people`

recidiviz/persistence/database/update_historical_snapshots.py
zdg2102: Done

recidiviz/persistence/database/update_historical_snapshots.py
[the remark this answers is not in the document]
zdg2102: Added a check that we haven't already set this field

recidiviz/persistence/database/database.py
```diff
@@ -141,16 +141,18 @@ def _query_people_and_open_bookings(session, region):
 
 
 def write_people(session, people):
-    """
-    Converts the given |people| into schema.Person objects and adds the
-    schema objects into the given |session|.
+    """Converts the given |people| into schema.Person objects and persists their
+    corresponding record trees.
 
     Args:
```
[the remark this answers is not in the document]
zdg2102: Done, and also removed Args/Returns from all docstrings

recidiviz/persistence/database/update_historical_snapshots.py
zdg2102: Done

recidiviz/persistence/database/database.py
```diff
@@ -162,86 +164,42 @@ def write_person(session, person):
...
+    root_people = [session.merge(root_person) for root_person in root_people]
 
     # Flush ensures all master entities, including newly created ones, have
     # primary keys set before creating historical snapshots
     session.flush()
 
-    logging.info('Flush complete for person: %s', person.person_id)
-
     # All historical snapshot changes should be given the same timestamp
     snapshot_time = datetime.now()
```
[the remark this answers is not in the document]
zdg2102: Yeah, you're right. I'd rather not change that in this PR, since it'll involve changing the callers as well, but I added a TODO to update it

recidiviz/persistence/database/update_historical_snapshots.py
jessex: Hm, interesting thought. I don't feel strongly about it in any event.

recidiviz/persistence/database/database.py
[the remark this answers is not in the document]
zdg2102: Filed #874

Issue #558, opened 2019-01-09
Update the create_scraper script to optionally accept a vendor


arian487:
It should have vendor specific templates for ease of use

jamwalla:
we have this now!

PR #803, opened 2019-02-01
Add --vendor option to create_scraper.py

KareemG:
Update for #558, although only for JailTracker (for now).


create_scraper.py
```diff
@@ -61,16 +71,24 @@ def create_scraper_files(subs):
     init_target = os.path.join(target_dir, '__init__.py')
     populate_file(init_template, init_target, subs)
 
-    scraper_template = os.path.join(template_dir, 'region_scraper.txt')
-    scraper_target = os.path.join(target_dir, subs['region'] + '_scraper.py')
-    populate_file(scraper_template, scraper_target, subs)
+    if vendor == 'jailtracker':
```
arian487: It looks like the only difference here is that you go into the jailtracker subdirectory, but otherwise the template names are the same.  In that case, you can set vendor to be '' by default and remove the if statement.  This will get large as we add more vendors so I think the much cleaner approach is to continue to expect the names to be the same:

os.path.join(template_dir, vendor, 'region_scraper.txt'

create_scraper.py
```diff
@@ -61,16 +71,24 @@ def create_scraper_files(subs):
...
+        scraper_template = os.path.join(template_dir, 'jailtracker', 'region_scraper.txt')
+    else:
+        scraper_template = os.path.join(template_dir, 'region_scraper.txt')
+    create_scraper(scraper_template)
 
-    yaml_template = os.path.join(template_dir, 'region.txt')
-    yaml_target = os.path.join(target_dir, subs['region'] + '.yaml')
-    populate_file(yaml_template, yaml_target, subs)
+    if vendor != 'jailtracker':
```
arian487: I would change this to
```suggestion
    if not vendor:
```

create_scraper.py
```diff
@@ -84,10 +102,15 @@ def create_test_files(subs):
     os.mkdir(target_test_dir)
 
     template_dir = os.path.join(ingest_dir, 'scraper_template')
-    test_template = os.path.join(template_dir, 'region_scraper_test.txt')
-    test_target = os.path.join(target_test_dir,
-                               subs['region'] + '_scraper_test.py')
-    populate_file(test_template, test_target, subs)
+
+    if vendor == 'jailtracker':
```
arian487: Same comment as above, you don't need to special case the if.

recidiviz/ingest/scraper_template/jailtracker/region_scraper_test.txt
```diff
@@ -0,0 +1,232 @@
+# Recidiviz - a platform for tracking granular recidivism metrics in real time
```
arian487: Can you use the base test model for the vendor scrapers?  We just recently decided to move to this model.  See:

https://github.com/Recidiviz/recidiviz-data/blob/master/recidiviz/tests/ingest/vendors/brooks_jeffrey/brooks_jeffrey_scraper_test.py

And then the children inherit from the vendor.  In most cases the test will just inherit and do nothing extra, but in some cases we might have extra logic we want to add a test for.  Similarly we want each child to have a test even if it is identical because if somneone changes the child scraper we want the test to complain.

arian487:
I've merged this, this will be a huge help to knocking out a TON of jailtracker scrapers :)

PR #807, opened 2019-02-01
refactored brooks jeffrey tests to test parse_total_bond directly instead of only relying on fixtures

terinpw:


recidiviz/ingest/vendors/brooks_jeffrey/brooks_jeffrey_scraper.py
```diff
@@ -143,4 +146,4 @@ def _split_charges(
 
 
 def _split(repeated_object_str: str) -> List[str]:
-    return [s for s in repeated_object_str.split('\n') if s]
+    return [s.strip() for s in repeated_object_str.split('\n') if s]
```
colincadams: not sure which you find more readable but I've been using `filter(None, map(str.strip, repeated_object_str.split('\n')))` for this patten. The advantage of the second is that if `s` is just a space it is still filtered out, whereas with the first one you'll end up with an empty string in the list.

Issue #798, opened 2019-01-31
Some Brooks Jeffrey Regions need special casing


terinpw:
So far:
us_ar_pope , us_al_pickens:
- Need to be able to parse bond amounts and type in same line (ex. "$5 CASH" or "$5 c/b")

us_ar_baxter, us_ar_columbia, us_ar_cross, us_ar_st_francis
- All data is in a weird div format that the data extractor can't automatically parse, but all of these websites share the same format.

terinpw:
baxter, columbia, cross, st francis, jefferson, and tom greene done with #878

PR #809, opened 2019-02-01
data extractor: convert nested key

jamwalla:
converts a key that is nested in several elements, e.g. `<foo><bar>key</bar></foo><baz>value</baz>`

#798


recidiviz/ingest/extractor/html_data_extractor.py
```diff
@@ -256,6 +256,22 @@ def _key_element_to_cell(self, key, key_element):
             if self._insert_cells_from_text(key, text, key_element):
                 return True
 
+        # <foo><bar>key</bar></foo><baz>value</baz>
+        # Finds the oldest ancestor with the same text as the key, then sets
+        # the key element and the ancestor's next element to td cells,
+        # if a next element with text exists.
+        parent = key_element.getparent()
+        grand_parent = parent.getparent() if parent else None
+        while grand_parent is not None and key_element.text == \
```
terinpw: nit, we should be able to get rid of the `is not None` for this and below -- and just rely on None evaluating to false.

recidiviz/ingest/extractor/html_data_extractor.py
```diff
@@ -256,6 +256,22 @@ def _key_element_to_cell(self, key, key_element):
...
+                grand_parent.text_content():
```
terinpw: Should we be using _get_text_from_element here too? and below

recidiviz/ingest/extractor/html_data_extractor.py
jamwalla: talked offline, I guess since these are iterable they evaluate to false when they have no children, so `is not None` is explicitly what we want

jamwalla:
thanks!

PR #811, opened 2019-02-01
Brooks jeffrey regions in IN, KS, and MS that don't require special casing

terinpw:


recidiviz/ingest/scraper_template/region_scraper_test.txt
```diff
@@ -21,7 +21,6 @@ from lxml import html
 
 from recidiviz.ingest import constants
 from recidiviz.ingest.models.ingest_info import IngestInfo
-from recidiviz.ingest.task_params import Task
```
jamwalla: hmm, we should keep this for tests that actually do stuff but this is something to consider for #803

recidiviz/ingest/__init__.py
```diff
@@ -56,10 +56,18 @@
...
+import recidiviz.ingest.us_ks_cherokee
+import recidiviz.ingest.us_ks_jefferson
+import recidiviz.ingest.us_ks_pratt
+import recidiviz.ingest.us_mo_stone
+import recidiviz.ingest.us_ms_clay
+import recidiviz.ingest.us_ms_kemper
+import recidiviz.ingest.us_ms_tunica
 import recidiviz.ingest.us_ms_desoto
 import recidiviz.ingest.us_mt_gallatin
-import recidiviz.ingest.us_mo_stone
```
jamwalla: oh no!

jamwalla: oh yes! it's alphabetized

recidiviz/ingest/scraper_template/region_scraper_test.txt
terinpw: Good catch -- reverted!

recidiviz/ingest/__init__.py
terinpw: I just alphabetized it! it's still there.

jessex:
Goodness, @terinpw

PR #813, opened 2019-02-01
Create script for generating enum migrations

zdg2102:


zdg2102:
@colincadams good idea. Created `recidiviz.tools` and moved there

jessex:
This is great, @zdg2102. Can we update the migration docs to describe how to use this?

zdg2102:
@jessex yep, I'll be updating the migration docs soon, since at this point we've pretty significantly changed the workflow, and I'll include this script

PR #814, opened 2019-02-01
MappableEnums ignore punctuation

jamwalla:
Curious what people's thoughts are - I made a couple decisions I'm not sure about, re removing punctuation from strings

- Decided to replace all punctuation with whitespace (rather than just strip punctuation from the ends)
- Decided to only do this for enums (rather than bake this into `normalize`)
- Had to special - case strings that are only punctuation (in_vigo maps `*` to `None`, but I wonder if we could have to map some symbol to an enum)

#733


jessex:
I like this, @jamwalla!

Responses to your questions:
* I like replacing punctuation throughout and not just at the ends! I think it makes for a cleaner experience for the scraper writer overall. But we should document this clearly, i.e. how to incorporate new values that include punctuation in their raw forms.
* If we baked it into normalize, what else would it impact besides enums? Any sort of string input? If so, I agree that it should focus on enums.
* I think that special case is so rare that your one line handling of it is a good solution.

And a question from me: are there tests we can add in `mappable_enums` to make sure that different raw forms of a string map to the same enum?

jamwalla:
@jessex, sorry, it took me a little bit to keep this updated. I added a test and some comments, can you take another look?

PR #820, opened 2019-02-02
Rewriting texas logic to handle arbitrary empty columns and rows, and also arbitrary header lengths

arian487:
Each PDF parsed slightly differently, and the last page on a multiple_table=True call also does not always work.  Probably the most annoying code I have ever written.  Thanks Texas.


arian487:
So it looks like crystl cty is not in the fips map, but is in the 1996 report, should we drop any rows that don't have a matching in the FIPS?

zacharylawrence:
> So it looks like crystl cty is not in the fips map, but is in the 1996 report, should we drop any rows that don't have a matching in the FIPS?

It looks like `crystl cty` may be: https://en.wikipedia.org/wiki/Crystal_City,_Texas, which lists the county as `Zavala` (does exist in `fips.csv`). In this case, I think it probably makes sense to special case it like in KY: [ky_aggregate_ingest.py#L184](https://github.com/Recidiviz/recidiviz-data/blob/728b647f0234f1f8e59b958f6f921f3e65a72a82/recidiviz/ingest/aggregate/regions/ky/ky_aggregate_ingest.py#L184)

recidiviz/ingest/aggregate/regions/tx/tx_aggregate_ingest.py
```diff
@@ -60,68 +64,100 @@ def _parse_date(filename: str) -> datetime.date:
...
         page = tabula.read_pdf(
             filename,
-            pages=page,
-            pandas_options={
-                'header': [1, 2, 3, 4, 5],
-            }
+            multiple_tables=True,
+            pages=page_num,
         )
+        page = page[-1]
```
zacharylawrence: When you index into the last table on each page, this seems like more of a subset of each page. Maybe consider something just like `df` (canonical DataFrame variable name)?

recidiviz/ingest/aggregate/regions/tx/tx_aggregate_ingest.py
```diff
@@ -60,68 +64,100 @@ def _parse_date(filename: str) -> datetime.date:
...
+    # concat because the column names might all be different.  Renaming them
+    # allows concat to pass happily.
+    for i, page in enumerate(pages):
+        # We don't care about % of capacity and total_other so we drop these
+        # columns.
+        row_len = len(page.iloc[0])
+        indices_to_drop = [row_len - 2, row_len - 6]
+        columns_to_drop = [page.columns[i] for i in indices_to_drop]
+        page.drop(columns_to_drop, axis='columns', inplace=True)
+        if report_date.year < 1996:
```
zacharylawrence: Thoughts on extracting this and the table definitions above into a seperate method?
```
columns_to_schema = _get_column_names(report_date)
```

recidiviz/ingest/aggregate/regions/tx/tx_aggregate_ingest.py
```diff
@@ -60,68 +64,100 @@ def _parse_date(filename: str) -> datetime.date:
...
             filename,
-            pages=page,
-            pandas_options={
-                'header': [1, 2, 3, 4, 5],
-            }
+            multiple_tables=True,
+            pages=page_num,
         )
+        page = page[-1]
+        page.dropna(axis='columns', inplace=True, thresh=5)
```
zacharylawrence: @rasmi Recommends against using `inplace=True` and instead just assigning to the same variable

recidiviz/ingest/aggregate/regions/tx/tx_aggregate_ingest.py
```diff
@@ -60,68 +64,100 @@ def _parse_date(filename: str) -> datetime.date:
             if DATE_PARSE_ANCHOR in line:
                 # The date is on the next line if anchor is present on the line
                 return dateparser.parse(lines[index+1]).date()
+        # If this doesn't work, try scraping it from the url name
+        if DATE_PARSE_ANCHOR_FILENAME in filename:
+            filename = filename.split('/')[-1]
```
zacharylawrence: Does this comment also apply here? https://github.com/Recidiviz/recidiviz-data/pull/791#discussion_r252890385

recidiviz/ingest/aggregate/regions/tx/tx_aggregate_ingest.py
```diff
@@ -60,68 +64,100 @@ def _parse_date(filename: str) -> datetime.date:
...
+    del columns_pre_1996['Conv. SJF Sentenced to State Jail']
+
+    columns_1996 = copy.deepcopy(columns_after_1996)
+    del columns_1996['Pretrial SJF']
+    del columns_1996['Conv. Sentenced to Co. Jail Time']
+    del columns_1996['Conv. SJF Sentenced to State Jail']
 
     pages = []
-    for page in range(1, num_pages):
+    for page_num in range(1, num_pages+1):
```
zacharylawrence: Should you just update `num_pages = 10` in this case?

recidiviz/ingest/aggregate/regions/tx/tx_aggregate_ingest.py
```diff
@@ -60,68 +64,100 @@ def _parse_date(filename: str) -> datetime.date:
...
         page = tabula.read_pdf(
             filename,
-            pages=page,
-            pandas_options={
-                'header': [1, 2, 3, 4, 5],
-            }
+            multiple_tables=True,
```
zacharylawrence: Could you add a super short comment about why you had to do this? Something along the lines of needing to parse the header as a separate table since it's unaligned with the data in special cases

recidiviz/ingest/aggregate/regions/tx/tx_aggregate_ingest.py
```diff
@@ -60,68 +64,100 @@ def _parse_date(filename: str) -> datetime.date:
...
+            pages=page_num,
         )
+        page = page[-1]
+        page.dropna(axis='columns', inplace=True, thresh=5)
+        # We want to remove all of the rows and columns that have no data.
+        indices_to_drop = []
+        for i in range(len(page.iloc[:, 0])):
```
zacharylawrence: To get the size of the DataFrame, you could just use `.shape` instead of needing to index

recidiviz/ingest/aggregate/regions/tx/tx_aggregate_ingest.py
```diff
@@ -60,68 +64,100 @@ def _parse_date(filename: str) -> datetime.date:
...
-            pages=page,
-            pandas_options={
-                'header': [1, 2, 3, 4, 5],
-            }
+            multiple_tables=True,
+            pages=page_num,
         )
+        page = page[-1]
+        page.dropna(axis='columns', inplace=True, thresh=5)
+        # We want to remove all of the rows and columns that have no data.
```
zacharylawrence: Why do we need to do this if we just called `dropna` above?

recidiviz/ingest/aggregate/regions/tx/tx_aggregate_ingest.py
```diff
@@ -60,68 +64,100 @@ def _parse_date(filename: str) -> datetime.date:
...
+        columns_to_drop = [page.columns[i] for i in indices_to_drop]
+        page.drop(columns_to_drop, axis='columns', inplace=True)
+        if report_date.year < 1996:
+            columns_to_schema = columns_pre_1996
+        elif report_date.year == 1996:
+            columns_to_schema = columns_1996
+        else:
+            columns_to_schema = columns_after_1996
+        page.columns = columns_to_schema.keys()
+        # We don't care about these two
```
zacharylawrence: Is this an old comment referencing the code above?

recidiviz/ingest/aggregate/regions/tx/tx_aggregate_ingest.py
```diff
@@ -60,68 +64,100 @@ def _parse_date(filename: str) -> datetime.date:
...
+    pages[-1].drop(pages[-1].tail(1).index, inplace=True)
+
+    # Build result for all the pages.  We rename the columns before calling
+    # concat because the column names might all be different.  Renaming them
+    # allows concat to pass happily.
+    for i, page in enumerate(pages):
+        # We don't care about % of capacity and total_other so we drop these
+        # columns.
+        row_len = len(page.iloc[0])
+        indices_to_drop = [row_len - 2, row_len - 6]
```
zacharylawrence: If we added these columns to `columns_after_1996` would it let us avoid needing to drop these columns manually and we could rely on `rename_columns_and_select` below? Something like:
```
columns_to_schema = .....

# Drop columns we never care about
del columns_to_schema['% of capacity']
del columns_to_schema['total_other SJF']
```

I'm not sure if this is idea though, thoughts?

recidiviz/ingest/aggregate/regions/tx/tx_aggregate_ingest.py
```diff
@@ -60,68 +64,100 @@ def _parse_date(filename: str) -> datetime.date:
             if DATE_PARSE_ANCHOR in line:
                 # The date is on the next line if anchor is present on the line
                 return dateparser.parse(lines[index+1]).date()
+        # If this doesn't work, try scraping it from the url name
+        if DATE_PARSE_ANCHOR_FILENAME in filename:
+            filename = filename.split('/')[-1]
+            date_str = filename.strip('.pdf').strip(DATE_PARSE_ANCHOR_FILENAME)
+            d = dateparser.parse(date_str).date()
+            return datetime.date(d.year, d.month, day=1)
```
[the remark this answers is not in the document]
arian487: Day=1.  The dateparser weirdly sets it to the second day of the month when parsing dates in this format...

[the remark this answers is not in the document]
zacharylawrence: Yeah, it uses the current day for values it can't parse............

recidiviz/ingest/aggregate/regions/tx/tx_aggregate_ingest.py
arian487: Done.

recidiviz/ingest/aggregate/regions/tx/tx_aggregate_ingest.py
arian487: Thats much cleaner.  Done.

recidiviz/ingest/aggregate/regions/tx/tx_aggregate_ingest.py
arian487: Ah very good catch.  This would have broken.  I realize now that there are a few formats you can upload it in.  I've changed this and made it more robust, and also changed a couple of the filenames to reflect that in the tests.

recidiviz/ingest/aggregate/regions/tx/tx_aggregate_ingest.py
arian487: Done.

recidiviz/ingest/aggregate/regions/tx/tx_aggregate_ingest.py
arian487: Because some rows can still have null values in them and worse, becausae we are assigning our own column headers we want to remove anything thats not data.  Sometimes the last table includes the headers...and sometimes it does not so the table might be:

county | column1 | column2
name1 | data        | data

Where none of the values are actually none, but its still somewhat nonsense (and variable data).  
The first dropna literally drops full columns because some of the PDFs parse entire null columns, and this second one basically walks until it catches an int.

recidiviz/ingest/aggregate/regions/tx/tx_aggregate_ingest.py
arian487: Yes.  Removed.

recidiviz/ingest/aggregate/regions/tx/tx_aggregate_ingest.py
arian487: discussed offline.  Done.

recidiviz/ingest/aggregate/regions/tx/tx_aggregate_ingest.py
arian487: Done.

recidiviz/ingest/aggregate/regions/tx/tx_aggregate_ingest.py
arian487: I don't think so, it can be misleading because the number of pages is in fact 9, not 10.

recidiviz/ingest/aggregate/regions/tx/tx_aggregate_ingest.py
```diff
@@ -60,75 +63,121 @@ def _parse_date(filename: str) -> datetime.date:
...
+        filename_date = filename.lower()
+        if DATE_PARSE_ANCHOR_FILENAME in filename_date:
+            # The names can be a few formats, the most robust way is to take
+            # all of the text after the anchor.
+            # (eg. report Jan 2017.pdf)
+            start = filename_date.index(DATE_PARSE_ANCHOR_FILENAME)\
+                    + len(DATE_PARSE_ANCHOR_FILENAME)
+            date_str = filename_date[start:].strip('.pdf')
+            d = dateparser.parse(date_str).date()
+            return datetime.date(d.year, d.month, day=1)
```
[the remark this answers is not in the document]
arian487: Done.

recidiviz/ingest/aggregate/regions/tx/tx_aggregate_ingest.py
```diff
@@ -60,75 +63,121 @@ def _parse_date(filename: str) -> datetime.date:
...
+                int(cell_val)
+                break
+            except ValueError:
+                indices_to_drop.append(i)
+        df = df.drop(indices_to_drop)
+        pages.append(df)
 
     # Drop last rows since it's the 'Totals' section
-    last_page = last_page.drop(last_page.tail(1).index)
+    pages[-1].drop(pages[-1].tail(1).index, inplace=True)
```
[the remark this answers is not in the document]
arian487: Done.

PR #825, opened 2019-02-04
Lumpkin ga and modify zuercher

colincadams:


recidiviz/ingest/vendors/zuercher/zuercher_scraper.py
```diff
@@ -180,140 +193,97 @@ def _postprocess_charge(self, booking: Booking, charge: Charge):
...
         # Split the charge on ';', strip any whitespace, and filter out any
         # empty strings.
-        charge_kvs = list(filter(None, map(str.strip, charge_text.split(';'))))
+        first_kv, *later_kvs = filter(None, map(str.strip,
+                                                charge_text.split(';')))
 
         # Sometimes a ';' ends up in the charge description. To handle this we
         # iterate backwards and if we don't recognize information we keep it
         # around and stick it in the charge description.
-        suffix = ''
```
colincadams: this diff looks huge, but I just restructured the for loop so that we only loop over the the second key and onward and skip the first key. Then we parse the first key separately. This was necessary as for lumpkin there are some charges that aren't prefixed with a key.

PR #826, opened 2019-02-04
add inferred drop to charge status schema list

terinpw:
@zdg2102, what's the best way to tell if a migration for this value has already happened, even though we're never setting it?


zdg2102:
`SELECT enum_range(NULL::charge_status)` against prod. It hasn't been migrated

terinpw:
@colincadams Done and migrated!

PR #827, opened 2019-02-04
handle NY special case for site scraper

arian487:
This PR special cases NY so we just always download the one pdf and adds the site scraper (which is just a static function).


recidiviz/ingest/aggregate/scrape_aggregate_reports.py
```diff
@@ -90,13 +93,14 @@ def scrape_aggregate_reports():
 
 
 def _get_file_to_upload(
-        path: str, fs: gcsfs.GCSFileSystem, url: str, pdf_name: str) \
+        path: str, fs: gcsfs.GCSFileSystem, url: str, pdf_name: str,
+        is_ny: bool) \
```
jessex: What do you think about generalizing from `is_ny` to something like `always_download`? I wouldn't want to have to rename this to `is_ny_or_some_other_state` in the future.

arian487: Done.

PR #832, opened 2019-02-04
Refactor Superion test to be a base test, add Guilford explicitly.

ohinds:
Follows the model of JailTracker scraper test.


jamwalla:
#675

Issue #652, opened 2019-01-17
Refactor `Ingest` module to support scraper/direct/aggregate forms of ingest


zacharylawrence:
For a more detailed description, see: https://github.com/Recidiviz/recidiviz-data/pull/648#discussion_r248542873

jamwalla:
done!

PR #834, opened 2019-02-04
regions refactor

jamwalla:
Moves all scrapers into `recidiviz/ingest/scrape/regions`. closes #231.


jessex:
Thank you so so much for this, @jamwalla! My feeble attempt that grew quickly stale was burning a whole in my todo list.

Hate to do this to you, but I doubt you've had a chance to see #652. It's a larger scale repackaging of the ingest package, and it would move all of the scraper-related code (i.e. not direct ingest, not aggregate ingest, and not ingest-general) into `recidiviz.ingest.scraper` and in turn would move all of the regions you've already moved here into `recidiviz.ingest.scraper.regions` instead of `recidiviz.ingest.regions`.

I have two suggestions:
- We can merge this PR as-is, because it gets us closer to the organization laid out in #652 and it will be easyish to just move this entire regions package all at once later.
- We can tackle #652 now, or at least the `scraper` portion of it, and move everything you have here one more time.

Whatever you would rather do right here right now is fine by me.

jamwalla:
@jessex I'm happy to do that in this change! @arian487 had a question about direct ingest - I think we wanted to have each subdirectory of `regions` be *either* a scraper or a direct ingest "scraper" (parser, etc). But scraper implies we're scraping a website. How do you see direct ingest fitting into this naming schema?

A smaller issue - I don't think python can handle both `recidiviz/ingest/scraper/` and `recidiviz/ingest/scraper.py`, but how about `recidiviz/ingest/scrapers`?

jessex:
[In my comment](https://github.com/Recidiviz/recidiviz-data/pull/648#discussion_r248542873) that spawned the issue, I had foreseen making a specific `recidiviz.ingest.direct` sub-package with the code, APIs, and docs for handling direct integration. Same goes with `recidiviz.ingest.aggregate`. So I invert the model a bit: instead of `recidiviz.ingest.regions` which includes all regions across all ingest channels, we'd have `recidiviz.ingest.[channel].regions`. The benefits of my approach are that you can easily see at a glance the set of regions available in each distinct channel, and each region is in the same parent package as the code that it relies on, which makes it easier to decompose the repository in the future, bundle/deploy different portions separately, etc.

Good call out on the naming conflict. If the other packages are `recidiviz.ingest.direct` and `recidiviz.ingest.aggregate` then maybe this one could be `recidiviz.ingest.scrape` or `recidiviz.ingest.scraping`.

PR #837, opened 2019-02-05
Fix texas by handling concatenated columns

arian487:
Some columns are concatenated together and are separated by a space.  We look for such columns, break them up and index them into the right spot.


recidiviz/ingest/aggregate/regions/tx/tx_aggregate_ingest.py
```diff
@@ -100,6 +100,18 @@ def _parse_table(filename: str, report_date: datetime.date) -> pd.DataFrame:
         numeric_elements = df.apply(pd.to_numeric, errors='coerce').notnull()
         rows_containing_data = numeric_elements.any(axis='columns')
         df = df.loc[rows_containing_data]
+        # Next we finally break up some of the columns that were incorrectly
+        # concatenated.
+        for column in df.columns[1:]:
+            # By this point we should only have numeric data in the rows,
+            # if this happens it means some columns were concatenated and they
+            # must be split
+            if ' ' in df[column].iloc[0]:
```
zacharylawrence: Instead of just checking row 0, you could check everything in the column with something like: `all(df[column].str.contains(' '))`

Thoughts on that?

PR #842, opened 2019-02-05
tool for pulling enum errors from logs

colincadams:


recidiviz/tools/read_enum_errors.py
```diff
@@ -0,0 +1,60 @@
...
+$ python -m recidiviz.tools.read_enum_errors --file FILE [--include-request]
+"""
+import argparse
+import ast
+import json
+import pprint
+from typing import Any, Dict, Tuple
+
+
+def enum_errors_from_logs(filename: str) -> Dict[Tuple[str, str, str], Any]:
```
arian487: Can you add comments for parts of this file?  Would help with the readability IMO.

recidiviz/tools/read_enum_errors.py
```diff
@@ -0,0 +1,60 @@
...
+        for log in logs:
+            lines = log['jsonPayload']['message'].splitlines()
+            enum_type, enum_string, region = [None] * 3
+            line_iter = iter(lines)
+            while True:
+                line = next(line_iter)
+                if line.startswith('recidiviz.common.constants.mappable_enum.'
+                                   'EnumParsingError'):
+                    _, text = map(str.strip, line.split(':', 1))
+                    assert text.startswith('Could not parse')
```
arian487: Do we want to make this assertion?  This will break if we change the log message for mappable enums.  I'm not sure what value there is in asserting the log format so long as its an enum parsing error.  Similarly i don't think we need to check the line starts with the module path, couldn't we just check that 'EnumParsingError is in the line?

recidiviz/tools/read_enum_errors.py
colincadams: Added comments and pulled out some sub methods. Hopefully that helps.

recidiviz/tools/read_enum_errors.py
colincadams: I think it's beneficial to assert this since we rely on this assertion when trimming the string on the next line. If we change the log format then this script will require updating and I'd rather fail loudly when possible.

We could just check that `EnumParsingError` is in the line, but it didn't seem harmful to be more strict here and should be more performant. If you prefer the readability of not doing it though I can change it.

PR #844, opened 2019-02-05
change date parsing to just use the filename

arian487:
In another PR, we'll have some backup date standard that we try if all else fails.


recidiviz/ingest/aggregate/regions/hi/hi_aggregate_ingest.py
```diff
@@ -98,20 +99,9 @@ def parse(filename: str) -> Dict[DeclarativeMeta, pd.DataFrame]:
...
-            # all of the text after the anchor.
-            # (eg. report Jan 2017.pdf)
-            start = filename_date.index(DATE_PARSE_ANCHOR_FILENAME) \
-                    + len(DATE_PARSE_ANCHOR_FILENAME)
-            date_str = filename_date[start:].strip('.pdf')
-    except Exception as e:
-        raise AggregateDateParsingError(str(e))
-    return dateparser.parse(date_str).date()
+    end = filename.index('.pdf')
+    start = end - 10
```
zacharylawrence: I don't think you can always subtract `10` here. It looks like some filenames look like: 
```
_wp-content_uploads_2018_12_pop-reports-eom-2018-11-30.pdf
```

and other like: 
```
_wp-content_uploads_2017_10_pop-reports-eom-2017-09-30-17.pdf
```

PR #852, opened 2019-02-05
move base scraper files into recidiviz/ingest/scrape

jamwalla:
this moves files called `scraper`* into `scrape/`, but @colincadams and I discussed moving everything ingest-related (except aggregate and direct work) into that directory. We'll do that in a follow up.


jamwalla:
oh, duh! My bad, I'm uploading the rest of this refactor now and I've included the test files as well.

PR #854, opened 2019-02-05
Remove `task_age_limit` from queue.yaml (and other small fixes)

colincadams:
It turns out that if both `task_age_limit` and `task_retry_limit`
are specified, it will retry the task until *both* are met. This
means that tasks are retried for up to 3 days instead of just 5
times.


colincadams:
Relevant doc: https://cloud.google.com/appengine/docs/standard/php/config/queueref

arian487:
This is an amazing find...but also why was this buried so deeply!?

colincadams:
Thanks! Super unintuitive behavior from Cloud Tasks

PR #855, opened 2019-02-05
HTML data extractor updates

terinpw:
- Automatically remove comments before parsing
- Remove nested key case (reverting #809)


recidiviz/ingest/extractor/html_data_extractor.py
```diff
@@ -252,8 +260,8 @@ def _key_element_to_cell(self, key: str, key_element: HtmlElement) -> bool:
 
         # <foo>key : value</foo>
         # Create new td elements for the key and the value and insert them.
-        text = self._get_text_from_element(key_element).strip()
-        if text.startswith(key):
+        text = self._get_text_from_element(key_element)
+        if text and text.startswith(key):
             if self._insert_cells_from_text(key, text, key_element):
                 return True
```
jamwalla: can you remove the block below here (<foo><bar>key</bar></foo><baz>value</baz>), since we removed the test, and mention that we're reverting #809?

terinpw: Yup, meant to do that, done now.

PR #866, opened 2019-02-06
finish refactor of scraper files into recidiviz/ingest/scrape

jamwalla:
finishes work from #834 #852


jessex:
That's an excellent question, @colincadams. I think there's a chance it would be used in direct ingest, but until we have our first such integration up and running, it seems a bit premature to pull that thread out of the scrape package. Sound fair?

PR #869, opened 2019-02-06
Fix broken link to `ingest/scrape/vendors`

ajsecord:


jamwalla:
I think this is also unnecessary from #897? I'm going to close it but re-open if I'm missing something. Thanks for this!

PR #872, opened 2019-02-06
remove unused packages from Pipfile

jamwalla:
removes the following packages from Pipfile:
```
altgraph
asn1crypto
astroid
backports.functools_lru_cache
backports_abc
bdist_mpkg
beautifulsoup4
certifi
cffi
chardet
click
configparser
cookies
coverage
cryptography
docopt
enum34
funcsigs
google-api-core
idna
ipaddress
isort
itsdangerous
jellyfish
jinja2
lazy-object-proxy
macholib
markupsafe
mccabe
modulegraph
nose
pbr
pluggy
protobuf
psycopg2
py
py2app
pyOpenSSL
pycparser
pymysql
pyparsing
pytest-mock
regex
responses
rope
singledispatch
six
slack-cleaner
slacker
tornado
tzlocal
urllib3
virtualenv
waitress
WebOb
werkzeug
xattr
zope.interface
```

close #872


arian487:
We need to keep protobuf and psycopg2.  Protobuf is imported directly from the generated proto file and psycopg2 is required as an engine backend by sqlalchemy.  It looks like its removed from the lock file.  I think google-api-core is also likely needed but I'm not sure.  

When submitting this, I would deploy this to staging and run a test region just to be sure nothing breaks before deploying to prof.

zacharylawrence:
> When submitting this, I would deploy this to staging and run a test region just to be sure nothing breaks before deploying to prof.

:+1:

PR #877, opened 2019-02-06
7 Superion counties in NC

ohinds:
NC Superion scrapers that didn't need too much special-casing.


recidiviz/ingest/scrape/regions/us_nc_wake/us_nc_wake_scraper.py
```diff
@@ -0,0 +1,63 @@
...
+    def __init__(self):
+        super(UsNcWakeScraper, self).__init__('us_nc_wake')
+
+    def populate_data(self, content, task: Task,
+                      ingest_info: IngestInfo) -> Optional[ScrapedData]:
+        scraped_data: Optional[ScrapedData] = super(
+            UsNcWakeScraper, self).populate_data(content, task, ingest_info)
+
+        person = scraper_utils.one('person', scraped_data.ingest_info) \
+            if scraped_data else ingest_info
```
terinpw: Will this work w the code below? Doesn't `one` return the Person and `ingest_info` just return the IngestInfo?

Maybe this will work b/c in the second case, person.bookings will be none... but either way, shouldn't we just return if None in this case?

recidiviz/ingest/scrape/regions/us_nc_cabarrus/us_nc_cabarrus_scraper.py
```diff
@@ -0,0 +1,48 @@
...
+    """Scraper implementation for us_nc_cabarrus."""
+    def __init__(self):
+        super(UsNcCabarrusScraper, self).__init__('us_nc_cabarrus')
+
+    def populate_data(self, content, task: Task,
+                      ingest_info: IngestInfo) -> Optional[ScrapedData]:
+        scraped_data = super(UsNcCabarrusScraper, self).populate_data(
+            content, task, ingest_info)
+
+        ingest_info = scraped_data.ingest_info if scraped_data else ingest_info
```
terinpw: Can we add a comment as to why this is necessary?

recidiviz/tests/ingest/scrape/regions/us_nc_wake/us_nc_wake_scraper_test.py
```diff
@@ -0,0 +1,72 @@
...
+
+_ENUM_OVERRIDE_HTML = html.fromstring(
+    fixtures.as_string('us_nc_wake', 'enum_override.html'))
+
+
+class TestUsNcWakeScraper(SuperionScraperTest, unittest.TestCase):
+    """Scraper tests for us_nc_wake."""
+    def _get_scraper(self):
+        return UsNcWakeScraper()
+
```
terinpw: For this scraper - shouldn't we also test the region specific populate data method? Or is that tested in the test below? If so, can we update the name to be more descriptive. 

Same for any other regions with specific populate_data methods.

recidiviz/common/constants/charge.py
```diff
@@ -111,22 +111,37 @@ def _get_default_map():
 }
 
 _CHARGE_STATUS_MAP = {
```
jessex: Thanks for this reorganization!

recidiviz/ingest/scrape/regions/us_nc_wake/us_nc_wake_scraper.py
ohinds: Ah yes, good point. I'll make it a blank person instead

recidiviz/ingest/scraper_template/superion/region_scraper.txt
```diff
@@ -0,0 +1,25 @@
+# Recidiviz - a platform for tracking granular recidivism metrics in real time
```
jessex: `recidiviz/tools/scraper_template/README.md` currently has a list of supported vendors for the generator that should be updated to include Superion, or generalized so it doesn't need to be updated with each new template.

recidiviz/ingest/scrape/regions/us_nc_cabarrus/us_nc_cabarrus_scraper.py
ohinds: Ugh, it's not. Just cruft.

recidiviz/tests/ingest/scrape/regions/us_nc_wake/us_nc_wake_scraper_test.py
ohinds: Done.

recidiviz/ingest/scraper_template/superion/region_scraper.txt
ohinds: Good catch! Done.

PR #882, opened 2019-02-06
Create migration to update aggregate table unique constraints

zdg2102:


jessex:
General question not related to this particular migration: does the filename of the migration matter to Alembic operation? And more specifically, is it possible to set a template for those filenames? I'd love if they were, for example, prepended with the date so that they'd appear in the repo and on the filesystem in order. I think that provides a nice readability improvement, to see how the schema evolved over time without going through a commit log.

zdg2102:
@jessex yeah, that would be nice. Filed #906 against myself to look into it

PR #889, opened 2019-02-07
create_scraper update ingest/__init__.py

jamwalla:


recidiviz/tools/create_scraper.py
```diff
@@ -86,17 +91,50 @@ def create_yaml(template):
...
 
+def _rewrite_init_file(filename, import_statement):
+    """rewrites recidiviz/ingest/__init__.py to include the new import
+    statement."""
+    gpl = []
+    docstring = []
+    imports = [import_statement + '\n']
+    with open(filename) as f:
+        stages = ('LICENSE', 'DOCSTRING', 'IMPORTS')
+        stage = 0
```
terinpw: Do we think it'd be clearer if we just use the stage names directly? Instead of using the indices and the list?

PR #897, opened 2019-02-07
Fix old path in docs.

ajsecord:


colincadams:
Actually we just removed this section of the doc, so this change is no longer necessary. Thanks for sending the PR though!

Issue #375, opened 2018-12-21
Disambiguate if a scraped entity matches to >1 database entity


jamwalla:
Entity matching is currently implemented with a single notion of equality (e.g. `is_person_match`). If there are multiple matches in the database for a single scraped entity, we should attempt to disambiguate if possible. This can happen within an entity and within child entities:
- If we match with two people, for example, see if we can disambiguate from race or gender
- If we match with two people, for example, see if we can disambiguate from entire charges (using basic equality)

PR #904, opened 2019-02-07
entity match people on booking ids

### LINKED ISSUE #863: Entity Matching: matched more than one database entity
https://pantheon.corp.google.com/errors/CPeD-a6h3s_fkAE?time=P1D&project=recidiviz-123

Should look into whether all of these instances are actually manifestations of #375, or if instead a person was accidentally written twice for some reason so the two database entities should really be a single entity.

jamwalla:
#375 closes #863


jamwalla:
thanks Josh!!

Issue #836, opened 2019-02-04
Cron updates timeout if too many cron jobs


arian487:
We need to stat thinking of having one cron while calls scraper/start on all regions, or figure out why we timeout...

rasmi:
Looks like we can use [Cloud Scheduler](https://cloud.google.com/scheduler/) which supports [App Engine endpoints](https://cloud.google.com/scheduler/docs/creating) for cron operations. It's $0.10/job/month (where a job is one cron entry, not one execution), so depending on how we split regions up, it could be quite pricey or fairly cheap. It may make more sense to stick to App Engine cron if the recently reworked region grouping is sustainable.

@colincadams @jessex

jessex:
Thanks for the lookout, @rasmi! Cloud Scheduler looks very slick. I think if the reworked region grouping is sustainable then I agree that there's not a huge incentive to switch at the moment. 

However, I also think that if it's not sustainable then going out to per-region crons on Cloud Scheduler might eventually prove too expensive. If we eventually have, say, 1000 distinct counties that we have scheduled starts and stops for (and trace amounts of other crons for aggregate and inferred release and so forth), then we're looking at $200 per month for scheduling. That might be okay, but at 2500 counties and $500 per month for scheduling it's likely not worth it.

PR #905, opened 2019-02-07
add timezone arg to scraper stop and start

arian487:
The motivation here is so we don't timeout when trying to upload cron.yaml.  We move the starts into threads so the actual call to scraper/start does not timeout.  This is a temporary solution while we work on cloud composer implementation.

#836


recidiviz/utils/regions.py
```diff
@@ -130,6 +131,11 @@ def get_supported_region_codes(full_manifest=False):
     manifest = get_full_manifest()
     supported_regions = manifest["regions"]
 
+    if timezone:
+        supported_regions = {region: info for region, info in
```
colincadams: what if here we compare the utc offsets of the time zones, so that if they specify a different city name than the canonical one but it is in the same timezone we still start it? You can get the utc offset like this:
```
tz = pytz.timezone('America/Los_Angeles')
tz.utcoffset(datetime.now(tz))
```

arian487: Discussed offline, lets leave this as is and add them later if it is needed.

PR #931, opened 2019-02-08
Add new jailtracker yaml fields and update global map

arian487:


recidiviz/ingest/scrape/vendors/jailtracker/jailtracker.yaml
```diff
@@ -20,6 +25,9 @@ key_mappings:
   charges.Counts: charge.number_of_counts
   charges.CrimeType: charge.charge_class
   charges.ChargeDescription: charge.name
+  charges.ArrestCode: charge.statute
 
   charges.BondAmount: bond.amount
   charges.BondType: bond.bond_type
+  # TODO Add bondsmen to jailtracker.
```
jessex: Is this going to be done in another issue? If so, let's add the issue number.

arian487: Done.

PR #932, opened 2019-02-08
Compare timezones based on utc offsets

colincadams:


colincadams:
Thanks for the reviews!

PR #935, opened 2019-02-08
Delete persistence/actions.py

zdg2102:


recidiviz/persistence/actions.py
```diff
@@ -1,42 +0,0 @@
...
-from recidiviz.common.ingest_metadata import IngestMetadata
-from recidiviz.persistence import persistence
-from recidiviz.utils.auth import authenticate_request
-
-actions = Blueprint('actions', __name__)
-
-
-@actions.route("/v1/records", methods=['POST'])
-@authenticate_request
-def write_record():
```
jessex: Is the idea here that we are not pursuing direct ingest for the moment or that we'll be redoing this endpoint?

zdg2102:
@jessex my understanding of @ohinds's design for direct ingest was that the files get uploaded to some separate storage system like google drive, and we have a cron that pulls any new files from there periodically. If that's the case, we don't need to support a direct HTTP endpoint for the upload, and I thought leaving it was confusing because it calls `persistence.write` but all the values are dummied out. Is there some part of the new direct ingest design that will still need the endpoint?

jessex:
Yeah I know his design called for the same serverless approach as in aggregate ingest but I didn't know if we were still planning to stand up an API as one of the channels for direct ingest from county-level agencies, or if that was distinctly out of Census scope. It sounds like it is, and that the only focus for now is on file-based uploads, which is reasonable.

zacharylawrence:
@zdg2102 @jessex I think we still need this API endpoint for direct ingest. My understanding is that the direct ingest flow looks like: `Google Drive -> GSC -> Cloud Function -> This API Endpoint -> Database`. 

We need this API because the Direct Ingest Cloud Function (not yet written - but will likely look like the Aggregate Ingest Cloud Function) can't directly talk to the Database. Instead, the Cloud Function needs to make an HTTP request to the monolith to ultimately persist the data.

PR #890, opened 2019-02-07
Adding us-in-boone scraper based on JailTracker

ajsecord:
Tested with `python -m recidiviz.tools.run_scraper --region us_in_boone --num_tasks 30 --lifo`.

PR #939, opened 2019-02-08
Adding global mappings found in us-in-boone data.

ajsecord:


ajsecord:
Blocking #890

recidiviz/common/constants/charge.py
```diff
@@ -152,6 +152,7 @@ def _get_default_map():
     'SENTENCED': ChargeStatus.SENTENCED,
     'SUPERVISED PROBATION': ChargeStatus.SENTENCED,
     'UNDER SENTENCE': ChargeStatus.SENTENCED,
+    'WARRANT SERVED': ChargeStatus.PENDING,
```
ohinds: I've been mapping these to 'PRETRIAL', because if there is a warrant, that means the charges have been filed.

ajsecord: Done, thanks! That's very helpful.

PR #951, opened 2019-02-11
Ca site aggregration

arian487:
Adds the code to handle california for state aggregation


recidiviz/ingest/aggregate/regions/ca/ca_aggregate_site_scraper.py
```diff
@@ -0,0 +1,88 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+# =============================================================================
+
+"""Scrapes the california aggregate site and finds pdfs to download."""
```
jessex: Super nit to the rescue: capitalize state names in comments and docs and so forth.

recidiviz/ingest/aggregate/regions/ca/ca_aggregate_site_scraper.py
```diff
@@ -0,0 +1,88 @@
...
+
+def _get_pdf_data(year, month_from, month_to, reporting_range):
+    return {
+        'DataType': 'Facility',
+        'ReportingRange': reporting_range,
+        'year': year,
+        'Month_From': month_from,
+        'Month_To': month_to,
+        'Sort_By': 'Date',
+        'jurisdictions': ['6', '21', '25', '28', '500', '36', '44', '50',
```
jessex: Can we comment on why this giant set of magical numbers?

recidiviz/ingest/aggregate/regions/ca/ca_aggregate_site_scraper.py
```diff
@@ -0,0 +1,88 @@
...
+        'RunQuery': 'Excel'
+    }
+
+
+def get_urls_to_download() -> List[Tuple[str, Dict]]:
+    """Get all of the urls that should be downloaded."""
+    page = requests.post(LANDING_PAGE, data=_get_landing_data()).text
+
+    # Formatting on the page is extremely weird so its easiest to just take a
+    # a slice of the data.
```
jessex: Duplicate "a"

recidiviz/ingest/aggregate/scrape_aggregate_reports.py
```diff
@@ -64,11 +65,13 @@ def scrape_aggregate_reports():
         'kentucky': ky_aggregate_site_scraper.get_urls_to_download,
         'new_york': ny_aggregate_site_scraper.get_urls_to_download,
         'texas': tx_aggregate_site_scraper.get_urls_to_download,
+        'california': ca_aggregate_site_scraper.get_urls_to_download,
```
jessex: Insert in alphabetical order.

recidiviz/ingest/aggregate/scrape_aggregate_reports.py
```diff
@@ -64,11 +65,13 @@ def scrape_aggregate_reports():
...
     }
     state = get_value('state', request.args)
     # We want to always download the pdf if it is NY because they always have
     # the same name.
     always_download = (state == 'new_york')
+    is_ca = (state == 'california')
```
jessex: Similar to my comment in the previous PR, I would consider generalizing this name to something like `append_year`

recidiviz/ingest/aggregate/scrape_aggregate_reports.py
```diff
@@ -77,10 +80,19 @@ def scrape_aggregate_reports():
     logging.info('Scraping all pdfs for %s', state)
 
     for url in urls:
-        pdf_name = urlparse(url).path.replace('/', '_').lower()
+        post_data = None
+        if isinstance(url, Tuple):
```
jessex: Hm, I know that we're not _really_ following a clean interface here and that each site scraper is just defining a method that to this point always had the same signature, but we're basically _trying_ to create a single interface and now we've broken that by having California return a different type, which causes us to do this type check, which feels non-ideal. Did you give any thought to having everyone return a structure that lists urls+payloads, or even going up a level of abstraction and having everyone return a list of request objects with which they can set whatever they like? This might be such a small use case that it doesn't warrant over-engineering, admittedly.

recidiviz/ingest/aggregate/regions/ca/ca_aggregate_site_scraper.py
arian487: Done.

recidiviz/ingest/aggregate/regions/ca/ca_aggregate_site_scraper.py
arian487: Added comment.

recidiviz/ingest/aggregate/regions/ca/ca_aggregate_site_scraper.py
arian487: Done.

recidiviz/ingest/aggregate/scrape_aggregate_reports.py
arian487: Done.

recidiviz/ingest/aggregate/scrape_aggregate_reports.py
arian487: So I think this is very specific logic to CA since we pull out the filename append from the post data.  In the case of CA the urls will always be the same but we still need consistent name uniqueness for each report (which is 'year' in the case of CA).  This is why I chose to make this is_ca, for other states the append might be different or the way to get append_year might be different.

recidiviz/ingest/aggregate/scrape_aggregate_reports.py
arian487: I thoiught of this and decided against it for your exact reason.  There are a total of 8 states that do aggregate reports and I don't see them growing any time soon.  I don't mind this type of minor casing for code that might never change again (or change in 5 years).

PR #952, opened 2019-02-11
Add crons for the state aggregates

arian487:


cron.yaml
```diff
@@ -75,6 +75,60 @@ cron:
...
+  - description: Run kentucky aggregate report
+    url: /scrape_aggregate_reports/scrape_state?state=kentucky
+    schedule: every day 09:00
+    timezone: America/New_York
+    retry_parameters:
+      min_backoff_seconds: 2.5
+      max_doublings: 5
+      job_age_limit: 9h
+
+  - description: Run new_york aggregate report
```
jessex: Minor nits, but we don't need to underscore New York here, and we should capitalize all state names in human-readable text

Issue #944, opened 2019-02-11
Add bondsmen to jailtracker and cases information to jailtracker


arian487:

jamwalla:
see #1084 because the sentence field here is a length of format `0y 0m 0d`, which dateparser doesn't support

PR #953, opened 2019-02-11
Add support to Jailtracker for handling cases repsonse

terinpw:
#944


recidiviz/ingest/scrape/vendors/jailtracker/jailtracker_scraper.py
```diff
@@ -205,15 +208,17 @@ def populate_data(
...
+                                                               ingest_info)
+        self._add_cases_information(ingest_info, task.custom[self._CASES])
+        self._sanitize_bond_types(ingest_info)
+        ingest_info = ingest_info.prune()
+        return ScrapedData(ingest_info=ingest_info, persist=True)
 
     def get_enum_overrides(self):
         return {
             'O': ChargeClass.PROBATION_VIOLATION,
+            'CLOSED': ChargeStatus.DROPPED,
```
arian487: Can't this be put into the global map for a charge status?

recidiviz/ingest/scrape/vendors/jailtracker/jailtracker_scraper.py
```diff
@@ -442,6 +445,69 @@ def is_parole_agency(agency):
...
+            if charge_status and not charge.status:
+                charge.status = charge_status
+            if bond:
+                charge.bond = bond
+            if sentence:
+                charge.sentence = sentence
+
+    def _any_charge_has_bond_amount(self, charges):
+        for charge in charges:
+            if charge.bond and charge.bond.amount and charge.bond.amount != '0':
```
arian487: does this break if the bond amount is '0.0'?

recidiviz/ingest/scrape/vendors/jailtracker/jailtracker_scraper.py
```diff
@@ -442,6 +445,69 @@ def is_parole_agency(agency):
...
+            charge_status = None
+
+            # Only use case-level bond information if bond amount not on charges
+            # already.
+            if not self._any_charge_has_bond_amount(charges):
+                bond = Bond(amount=str(case.get('BondAmount', '')),
+                            bond_type=case.get('BondType', ''))
+
+            if 'Sentence' in case:
+                length = case.get('Sentence', '')
```
ohinds: Do you need the second argument, given that you checked that 'Sentence' is a key just above?

recidiviz/ingest/scrape/vendors/jailtracker/jailtracker_scraper.py
```diff
@@ -442,6 +445,69 @@ def is_parole_agency(agency):
...
+                sentence = Sentence(min_length=length, max_length=length)
+
+            if 'Status' in case:
+                charge_status = case.get('Status', '')
+
+            self._add_to_charges(bond, sentence, charge_status, charges)
```
ohinds: Ditto here.

recidiviz/ingest/scrape/vendors/jailtracker/jailtracker_scraper.py
ohinds: I think we shouldn't because this might mean COMPLETED_SENTENCE or just SENTENCED in other places. We determined based on context that this meant DROPPED here.

arian487: Fair point.

recidiviz/ingest/scrape/vendors/jailtracker/jailtracker_scraper.py
terinpw: Good point, removing it now

recidiviz/ingest/scrape/vendors/jailtracker/jailtracker_scraper.py
terinpw: It would, I updated it so that it could handle this case.

Issue #879, opened 2019-02-06
It should be easy to add to the vendor scraper enum overrides in the child scraper 


ohinds:
Currently, a child scraper would have to manually call `super(...).get_enum_overrides` and update the returned dict with new values. It would be better to have a way to add to enum overrides so the vendor overrides were respected automatically.

jamwalla:
something to figure out for #816

Issue #816, opened 2019-02-01
restructure enum overrides


jamwalla:
2 issues came up with our current enum overrides setup: it would be nice to specify that the same string maps to an enum for one field but `None` for another field; it would be nice to specify that the same string maps to two different enum values. We rely too much on the type of enum value for this process.

Instead, we want a nested object, like
```python
{Race: {'A': Race.ASIAN}, ChargeClass: {'A': None}}
```

In fact, we'll make it an object.
```python
overrides = EnumOverrides.Builder()
overrides.add('A', Race.ASIAN)
overrides.ignore(ChargeClass, 'A')
return overrides.build()
```

jessex:
I like the builder option! We can hammer this out in code review of course, but I think builders are best when they ready fluidly, e.g.
```
return EnumOverrides.Builder()
        .add('A', Race.Asian)
        .ignore(ChargeClass, 'A')
        .build()
```

colincadams:
@jessex I agree I like when each method from the builder returns itself so that you can chain them nicely, but during our group discussion I think @zacharylawrence  mentioned that this is atypical in python?

jessex:
It's true that the builder pattern is a bit atypical, largely because you have named parameters available. I was responding mostly to that builder example in the description which I think would be improved by making it chainable, and if you're pursuing a builder in Python anyway then I don't yet know of a great reason that you wouldn't make it more chainable. I could be way wrong though!

jessex:
Ah, that's fair. Frustrating, but fair.

ohinds:
Here's a related issue that results in a lot of `populate_data` overrides:

I get BondType: 'PENDING - WAITING TO SEE MAGISTRATE', that ain't a BondType, that's a BondStatus, so I would love a way to map that string to BondStatus.PENDING without doing a sad, seemingly unnecessary string match and nothing else in `populate_data`.

jamwalla:
One solution could look like:
```python
is_pending(s): s.startswith('PENDING')
overrides.remap(BondType, is_pending, BondStatus.PENDING)
```
This could be simpler, e.g. if we wanted to narrow this down to substring matches; or it could be more complex, e.g. if we wanted to map to different BondStatuses depending on the rest of the string.

@arian487 had another idea for how this should look.

arian487:
Mine is basically identical to James' solution except instead of using a lambda, I think using a regex could be cleaner?

ohinds:
Seems like you would need:
[...]

PR #954, opened 2019-02-11
enum overrides

jamwalla:
contains the logic, in EnumOverrides, for #816 and #879; @ohinds and I will move the scrapers to this object in a follow-up.


recidiviz/common/constants/enum_overrides.py
```diff
@@ -0,0 +1,92 @@
...
+            self._maps[enum_class][k] = v
+
+        def ignore(self, k: str, enum_class: MappableEnumMeta = None):
+            """Marks strings exactly matching |k| as ignored values for
+            |enum_class|. If |enum_class| is None, ignore |k| for all enums."""
+            self._ignores[enum_class].add(k)
+
+        def map(self,
+                enum_class: MappableEnumMeta,
+                matcher: Callable[[str], bool],
```
arian487: At a high level, I'd prefer to keep the API very simple so I'd rather have map functionality baked into add, and the user can optionally pass in two values which define add behaviour, one is a matcher lambda (or regex!) and the other is the enum_class it overrides to.  By default if neither is set you just match exact strings but if they are set, you have this functionality instead.  That way a user doesn't have to use the lambda function to map exact matches of bond type to something in charge status for example.  

As discussed offline make sure to maintain two separate maps still as you are now so we can prioritize exact matches over lambda matches.

recidiviz/common/constants/mappable_enum.py
```diff
@@ -31,6 +31,10 @@ def __init__(self, cls: type, string_to_parse: str):
         super().__init__(msg)
 
 
+class MappableEnumMeta(EnumMeta):
```
zacharylawrence: As discussed this offline, let's rename:
MappableEnumMeta -> EntityEnumMeta
MappableEnum -> EntityEnum

And pull it into a new PR!

recidiviz/common/constants/enum_overrides.py
```diff
@@ -0,0 +1,92 @@
...
+                         Dict[str, MappableEnum]] = maps
+        self._ignores: Dict[Optional[MappableEnumMeta], Set[str]] = ignores
+        self._lambda_maps: Dict[MappableEnumMeta,
+                                Set[Tuple[Callable[[str], bool],
+                                          MappableEnum]]] = lambda_maps
+
+    def should_ignore(self,
+                      k: str,
+                      enum_class: MappableEnumMeta = None) -> bool:
+        return k in self._ignores[enum_class] or k in self._ignores[None]
```
ohinds: I like that this is concise, but it feels a little odd to execute the exact same test twice in the common case.

ohinds:
> Couple high level comments to start, can you update the documentation with usage and examples as well and make sure that once we agree on this, to add the actual child scraper changes throughout as part of this PR.

I'm in favor of merging this, then taking a pass through the code. I'm happy to take on that task.

jamwalla:
Moved EntityEnum changes out into #968.

recidiviz/common/constants/enum_overrides.py
```diff
@@ -0,0 +1,93 @@
...
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+# ============================================================================
+"""Scraper-specific mappings from strings to enum values."""
+from collections import defaultdict
+from typing import Callable, Dict, Set, Tuple, Optional, Union
+
+from recidiviz.common.constants.entity_enum import EntityEnumMeta, EntityEnum
+
+
+class EnumOverrides:
```
zacharylawrence: Can this class be a frozen `attrs` class? Ideally also making the members of the class also immutable?

Is this worth it?

recidiviz/common/constants/enum_overrides.py
zacharylawrence: Since you're using `sets`, you could consider using the union like:
```
return k in (self._ignores[enum_class] | self._ignores[None])
```

recidiviz/common/constants/enum_overrides.py
```diff
@@ -0,0 +1,93 @@
...
+    def should_ignore(self,
+                      k: str,
+                      enum_class: EntityEnumMeta = None) -> bool:
+        return k in self._ignores[enum_class] or k in self._ignores[None]
+
+    def parse(self,
+              k: str,
+              enum_class: EntityEnumMeta) -> Optional[EntityEnum]:
+        if self.should_ignore(k) or self.should_ignore(k, enum_class):
+            return None
```
zacharylawrence: I think we need to denote "ignored mappings" from "an override not being specified (ie. returning `None` below)". For example, setting a string to "ignore" means I want to ignore what's in the global map. If an enum is never specified in the enum_overrides object, then we should still default to the global map.

recidiviz/common/constants/enum_overrides.py
```diff
@@ -0,0 +1,93 @@
...
+
+        direct_lookup = self._maps[enum_class].get(k)
+        if direct_lookup:
+            return direct_lookup
+
+        for matcher, v in self._lambda_maps[enum_class]:
+            if matcher(k):
+                return v
```
zacharylawrence: If we want to short circuit here, then we need to ensure that `self._lambda_maps` has a deterministic order (ie. Use `ordereddict` instead of a `dict`).

Personally, I think we should just throw an exception if a provided string matches multiple matchers since it seems very unlikely that a scraper writer would knowingly depend on the order of matchers. Ie, something like this might be possible:
```
matches = [v for matcher, v in self._lambda_maps if matcher(k)]
return more_itertools.one(matches)
```

Thoughts on this?

recidiviz/common/constants/enum_overrides.py
```diff
@@ -0,0 +1,93 @@
...
+        self._maps: Dict[EntityEnumMeta,
+                         Dict[str, EntityEnum]] = maps
+        self._ignores: Dict[Optional[EntityEnumMeta], Set[str]] = ignores
+        self._lambda_maps: Dict[EntityEnumMeta,
+                                Set[Tuple[Callable[[str], bool],
+                                          EntityEnum]]] = lambda_maps
+
+    def should_ignore(self,
+                      k: str,
+                      enum_class: EntityEnumMeta = None) -> bool:
```
zacharylawrence: When would you ever call `should_ignore` but not want to specify an enum_class?

recidiviz/common/constants/entity_enum.py
```diff
@@ -92,3 +97,87 @@ def _parse_to_enum(cls, label, complete_map):
...
+            return None
+
+        direct_lookup = self._maps[enum_class].get(label)
+        if direct_lookup:
+            return direct_lookup
+
+        matches = [v for matcher, v in self._lambda_maps[enum_class]
+                   if matcher(label)]
+        if len(matches) > 1:
+            raise ValueError("Overrides map matched too many values from label"
```
jessex: How do we avoid this happening? Is it just a matter of clear documentation and asking scraper writers to ensure that none of their override matchers can ever return multiple values?

jamwalla: Yep, we'll make that clear in the scraper docs and I added a note to the docstring in Builder.add(); I don't think there's a programmatic way to defend against it until it happens.

recidiviz/common/constants/enum_overrides.py
zacharylawrence: As discussed offline, we will just explicitly check `should_ignore` in that case.

recidiviz/common/constants/enum_overrides.py
```diff
@@ -0,0 +1,122 @@
...
+
+from collections import defaultdict
+from typing import Callable, Set, Union
+from typing import Dict, Optional
+
+import attr
+from recidiviz.common.constants.entity_enum import EntityEnum, EntityEnumMeta
+
+
+# pylint: disable=unsubscriptable-object
```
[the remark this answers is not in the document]
jamwalla: done

recidiviz/common/constants/enum_overrides.py
```diff
@@ -0,0 +1,122 @@
...
+@attr.s(frozen=True)
+class EnumOverrides:
+    """Contains scraper-specific mappings from string keys to EntityEnum
+    values. EnumOverrides objects should be created using EnumOverrides.Builder.
+    """
+    _maps: Dict[EntityEnumMeta, Dict[str, EntityEnum]] = attr.ib()
+    _ignores: Dict[Optional[EntityEnumMeta], Set[str]] = attr.ib()
```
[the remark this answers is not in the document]
jamwalla: good call!

recidiviz/common/constants/enum_overrides.py
```diff
@@ -0,0 +1,122 @@
...
+    _global_ignores: Set[str] = attr.ib()
+    _lambda_maps: Dict[EntityEnumMeta, Set['_EnumMatcher']] = attr.ib()
```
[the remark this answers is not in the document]
jamwalla: i like this!

recidiviz/common/constants/entity_enum.pyi
```diff
@@ -0,0 +1,20 @@
+# Stubs for entity_enum (Python 3.7)
+#
```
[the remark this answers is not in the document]
jamwalla: done

Issue #948, opened 2019-02-11
us_ga_gwinnett enum errors


colincadams:
```
 ('us_ga_gwinnett', 'BondType', 'A. THORPE#133'),
 ('us_ga_gwinnett', 'BondType', 'ANDERSON'),
 ('us_ga_gwinnett', 'BondType', 'APPLBY 703'),
 ('us_ga_gwinnett', 'BondType', 'ARIEMMA'),
 ('us_ga_gwinnett', 'BondType', 'BAGLEY #891'),
 ('us_ga_gwinnett', 'BondType', 'BAHAM T1'),
 ('us_ga_gwinnett', 'BondType', 'BLACKBURN #1686'),
 ('us_ga_gwinnett', 'BondType', 'BLAXTON'),
 ('us_ga_gwinnett', 'BondType', 'BOERRIGER 179'),
 ('us_ga_gwinnett', 'BondType', 'BRATIC # 125'),
 ('us_ga_gwinnett', 'BondType', 'BRATIC #125'),
 ('us_ga_gwinnett', 'BondType', 'BREITFELDER 168'),
 ('us_ga_gwinnett', 'BondType', 'BRIGHT'),
 ('us_ga_gwinnett', 'BondType', 'BRITT, J'),
 ('us_ga_gwinnett', 'BondType', 'BROCK'),
 ('us_ga_gwinnett', 'BondType', 'BRUCKS 755'),
 ('us_ga_gwinnett', 'BondType', 'BULLOCK 68'),
 ('us_ga_gwinnett', 'BondType', 'C.G. CLARK 1712'),
 ('us_ga_gwinnett', 'BondType', 'CAIN'),
 ('us_ga_gwinnett', 'BondType', 'CALLAWAY, JN 10'),
 ('us_ga_gwinnett', 'BondType', 'CARTER 417'),
 ('us_ga_gwinnett', 'BondType', 'CASAL, K'),
 ('us_ga_gwinnett', 'BondType', 'CASH, NO BOND'),
 ('us_ga_gwinnett', 'BondType', 'CASH, SURETY, NO BOND'),
 ('us_ga_gwinnett', 'BondType', 'CATOGGIO'),
 ('us_ga_gwinnett', 'BondType', 'CAUDLE SO752'),
 ('us_ga_gwinnett', 'BondType', 'CHEEKS SO643'),
 ('us_ga_gwinnett', 'BondType', 'COFIELD # 1732'),
[...]

PR #955, opened 2019-02-11
fixes for us_ga_gwinnett

jamwalla:
this should fix everything in #948, but let's keep it open so I can check back and see that the right things are getting populated in Arrest.officer_name


recidiviz/ingest/scrape/regions/us_ga_gwinnett/us_ga_gwinnett_scraper.py
```diff
@@ -44,8 +44,14 @@ def populate_data(self, content, task: Task,
...
             for booking in person.bookings:
                 for charge in booking.charges:
+                    bond = charge.bond
                     if charge.fee_dollars and \
                             not charge.fee_dollars.startswith('$'):
                         charge.fee_dollars = None
+                        if bond and bond.bond_type:
+                            booking.create_arrest(officer_name=bond.bond_type)
+                            bond.bond_type = None
+                    if bond.amount == 'NO BOND':
```
colincadams: isn't this already handled in the converter?

recidiviz/ingest/scrape/regions/us_ga_gwinnett/us_ga_gwinnett_scraper.py
```diff
@@ -44,8 +44,14 @@ def populate_data(self, content, task: Task,
         for person in ingest_info.people:
             for booking in person.bookings:
                 for charge in booking.charges:
+                    bond = charge.bond
                     if charge.fee_dollars and \
                             not charge.fee_dollars.startswith('$'):
                         charge.fee_dollars = None
+                        if bond and bond.bond_type:
+                            booking.create_arrest(officer_name=bond.bond_type)
```
colincadams: so here we use `charge.fee_dollars.startswith('$')` to detect that the columns shifted? that makes sense, can you add a test case for where it is shifted?

jamwalla: yeah! added

recidiviz/ingest/scrape/regions/us_ga_gwinnett/us_ga_gwinnett_scraper.py
jamwalla: When bond.amount is NO BOND, the bond type can sometimes be garbage, like 'CASH, SURETY, NO BOND', which per convo with @ohinds we can just throw away (since NO BOND is more restrictive). Is it clearer to set it to None?

PR #956, opened 2019-02-11
data extractor: don't overwrite fields that were previously set to None

jamwalla:
If the data extractor comes across an object for which a field is missing, and then attempts to write a subsequent object using the same field, it didn't create a new object because the field wasn't set on the last one. It should, tho.

Co-Authored-By: colincadams <colincadams@users.noreply.github.com>


recidiviz/ingest/extractor/data_extractor.py
```diff
@@ -53,10 +54,16 @@ def __init__(self, key_mapping_file=None):
                 else value[0].split('.')[0]
                 for value in self.multi_keys.values())
 
+        # Stores key/value pairs of (object id, field) once
+        # _set_or_create_object has attempted to set a value on the object's
+        # field. Gets cleared on each run of extract_and_populate_data.
```
jessex: I don't personally love the idea of maintaining a dirty cache that we need to clear manually on invocations to a highly-trafficked method. It feels like the kind of thing we might eventually screw up when we make some unforeseeable-at-the-moment interface change, leading to nasty parse bugs. Would it be possible to just have this live entirely within the scope of `extract_and_populate_data`?

jamwalla: I agree 100%. Do you think the code is clearer now, where _set_or_create takes the seen map explicitly as a parameter, which has to be passed around the right way in extract_and_populate? I think this is the best solution that doesn't involve restructuring the implementation further, but maybe we should consider that.

jessex: I do think this is an improvement! I like that its more functional/less stateful, and I like that it's more obvious what's happening by looking at the method signature. But I'd like another reviewer to weigh in, too.

jamwalla:
Cool, thanks all! The issue occurs specifically when this missing/empty field is the first in the object, i.e. if we set anything else newly on the object first it won't matter if later fields are missing. This has worked pretty well for the HTML data extractor, where, for example, every booking has an admission date listed first. I can't think of an easy way to handle JSON object structure better, but that approach would definitely seem more correct.

iow I totally agree, @colincadams. I filed #989 but the fix seems big/difficult/possibly not worth, let's run with this for now!

rasmi:
It seems this PR [broke the build](https://travis-ci.org/Recidiviz/recidiviz-data/builds/493231596#L978) -- @jamwalla can you take a look?

rasmi:
(it was actually a flaky test from #985, sorry!)

PR #957, opened 2019-02-11
Rewrite of database README to reflect new process and VM use

zdg2102:
This is pretty much a complete rewrite, so I'd ignore the diff against the existing file, it just adds noise.


recidiviz/persistence/database/README.md
```diff
@@ -1,95 +1,221 @@
 # Database
 
+## Overview
+
+The database is a vanilla SQL database with application-time temporal tables.
+
+### Temporal tables
```
zacharylawrence: Could you also add a quick note about how (& why) we never delete any rows for the master table?

recidiviz/persistence/database/README.md
```diff
@@ -1,95 +1,221 @@
...
+
+1. Run `generate-empty-migration -m <migration name>`.
+
+2. Follow the example in [change\_aggregate\_datetime\_to\_date](https://github.com/Recidiviz/recidiviz-data/blob/master/recidiviz/persistence/database/migrations/versions/997ed5aca81f_change_aggregate_datetime_to_date.py) for how to apply a transformation to existing data.
+
+#### Applying the migration
+
+TODO(garciaz): work with rasmi@ and arian487@ on how this should be timed with creating a new release
+
+1. Send a PR containing the migration for review.
```
zacharylawrence: How is this done? But putting your git credentials on the shared VM?

recidiviz/persistence/database/README.md
zdg2102: Done

recidiviz/persistence/database/README.md
zdg2102: Discussed offline

recidiviz/persistence/database/README.md
```diff
@@ -1,95 +1,223 @@
...
-1. Set environment variables to point to **DEV** database. Run `alembic revision --autogenerate -m "<MIGRATION_NAME>"`. This will generate an initial migration to resolve any differences between the dev and prod schemas.
+Create `/etc/profile.d/prod-data-aliases.sh` with the below content:
 
-2. Run `alembic upgrade head`. This will update the dev database to match the prod schema.
+```bash
+source /etc/profile.d/.env
 
-3. In `recidiviz/persistence/database/migrations/versions`, delete the migration that was just run. As a temporary migration used to update dev only, it should not be checked in.
+alias dev-psql="psql \"sslmode=disable hostaddr=$DEV_DATA_IP \
+user=postgres dbname=postgres\”"
```
jessex: We don't consider the username or the database name to be secrets?

zdg2102: Replaced with environment variables in both the docs and on the actual VM

PR #958, opened 2019-02-11
set fields via __dict__ in IngestObject __init__ methods

jamwalla:
@zacharylawrence I saw you did this for BuildableAttr and I liked the pattern, but I'm not sure if it's worth the readability hit in this case. What do you think?


jamwalla:
@jessex Zach and I went back and forth on this offline, and aren't sure what makes more sense. We could make IngestObjects all BuildableAttrs, which would give us the behavior of `__setattr__` for free, but we aren't currently treating IngestInfo as immutable. It would be a little weird to shuffle around a Person.Builder that contains a list of Booking.Builders, but it could work, and we'd expose only the IngestInfo.Builder.build method. It would be nice to have our "data classes" all share the same format; on the other hand this seems like a pretty big refactoring job for something that's working well now. Do you have any opinions?

jessex:
Hm, tricky. All of the statements in that comment sound reasonable and correct to me but it's hard to weigh them against each other.

To me, the big high-level question is how well `IngestInfo` can be made to fit a builder pattern? If we can make it buildable cleanly then I think we should strive to use `BuildableAttr` for consistency. Otherwise, the statement "it would be nice to have our "data classes" all share the same format" doesn't seem so important, as I think it would be totally reasonable to say that our buildable classes use `BuildableAttr` and our non-buildable ones do not.

>It would be a little weird to shuffle around a Person.Builder that contains a list of Booking.Builders, but it could work, and we'd expose only the IngestInfo.Builder.build method

Yeah, I think this would be pretty annoying to read and follow, and the post-process of building all of the builders floating around seems like going against the grain of the builder pattern. I don't have an idea yet on how to do this without having floating builders, though, since the way we construct ingest info is really piecemeal and variable depending on the structure of the data being extracted.

jamwalla:
We debated some more offline. I'm going to close this and file #1041 to move IngestInfo to attrs - I didn't realize that we lose meaningful typing in this PR. Thanks for the good conversations here, all!

PR #962, opened 2019-02-11
Add foreign key constraints to master ID columns on historical tables

zdg2102:


jessex:
Just to make sure I understand the impetus here: we didn't and still don't have foreign key constraints from temporal table to temporal table. And presumably we didn't have foreign key constraints from temporal table to master table because we were just in the mindset of "don't have foreign key constraints on temporal tables?" But now we realized that we want them to make sure that there is zero chance of a temporal table ever being orphaned?

zdg2102:
@jessex originally when I created the historical tables, I intended their primary key to be the master key column (e.g. the `person_id` column on `person_history` that's shared with the master `person` table). In my mental model at the time, `person_id` wasn't a foreign key, but rather a single primary key that was shared by both the master and historical tables. Then when I actually started setting up the database, I discovered SQLAlchemy requires a unique primary key column on every table, so I created normal primary keys on the historical tables (e.g. `person_history_id`). Ideally I should have changed the master key columns to be foreign keys at that point, but I didn't think about it. I don't think it's particularly likely these relationships will ever get violated or that we'll ever end up with orphaned historical rows, but explicitly marking it as a foreign key seems to me to be a good way to clarify the relationship between these various entities for someone using the schema.

Issue #584, opened 2019-01-10
Figure out how to share qps across vendors/queues


arian487:
Jailtracker is an example where the QPS needs to be shared as it all hits the same backend.

PR #964, opened 2019-02-12
Break out manifest and queue.yaml into region specific directories

### LINKED ISSUE #942: Split `region_maniest.yaml` into per-region files
(Also see #943  for a similar thing with `cron.yaml` and `queue.yaml`.)

The global config file `region_manifest.yaml` is a regular source of merge conflicts for new scrapers. It will be tough to scale contributions beyond 1-2 contributors if every new scraper that gets submitted generates a merge conflict with all the other in-progress scraper pull requests.

The `region_manifest.yaml` file could be split into per-region files since we [own the code](https://github.com/Recidiviz/recidiviz-data/blob/cdd82e30e236430509a8f5dd8709d2fa57be9204/recidiviz/utils/regions.py#L163).

This feature request is to split the single `region_manifest.yaml` file into one YAML file per region, stored next to the region scraper's code, for example, `ingest/scrape/regions/us_in_boone`.

1.  Write a throw away script to break out the `region_manifest.yaml` into a one YAML file per region.
1.  Change the code in `utils/regions.py` to search `ingest/scrape/regions` and load all the YAML files.
1.  If performance isn't adequate (disk search instead of loading one file), decide on a caching strategy.
1.  Update documentation for new scrapers.

Keep the script from the first step until all in-progress scraper PRs have landed so authors can update their PRs while we're migrating.

### LINKED ISSUE #943: Assemble `cron.yaml` and `queue.yaml` from per-region files to prevent merge conflicts
(Also see #942 for a similar thing with `region_manifest.yaml`.)

The global config files `cron.yaml` and `queue.yaml` are regular sources of merge conflicts for new scrapers. It would be tough to scale contributions beyond 1-2 contributors if every new scraper generates a merge conflict with all in-progress scraper PRs.

It looks like `cron.yaml` and `queue.yaml` need to stay as single files, at least for deployment time. This feature request is to split these files into per-region YAML files and assemble the global `cron.yaml`/`queue.yaml` file at deployment time.

1.  Write a temporary script to split the global files into per-region files.
1.  Write a script to assemble each of the per-region files back into single global files. Test that the round-trip through the first two steps generates identical global files.
1.  Test with release engineer to double-check that the files still work for deployment.
1.  Delete global files from the git repo, add them to .gitignore.
1.  Update docs for scraper authors.

Keep the temp script from the first step until all in-flight scraper PRs can be updated.

colincadams:
Any scrapers that are in progress if/when this is merged can run `python -m recidiviz.tools.split_manifest --region REGION` to generate their yaml file and then revert their changes to region_manifest and queue.

fixes #942, fixes #943


colincadams:
Also note some existing issues this fixes:
- `us_nc_union` exists in the `region_manifest.yaml` but has not yet been implemented
- `us_nj_bergen` is implemented but does not have a queue in `queue.yaml`

I think these are a testament to how these two files have become too large and unwieldy, and breaking them up will make it much easier to manage and spot issues.

recidiviz/tests/ingest/scrape/infer_release_test.py
```diff
@@ -32,38 +32,28 @@
...
-            'region_code': 'us_nc',
-            'scraper_package': 'us_nc',
-            'timezone': 'America/New_York'
-        }
+_MANIFESTS = [
+    {
+        'agency_name': 'agency_name',
+        'agency_type': 'jail',
+        'base_url': 'base_url',
+        'queue': 'queue',
```
jessex: This doesn't totally match the actual manifest format anymore, where `queue` is an object.

recidiviz/tools/build_queue_config.py
```diff
@@ -0,0 +1,56 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+# ============================================================================
+"""Builds queue.yaml from the region configs"""
```
jessex: Slight nit: reword "configs" to "manifests"

recidiviz/tools/build_queue_config.py
```diff
@@ -0,0 +1,56 @@
...
+class NoAliasDumper(yaml.Dumper):
+    def ignore_aliases(self, data):
+        return True
+
+def build_queues():
+    queues = []
+    for region in regions.get_supported_regions():
+        if 'name' in region.queue:
+            assert (len(region.queue) == 1 and \
+                    not region.queue['name'].startswith(region.region_code))
```
jessex: Why do we have this constraint? Is it to avoid unnecessary redundancy, i.e. the queue name is basically just a copy of the region code itself?

recidiviz/tools/split_manifest.py
```diff
@@ -0,0 +1,78 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+# ============================================================================
+"""Splits the region manifest into the region-specific directories.
+
+Reads the corresponding entries form `region_manifest.yaml` and `queue.yaml` to
```
jessex: form --> from

colincadams: Done

recidiviz/tools/build_queue_config.py
colincadams: Done

recidiviz/tests/ingest/scrape/infer_release_test.py
colincadams: Done

recidiviz/tools/build_queue_config.py
colincadams: What I was thinking here is a region would either (1) create it's own queue, so they would supply at least a rate and any additional params and a queue would be created with the name of the form `us-pa-greene-scraper` or (2) use an existing queue, in which they would supply the name of the queue to use. I didn't see a purpose for allowing a custom name to be supplied for a region-specific queue.

However, I think it would be better to be more explicit about this yaml structure so that it's more obvious what each does, like rename `name` to `shared` or have a `create_queue` and `existing_queue` top level keys (and either way add documentation for it). Since none of the regions actually share queues yet, maybe it would be better to ditch the `name` param entirely for now and then create a separate PR to support queue sharing. What do you think?

jessex: >Since none of the regions actually share queues yet, maybe it would be better to ditch the `name` param entirely for now and then create a separate PR to support queue sharing.

Normally I would agree with this line of thinking 💯, but given that we have an outstanding need to handle "global" rate-limiting for regions that are hosted on shared infrastructure and we're coming up on the homestretch, I'd like to see us either have the support baked in now or have confidence that the rate-limiting approach is going to rely on something other than shared queues in this fashion.

colincadams: I think vendor-level rate limiting should be handled by sharing a queue, I just think it would be more simpler to get this PR in with all of the region changes, and then add the additional functionality for queue sharing in a follow-up

arian487: Just to chime in, we've discussed offline that the next step to have shared queues (which is most relevant for jailtracker) is to actually figure out what the rate limit of a jail tracker queue is, and also make sure our requests are distributed across N ips (we need to decide N) from the proxy.  And then of course baking that logic into the create scraper script for each vendor that may share their QPS allowance.  We're definitely going to work on that today and tomorrow but IMO its ok to get this change in now.  Its a huge win for the scraper writers who wouldn't need to worry about colliding which each other on the queue and region manifest files.  @colincadams has 'solving shared queues' as the very next item on his docket so it is highly prioritized :)

As for the name, I think we should have a field in the region manifest called shared-queue, and if a queue name is set it uses those defaults, but otherwise creates one per region.

jessex: Thanks, @arian487! All sounds well to me, including the name suggestion.

recidiviz/ingest/scrape/regions/us_vt/manifest.yaml
```diff
@@ -0,0 +1,4 @@
+agency_name: Vermont
+agency_type: unified
+base_url: $url
```
colincadams: Looks like `us_vt` doesn't use the `base_url` as this is how it is set in the existing manifest as well.

colincadams:
Alright, I believe I believe this is good to go. This does not include the functionality in `build_queue_config` for a vendor to create a shared queue, but it does include the functionality for a region to use the shared vendor queue if it exists. The former will be tackled in #584.

PR #966, opened 2019-02-12
Fix minor issues in database README

zdg2102:
- Clarify that you need to run `pipenv shell` before running the migration commands
- Clarify where the `alembic_version` value comes from
- Fix issue with environment variables not distinguishing between the two `prod-data` roles


recidiviz/persistence/database/README.md
```diff
@@ -44,6 +44,8 @@ All migrations should be run from the `prod-data-client` VM, which is accessed b
 
 If it is your **first time logging in to the VM**, run `initial-setup` from your home directory. This will set up the git repository, pip environment, and SSL certificates you will need to run migrations.
 
+NOTE: All commands below assume you are running in your pip environment. To launch it, run `pipenv shell` from the top-level package of the git repository.
+
```
terinpw: Is `pip environment` the right terminology? Or is it `pipenv shell`?  

Not really sure, @rasmi do you know?

jessex: The docs on pipenv shell typically use it in the context of activating your "virtualenv" or "virtual environment." Having said that, I think either of those or "pip environment" or even "pip shell" are reasonable.

PR #973, opened 2019-02-12
validate timezones

colincadams:


recidiviz/ingest/scrape/ingest_utils.py
```diff
@@ -18,15 +18,22 @@
...
+import pytz
 
 from recidiviz.common import common_utils
 from recidiviz.ingest.scrape import constants
 from recidiviz.ingest.models.ingest_info_pb2 import IngestInfo
 from recidiviz.utils import regions
 
 
-def validate_regions(region_list, timezone=None):
+def validate_timezone(timezone: Optional[str]) -> Optional[tzinfo]:
```
zacharylawrence: This is actually parsing the timezone, right?

colincadams: Kind of, it didn't seem that different than what `validate_scrape_types` does, getting the enum values corresponding to the input. I'd be hesitant to name it `parse_timezone` because I think that implies we are doing something smart like cleaning up the string first. What about `lookup_timezone` instead?

zacharylawrence: Got it, makes sense! I like `lookup` and it seems to match the exceptions raised when an invalid string is passed in:
`pytz.exceptions.UnknownTimeZoneError: 'XXXXXXXXXXX'`

PR #977, opened 2019-02-12
Suppress warning created in 698725f2c2b948db84f62f43e89a2f1ad2ad0a0a

ohinds:


recidiviz/ingest/scrape/scraper.py
```diff
@@ -52,6 +53,10 @@ def __init__(self, region_name):
 
         """
 
+        # Passing verify=False in the requests produces a warning,
+        # disable it here.
+        urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
```
jessex: What's the full scope of this? We suppress all warnings related to any request security concerns, for all outbound requests from any scrape task? I wonder if that's a bit heavy-handed.

zacharylawrence:
Does it make sense to update update our pytest `setup.cfg` like this instead of inline code: https://docs.python.org/3/using/cmdline.html?highlight=#cmdoption-w

recidiviz/ingest/scrape/scraper.py
ohinds: Currently, all outgoing requests explicitly do not check SSL certs. We talked about this in person and no one had a problem not checking certs, due to the low likelihood that anyone would try to pretend to be a site we hit. I had to add this somewhere because I found a site that has expired certs (or something). Instead of copying `fetch_page` and disabling checks there, we decided it was ok globally.  

By inspection of the urllib3 source, this warning is only raised when the connection is unverified, so it's a direct consequence of the above described change, and only happens because of that change.

ohinds:
> Does it make sense to update update our pytest `setup.cfg` like this instead of inline code: https://docs.python.org/3/using/cmdline.html?highlight=#cmdoption-w
> 
> Side Note: Should we also enable `-Werror` to prevent these warning in the future? Looking at the state of our code now, it looks like enabling `Werror` creates 144 new errors!

I think I prefer having this warning suppressed in code, but I can be talked out of it. I'm more likely to see it there than in config, but maybe that's just me?

jessex:
>Side Note: Should we also enable `-Werror` to prevent these warning in the future? Looking at the state of our code now, it looks like enabling `Werror` creates 144 new errors!

Yes we should! File an issue for it, if you would be so kind?

recidiviz/ingest/scrape/scraper.py
jessex: >due to the low likelihood that anyone would try to pretend to be a site we hit.

A low likelihood of something that could have truly bad consequences, e.g. intentional injection of made-up information, for a project as critical as ours seems like something you'd still want to protect against.

zacharylawrence:
> > Side Note: Should we also enable `-Werror` to prevent these warning in the future? Looking at the state of our code now, it looks like enabling `Werror` creates 144 new errors!
> 
> Yes we should! File an issue for it, if you would be so kind?

Created #999

zacharylawrence:
@ohinds @jessex Just wondering if there are any updates on this?

jessex:
I'd suggest moving forward by filing an issue to explore verifying SSL certs on outgoing requests, and suppressing the warning in the code, so that it's more visible since we'll want to reexplore this when we do the aforementioned issue. Sound reasonable?

ohinds:
> I'd suggest moving forward by filing an issue to explore verifying SSL certs on outgoing requests, and suppressing the warning in the code, so that it's more visible since we'll want to reexplore this when we do the aforementioned issue. Sound reasonable?

Opened #1300

PR #981, opened 2019-02-12
Shared jailtracker queue

### LINKED ISSUE #584: Figure out how to share qps across vendors/queues

colincadams:
fixes #584


recidiviz/ingest/scrape/vendors/jailtracker/queue.yaml
```diff
@@ -0,0 +1,2 @@
+# TODO: set rate
+rate: 0/m
```
colincadams: @ohinds you said you had gotten a rate, but it was specifically per ip right?

recidiviz/ingest/scrape/regions/us_pa_greene/manifest.yaml
```diff
@@ -1,4 +1,6 @@
 agency_name: Greene County Detention Center
 agency_type: jail
+# TODO: is this out of date?
```
colincadams: @arian487 I'll remove this comment later but this is a jailtracker region now right? Is this base url out of date / unused or they the only one with a custom url?

recidiviz/ingest/scrape/vendors/jailtracker/queue.yaml
arian487: Yea the rate was 30qpm per ip.  I don't think we know the true limit (we don't want to find out).  
@jessex How hard/expensive is it to configure the proxy to balance requests across N ips?

recidiviz/ingest/scrape/regions/us_pa_greene/manifest.yaml
arian487: Yea so the jailtracker scrapers don't actually use the URLs in the manifest at all.  Its built in code based on the region index so we can remove this.

recidiviz/ingest/scrape/vendors/jailtracker/queue.yaml
jessex: So Luminati has _a lot_ of variability. You can choose from IPs coming from the Luminati datacenter, or IPs spread across the residential backbone, or IPs spread across mobile carriers. Then within each of those three options you can select different options for how IPs are allocated. We're using the datacenter option which is cheapest because it has a static pool instead of a dynamic pool, and an order of magnitude smaller pool (400k IPs available versus 4MM in the residential pool). Within the datacenter option, you can either:
* Use a static pool of IPs with unlimited bandwidth (you can use either 1, or 50 or more) and we've been using this with a pool of 1 IP, which is $1 per IP per month. We're paying $2 dollars per month total for both proxies (prod and stage).
* Use a static pool of IPs and pay $0.10/GB for bandwidth, and size the pool however you want, e.g. 10 IPs if you'd like instead of having to jump to 50, with $0.50 per IP per month. 
* Or you can use a pool of 20,000 IPs and pay $0.50/GB for bandwidth and nothing per IP. 

We're using almost 1.5GB per day on big days and have used a total of 8.64GB in the last 30 days. If we stayed at that rate month-over-month then we could hop up to the big pool and pay for bandwidth and not have to worry so much about rate limiting anymore and still pay almost nothing. But I am not sure we have enough information to model how our total bandwidth is going to increase over time.

[There's more detailed info in here](https://luminati.io/faq) but I recall giving you Luminati access previously so I would strongly recommend you and Colin sitting down and looking at this to see if we have the same understanding. You click "Zones" on the left and then click the pencil icon on one of the zones to edit how it's configured. 

P.S. We can also configure which countries are IPs come from, which I recall being an issue in at least one county. CC @ohinds who I think ran into that previously?

colincadams: Alright now that we have 20k ips to play with, where do we think we should set this limit? Is something like 250/m so that we can support 50 regions (of the 150) at 5/m reasonable? And then assuming that goes swimmingly we can bump it up as needed?

PR #985, opened 2019-02-13
Add environment to manifest so we choose where to run

arian487:
This helps us shore up the release process and better control where things run.


recidiviz/ingest/scrape/scraper_control.py
```diff
@@ -109,14 +110,21 @@ def _start_scraper(region, scrape_type):
...
-            start_scraper_thread = threading.Thread(
-                target=_start_scraper,
-                args=(region_name, scraper_type)
-            )
-            start_scraper_thread.start()
-            start_scraper_threads.append(start_scraper_thread)
+            # If we got all regions, we only want to start them in the right
+            # environment.  We only do this if all is passed to still allow
+            # people to manually run any scrapers they wish.
+            if (region_value[0] == 'all' and\
```
colincadams: The '\' shouldn't be necessary as the conditions are already in parentheses

colincadams: This whole thing could be written as `region_value[0] != all or region_env == gae_env`. However, I think this should be pushed down into the `validate_regions`, alongside the timezone check. Then we don't have to worry about 'all' here as well. WDYT?

recidiviz/utils/environment.py
```diff
@@ -45,7 +45,19 @@ def in_prod():
         True if on hosted GAE instance
         False if not
     """
-    return os.getenv('RECIDIVIZ_ENV', '') == 'production'
+    return (os.getenv('RECIDIVIZ_ENV', '') == 'production' or
```
colincadams: nit: This could likely just be `'RECIDIVIZ_ENV' in os.environ` to check if it is set at all. Also fine as is, assuming we won't be adding any other environments.

setup.cfg
```diff
@@ -1,5 +1,5 @@
 [tool:pytest]
-addopts = --cov-report term-missing:skip-covered --cov=.
+addopts = --cov-report term-missing:skip-covered
```
colincadams: reason for this change?

recidiviz/tools/scraper_template/manifest.txt
```diff
@@ -2,3 +2,4 @@ agency_name: $agency
 agency_type: $agency_type
 base_url: $url
 timezone: $timezone
+environment: staging
```
colincadams: we should make the same change in the vendor specific manifests (i.e. jailtracker)

colincadams:
We should also update `recidiviz/tools/split_manifest.py` to set this `environment: staging`, just in case any in progress scrapers use the tool.

recidiviz/utils/regions.py
```diff
@@ -65,6 +65,7 @@ class Region:
     region_code: str = attr.ib()
     agency_name: str = attr.ib()
     agency_type: str = attr.ib()
+    environment: str = attr.ib()
```
colincadams: Add this to the docstring!

recidiviz/ingest/scrape/scraper_control.py
arian487: Agreed this makes sense, done.

recidiviz/utils/environment.py
arian487: I think thats safe but I prefer to be explicit in exactly what environments should support this behaviour.

setup.cfg
arian487: Mistake, its removed now.

recidiviz/tools/scraper_template/manifest.txt
arian487: ah good point.  Done.

recidiviz/utils/regions.py
arian487: Oops, done.

arian487:
Addressed comments!

recidiviz/tools/split_manifest.py
```diff
@@ -44,6 +44,7 @@ def split_manifest(region_to_use: Optional[str] = None):
             'agency_type': old_manifest['agency_type'],
             'base_url': old_manifest['base_url'],
             'timezone': old_manifest['timezone'],
+            'environment': old_manifest['timezone'],
```
colincadams: `'environment': 'staging',`

recidiviz/tests/ingest/scrape/scraper_control_test.py
```diff
@@ -113,6 +119,40 @@ def test_start_timezone(self, mock_docket, mock_tracker, mock_sessions,
...
+            self, mock_docket, mock_tracker, mock_sessions, mock_region,
+            mock_supported, mock_environment):
+        """Tests that the start operation chains together the correct calls."""
+        mock_docket.return_value = None
+        mock_tracker.return_value = None
+        mock_sessions.return_value = None
+        mock_environment.return_value = 'staging'
+        fake_scraper = FakeScraper()
+        mock_region.return_value = FakeRegion(
+            fake_scraper, environment='production')
```
ohinds: Is it important here that we're setting the environment to production here, but to staging in the mock_environment?

setup.cfg
jamwalla: I think this got left over? #1005

PR #986, opened 2019-02-13
fix travis

colincadams:


.travis.yml
```diff
@@ -27,20 +27,20 @@ script:
...
+- git stash --all
+- python -m recidiviz.tools.build_queue_config
 deploy:
-- provider: script
-  script: python -m recidiviz.tools.build_queue_config
-  on:
-    all_branches: true
-    condition: $TRAVIS_TAG =~ ^v[0-9]+\.[0-9]+\.[0-9]+
 - provider: gae
+  skip_cleanup: true
```
jessex: This leaves the `queue.yaml` on the local file system or something?

colincadams: Yes this leaves the `queue.yaml` file accessible on the local file system. Otherwise before each step travis runs `git stash --all` and then after each step it runs `git stash pop`. This meant that `queue.yaml` got stashed. Instead this has us do the `git stash -all` cleanup before deployment and then generates the `queue.yaml`. This seemed fine to me as neither of our true deploy steps generate any files.

jessex: :+1:!

PR #993, opened 2019-02-13
Add booking ID to bond and sentence

zdg2102:


recidiviz/persistence/database/database.py
```diff
@@ -180,12 +185,47 @@ def _save_record_trees(session: Session,
     root_people = [session.merge(root_person) for root_person in root_people]
     session.flush()
 
+    _overwrite_dummy_booking_ids(root_people)
```
jessex: The way that we end up with -1 booking ids in the database is where we fail to invoke `flush` after this call before the codepath terminates, i.e. if an error is raised somewhere in the interim. Where does that next `flush` actually occur and should we just be doing it right here?

zdg2102: I don't believe with this code it should be possible for the -1 IDs to ever actually make it to the database. We set the dummy IDs, flush the session, and then remove the dummy IDs, all one right after another within the same function, so the dummy IDs exist for the shortest time possible. Because there's no possibility of a commit occurring between those steps (because they're consecutive within one function), it means we're still in the scope of a single transaction, so if an error is raised anywhere while the dummy IDs are still set, the whole transaction will fail and none of the dummy IDs will end up in the database.

jessex: Ah I see where I went wrong. I thought that we had `autocommit` set to true on our session maker, but I see now that we let it default to false, which means that `flush` calls just flush the changes down but don't actually commit the transaction. We manually `commit` in the top-level `persistence.write` so we're safe here. Thanks, @zdg2102!

PR #994, opened 2019-02-13
move scrapers to EnumOverrides object

### LINKED ISSUE #816: restructure enum overrides

### LINKED ISSUE #879: It should be easy to add to the vendor scraper enum overrides in the child scraper

jamwalla:
close #816, close #879. Follow-up to #954. 

Co-Authored-By: ohinds <ohinds@users.noreply.github.com>


recidiviz/common/common_utils.py
```diff
@@ -29,9 +30,15 @@ def is_generated_id(id_str: Optional[str]) -> bool:
     return id_str is not None and id_str.endswith(_GENERATED_ID_SUFFIX)
 
 
-def normalize(s: str) -> str:
+def normalize(s: str, remove_punctuation: bool = False) -> str:
```
jessex: 👍

recidiviz/ingest/scrape/regions/us_co_mesa/us_co_mesa_scraper.py
```diff
@@ -181,19 +182,19 @@ def populate_data(self, content, task: Task,
...
-            'CASH CASH': BondType.CASH,
-            'CS CASH SURETY': BondType.UNSECURED,
-            'CSP CASH SURETY PROPERTY': BondType.UNSECURED,
-            'SURETY SURETY': BondType.UNSECURED,
-            'PR PERSONAL RECOGNIZANCE': BondType.NO_BOND,
-            'PRS PERSONAL RECOGNIZANCE CO SIGN': BondType.NO_BOND,
-        }
+        overrides_builder = EnumOverrides.Builder()
+
+        overrides_builder.add('CASH CASH', BondType.CASH)
```
jessex: What's the ordering we're using here? It doesn't appear to be alphabetical by either side.

recidiviz/ingest/scrape/regions/us_ga_toombs/us_ga_toombs_scraper.py
```diff
@@ -41,13 +41,13 @@ def get_jurisdiction_name():
...
-            'NA': None,
-        }
+        overrides_builder = super(
+            UsGaToombsScraper, self).get_enum_overrides().to_builder()
+
+        overrides_builder.add('PROPERTY Bond', BondType.SECURED)
+        overrides_builder.add('REVOKED BOND', BondType.NO_BOND)
+        overrides_builder.add('CHILD SUPPORT RELEASE PAYMENT', BondType.CASH)
+        overrides_builder.add('OR PER JUDGE', BondType.NO_BOND)
+        overrides_builder.add('NA', None)
```
jessex: Should this be an `ignore` call instead? If not, what's the difference between `add('foo', None)` and `ignore('foo')`?

recidiviz/ingest/scrape/regions/us_in_perry/us_in_perry_scraper.py
```diff
@@ -67,17 +70,10 @@ def populate_data(self, content, task: Task,
...
-
-            # Race
-            'UNKOWN': Race.EXTERNAL_UNKNOWN,
-        }
+    def get_enum_overrides(self) -> EnumOverrides:
+        overrides_builder = super(UsInPerryScraper,
+                                  self).get_enum_overrides().to_builder()
+        overrides_builder.add('NOT FILED BY PROSECUTOR', ChargeStatus.PENDING)
+        overrides_builder.add('UNKOWN', Race.EXTERNAL_UNKNOWN)
+        overrides_builder.ignore('SERVE TIME', BondType)
```
jessex: Another ordering nit but in some places we seem to order alphabetically by the first arg regardless of the builder method being invoked, and in others we order first by the method. What sort of convention do we want to develop?

recidiviz/ingest/scrape/regions/us_ga_toombs/us_ga_toombs_scraper.py
jessex: Looking at the code again, it looks like maybe we don't ever want to pass `None` as the second arg to `.add`. This just a mistake?

recidiviz/ingest/scrape/regions/us_co_mesa/us_co_mesa_scraper.py
jamwalla: yeah! left hand side ordering, so we can scan the list for string values. good catch!

recidiviz/ingest/scrape/regions/us_in_perry/us_in_perry_scraper.py
jamwalla: we should order on strings, not builder methods.

recidiviz/ingest/scrape/regions/us_ga_toombs/us_ga_toombs_scraper.py
jamwalla: yep, that should be `ignore`, thanks!

jamwalla:
I ordered the overrides by string value (first arg), regardless of builder method, so we can scan and see if a string is included. Thanks @zdg2102 for the style guide!

PR #995, opened 2019-02-14
Adding us-ky-fulton scraper based on JailTracker

yentangen:


recidiviz/ingest/scrape/regions/us_ky_fulton/manifest.yaml
```diff
@@ -0,0 +1,5 @@
+agency_name: $agency
+agency_type: jail
+base_url: http://www.fultonkyjail.com/FCDC_inmate_list.html
+timezone: America/Chicago
+# TODO(#584): use shared queue once exists
```
arian487: Maybe you haven't pulled but this needs to have: 'environment: staging' as a key as well

yentangen: @arian487 Yes I've run the county with run_scraper (~600 tasks) and there appears to be no issues.
And oops, it's been a few days since I merged with master.. I did that now and added to the manifest.

arian487:
Looks great!

PR #997, opened 2019-02-14
Update scraper_control_test.py

jamwalla:
order not guaranteed from #985


colincadams:
oops didn't see this, pushed a fix in #981

PR #1000, opened 2019-02-14
Move secret file creation to after git stash.

rasmi:
Fixes this: https://travis-ci.org/Recidiviz/recidiviz-data/builds/493345876#L1097


.travis.yml
```diff
@@ -30,6 +27,9 @@ after_success:
 before_deploy:
 - git stash --all
 - python -m recidiviz.tools.build_queue_config
+- 'if [ "$TRAVIS_PULL_REQUEST" = "false" ]; then openssl aes-256-cbc -K
```
colincadams: From looking at the build history, it seems that `before_deploy` is only run if any of the deploy steps will be run, so the `if` check here shouldn't be necessary

rasmi: Good catch, thanks!

PR #1001, opened 2019-02-14
IngestInfo.get_all_*

jamwalla:


recidiviz/ingest/models/ingest_info.py
```diff
@@ -78,6 +78,34 @@ def prune(self) -> 'IngestInfo':
         self.people = [person.prune() for person in self.people if person]
         return self
 
+    def get_all_people(self, predicate=bool) -> List['Person']:
+        return [person for person in self.people if predicate(person)]
+
+    def get_all_bookings(self, predicate=bool) -> List['Booking']:
+        return [booking for person in self.get_all_people()
+                for booking in person.bookings
+                if booking and predicate(booking)]
```
zacharylawrence: Is it intentional to skip empty bookings by checking `if booking`?

jamwalla: nope, we should return empty bookings, good catch!

PR #1002, opened 2019-02-14
Add test ensuring persistence and schema enums match

zdg2102:
Also:
- Remove obsolete TODOs around using Python enums in schema
- Consolidate enums with enum values in `schema.py`
- Add enum values to `schema.py` that were previously missing


recidiviz/tests/persistence/database/schema_test.py
```diff
@@ -120,6 +130,48 @@ def testMasterAndHistoryColumnsMatch_SentenceRelationship(self):
                                        'valid_to'])
 
 
+    # Test case ensuring enum values match between persistence layer enums and
+    # schema enums
+    def testPersistenceAndSchemaEnumsMatch(self):
+        # Mapping between name of schema enum and persistence layer enum. This
```
jessex: So if we created an entirely new enum but didn't add it to `test_mapping` then it would not be included in the test, and there _could_ be a missed mapping, right? There's not a super easy way to catch the existence of a new enum type though, so this might be alright.

recidiviz/tests/persistence/database/schema_test.py
```diff
@@ -120,6 +130,48 @@ def testMasterAndHistoryColumnsMatch_SentenceRelationship(self):
...
+        # map controls which pairs of enums are tested.
+        #
+        # If a schema enum does not correspond to a persistence layer enum,
+        # it should be mapped to None.
+        test_mapping = {
```
jamwalla: should this mapping live in our prod code or be baked into it somehow?

recidiviz/tests/persistence/database/schema_test.py
```diff
@@ -120,6 +130,48 @@ def testMasterAndHistoryColumnsMatch_SentenceRelationship(self):
...
+            'sentence_relationship_type': None,
+            'degree': charge.ChargeDegree,
+            'charge_class': charge.ChargeClass,
+            'charge_status': charge.ChargeStatus,
+            'court_type': charge.CourtType,
+            'report_granularity': None
+        }
+
+        schema_enums_by_name = {}
+        for attribute_name in dir(schema):
```
jamwalla: what do you think about giving the schema enums a semantic type - 
```python
class SchemaEnum(Enum, metaclass=SchemaEnumMeta)
```
so that we could more clearly say
```python
inspect.getmembers(schema, lambda obj: isinstance(obj, SchemaEnumMeta))
```
? If this test is the only time we need to collect or work over all schema enums, this might not be necessary - up to you.

recidiviz/tests/persistence/database/schema_test.py
```diff
@@ -158,3 +210,27 @@ def _assert_columns_match(self, table_a, table_b, table_a_exclusions=None,
...
+    # exclusions in the same way as the master/historical column comparison
+    # tests.
+    def _assert_enum_values_match(self, schema_enum, persistence_enum):
+        schema_enum_values = set(schema_enum.enums)
+        # pylint: disable=protected-access
+        persistence_enum_values = set(persistence_enum._member_names_)
+        # Raise an error instead of asserting so a more explicit message can
+        # be displayed
+        if schema_enum_values != persistence_enum_values:
+            raise RuntimeError('Values of schema enum {schema_enum} did not '
```
jamwalla: is there a reason this is a `RuntimeError` and not a `self.assertEqual`?

recidiviz/tests/persistence/database/schema_test.py
jamwalla: We could even distinguish between persistence and aggregate enums this way.

recidiviz/tests/persistence/database/schema_test.py
zdg2102: Discussed offline, there probably isn't anywhere outside this test that mapping would be used

recidiviz/tests/persistence/database/schema_test.py
zdg2102: The schema enums actually do have a separate type. They're `sqlalchemy.Enum`s rather than regular Python enums. I changed the import so that in the `isinstance` call you can explicitly see from the prefix what the type is

jamwalla: oh, nice!!

recidiviz/tests/persistence/database/schema_test.py
zdg2102: Yeah, because this same method runs for every enum pair, the error message for `assertEqual` doesn't have enough detail because it doesn't specify which pair failed. I have a comment before the equality check clarifying why it's written this way

jamwalla: duh, sorry!! Can we get the same from the `msg` kwarg in assertequal?

zdg2102: Oh nice, I didn't know that existed. That also works better because the `assertEqual` states which values are missing, so I don't need to explicitly list out all the values in the message

PR #1011, opened 2019-02-15
Bullitt2

eitalse:
created jailtracker scraper for Bullitt County


recidiviz/ingest/scrape/regions/us_ky_bullitt/us_ky_bullitt_scraper.py
```diff
@@ -0,0 +1,46 @@
...
+"""Scraper implementation for us_ky_bullitt."""
+from recidiviz.common.constants.charge import ChargeClass
+from recidiviz.common.constants.enum_overrides import EnumOverrides
+from recidiviz.ingest.scrape.vendors.jailtracker.jailtracker_scraper import \
+    JailTrackerScraper
+
+
+class UsKyBullittScraper(JailTrackerScraper):
+    """Scraper implementation for us_ky_bullitt."""
+    def __init__(self):
```
zacharylawrence: Could you remove the extra one after this method? That way we can keep 1 space between these methods.

recidiviz/ingest/scrape/regions/us_ky_bullitt/us_ky_bullitt_scraper.py
```diff
@@ -0,0 +1,46 @@
...
+        A JailTracker landing page URL ends with: "/jailtracker/index/<INDEX>".
+        This value can either be text or a number. In either case, this method
+        should return the value as a string.
+        """
+        return 'BULLITT_COUNTY_KY'
+
+    def get_enum_overrides(self) -> EnumOverrides:
+        overrides_builder = super(UsKyBullittScraper,
+                                  self).get_enum_overrides().to_builder()
+        # Add mappings here by calling overrides_builder.add() or .ignore()
```
zacharylawrence: @eitalse @jamwalla Thoughts on if we should leave this comment here? Or have child scrapers remove it?

PR #1015, opened 2019-02-15
correctly compare bonds/sentences when matching charges

### LINKED ISSUE #1014: call relevant is_match on bond and sentence when matching charges
currently we determine charge equality by using `==` equality, not match logic, on is_charge_match.

terinpw:
closes #1014 

Co Authored By: jamwalla


recidiviz/persistence/entity_matching_utils.py
```diff
@@ -162,7 +162,17 @@ def is_charge_match(
...
+
+    if db_entity.external_id or ingested_entity.external_id:
+        return db_entity.external_id == ingested_entity.external_id
+
+    sentences_match = is_sentence_match(
+        db_entity=db_entity.sentence, ingested_entity=ingested_entity.sentence)
+    bonds_match = is_bond_match(
+        db_entity=db_entity.bond, ingested_entity=ingested_entity.bond)
+
+    return _sanitize_charge(db_entity) == _sanitize_charge(ingested_entity) \
```
jessex: Micro-optimization nit: if we put `sentences_match` and `bonds_match` at the front of the conditional, we don't need to evaluate the two `_sanitize_charge` calls unless there might actually be a match.

terinpw: Good point -- done

PR #1016, opened 2019-02-15
Added JRE installation instructions.

rasmi:


README.md
```diff
@@ -91,6 +91,18 @@ pipenv shell
 
 Finally, run `pytest`. If no tests fail, you are ready to develop!
 
+**NOTE**: If some `recidiviz/tests/ingest/aggregate` tests fail, you may need to install the Java Runtime Environment (JRE).
+
+On a Mac with [Homebrew](https://brew.sh/), you can install the JRE with:
```
jessex: Are there any version requirements to be aware of?

rasmi:
I'm fine adding a line that says:

> Alternatively, you can ignore those tests with:
> ```
> pytest --ignore=recidiviz/tests/ingest/aggregate
> ```

Our volunteer scraper writers need not worry about aggregate ingest tests -- it may be a pain to install the JRE for some.

jessex:
I think both of these instructions are worth mentioning. Let's be as thorough as possible.

README.md
rasmi: According to Tabula docs, Java >= 7: https://github.com/tabulapdf/tabula/tree/cd8a1b03e1d499a6a480f7d38c1b1ba86967935c#using-tabula

If installing with `brew` or `apt`, you have to explicitly list a lower version (<=6) for these commands not to work, so the instructions should be fine, but I'm happy to add `JRE >=7` to be explicit!

PR #1017, opened 2019-02-15
match bonds and sentences on db relationship <= ingested relationship

jamwalla:


jessex:
Can you provide more context for this one, @jamwalla or @terinpw?

jamwalla:
This is a corner case that came up while discussing #1015 - if an ingested child entity appears to match with one in the database, except that it has more parent relationships than it did before, we should still match it. For example, if a booking has a single charge with a single bond, and then the next day it has two charges that share that same bond, we should call the db and ingested bonds a match. The opposite relationship is not true: if two charges share a bond and then one of those charges disappears from the website, we need to create a new bond for the remaining charge, and continue to store historical information about the charge that has disappeared.

recidiviz/persistence/entity_matching.py
```diff
@@ -198,8 +198,9 @@ def _is_match_with_relationships(*, db_entity, ingested_entity):
         matcher = getattr(utils, 'is_{}_match'.format(name))
         obj_match = matcher(db_entity=db_entity,
                             ingested_entity=ingested_entity)
-        relationship_match = ing_relationship_map[ing_entity_id] == \
-                             db_relationship_map[db_entity_id]
+        relationship_match = (db_relationship_map[db_entity_id] <=
```
jessex: What does the operation actually check? `db_relationship_map` and `ing_relationship_map` are both `Dict[str, Set[Entity]]` so we are using the `<=` operator to compare two sets of entities. I've never seen that before!

jamwalla: subset! `a <= b` is short for `a.issubset(b)`

terinpw: It functions as a subset operator -- `a <= b` with sets is equivalent to `a.issubset(b)`.  So here we're checking that all relationships to the db bond/sentence exist in the ingested information (although the ingested bond/sentence can have new relationships as @jamwalla talked about). 

Should we add a comment to this effect for readability? or maybe use `issubset` directly?

terinpw: @jamwalla beat me to it haha

jessex: Makes sense! I think that using the comparison operator in that fashion feels a bit magical, so I would suggest an explicit call to `issubset`.

PR #1030, opened 2019-02-19
Jefferson County, KY scraper.

mutexlox:


mutexlox:
Oops, somehow description got lost when I tried to add it with the `hub` command.

I wrote a scraper for Jefferson County, KY. This county uses Archonix, so all I wrote was the configuration.

recidiviz/ingest/scrape/regions/us_ky_jefferson/us_ky_jefferson.yaml
```diff
@@ -0,0 +1,24 @@
...
+  "Booking #": booking.booking_id
+  Booking Date-Time: booking.admission_date
+
+css_key_mappings:
+  "ctl00_ContentPlaceHolder1_spnInmateName": person.fullName
+
+keys_to_ignore:
+  - Custody Status
+  - Release Date-Time
+  - BirthDate
```
arian487: Why do we ignore birthdate and custody status?

mutexlox: I was basing this on recidiviz/ingest/scrape/regions/us_mt_gallatin/us_mt_gallatin.yaml, which ignores custody status.

I'm also not sure in general what keys are available to map to (I wasn't able to find documentation for it). Can you point me in the right direction there?

jamwalla: Documentation for IngestInfo is [here](https://github.com/Recidiviz/recidiviz-data/tree/master/recidiviz/ingest/models) and should list all the fields (though the lists of valid values might be out of date, sorry!).

The Gallatin scraper should probably scrape custody status, though at a quick glance it seems to always be empty. I'd add it anyways, since nothing will be written if the field is blank and we'll catch issues later if the field isn't what we expect.

PR #1031, opened 2019-02-19
Okeechobee, Florida county scraper

dmcc:
This is my first scraper, so feedback is especially welcome!

I'm not sure about some of the mappings, particularly what to do about the various things I mapped to `ChargeStatus` and `HoldStatus` values. In general, this county seems to put many different types of things in one field ("Booking Type"), so I wasn't sure the best way to interpret them.

Also includes a simple version of `CsvDataExtractor`, since it seemed like a useful thing to pull out.


recidiviz/ingest/extractor/csv_data_extractor.py
```diff
@@ -0,0 +1,84 @@
...
+            for k, v in row.items():
+                if k not in self.all_keys:
+                    raise ValueError('Unmapped key: %s' % k)
+
+                if not v:
+                    continue
+                self._set_value_if_key_exists(k, v, ingest_info)
+
+    def _set_value_if_key_exists(self, lookup_key, value, ingest_info):
+        seen_map: Dict[int, Set[str]] = defaultdict(set)
```
jamwalla: can you pull `seen_map` up into `_extract` so that it's created once per call to `extract_and_populate_data`, and keeps track of keys for the whole extraction? Or is the idea that this isn't necessary since we know we expect one person per row.

recidiviz/ingest/scrape/regions/us_fl_okeechobee/us_fl_okeechobee_scraper.py
```diff
@@ -0,0 +1,135 @@
...
+        elif reason in _HOLDS:
+            # Convert to a hold status.
+            hold = booking.get_recent_hold() or booking.create_hold()
+            hold.status = reason
+            if reason == 'EXTRADITION':
+                hold.jurisdiction_name = reason
+            booking.admission_reason = None
+            booking.prune()
+
+    def get_more_tasks(self, content, task: Task) -> List[Task]:
```
jamwalla: this could be the single task returned from `get_initial_task` instead

recidiviz/ingest/extractor/csv_data_extractor.py
dmcc: I think it shouldn't be necessary given the one person per row assumption, but I'm still new to this code. I've moved it so that it's made once per person just to be safe.

recidiviz/ingest/scrape/regions/us_fl_okeechobee/us_fl_okeechobee_scraper.py
dmcc: If I moved this logic into `get_initial_task`, it would leave `get_more_tasks` abstract, so I think I'd still need to have a stub version of `get_more_tasks`. Is that preferable?

recidiviz/ingest/extractor/csv_data_extractor.py
jamwalla: Thanks! I think this is the "right" place, i.e. in line with the other extractors, so if we extend what this extractor can do we'll be set up properly. But you're right that this isn't necessary right now.

recidiviz/ingest/scrape/regions/us_fl_okeechobee/us_fl_okeechobee_scraper.py
jamwalla: Ah, yeah, I saw this in another scraper recently. I wonder if `get_more_tasks` shouldn't be abstract but should raise a `NotImplementedError` if it's called in the base scraper. This looks good for now!

PR #1035, opened 2019-02-19
add Ky allen prison

shadowind:


recidiviz/ingest/scrape/regions/us_ky_allen/manifest.yaml
```diff
@@ -0,0 +1,6 @@
+agency_name: Allen County Detention Center
+agency_type: prison
```
jamwalla: i *believe* this is a jail and not a prison?

jamwalla:
made that one change and merged, thank you for this!

PR #1039, opened 2019-02-20
Bledsoe County TN Scraper

jdulotr:
Hi there, This is my third try for scraping. In this one, the webpage given in the trix is leading to a money transfer page to inmates and thus the only information can be scraped is the inmate name. 
Please let me know if it's fine with only inmate names.


recidiviz/ingest/scrape/regions/us_tn_giles/us_tn_giles_scraper.py
```diff
@@ -0,0 +1,67 @@
...
+from recidiviz.common.constants.enum_overrides import EnumOverrides
+from recidiviz.ingest.scrape.vendors.jailtracker.jailtracker_scraper import \
+    JailTrackerScraper
+from recidiviz.common.constants.bond import BondType
+from recidiviz.ingest.scrape.task_params import ScrapedData, Task
+from recidiviz.ingest.models.ingest_info import IngestInfo, Person
+from recidiviz.ingest.scrape import scraper_utils
+
+
+class UsTnGilesScraper(JailTrackerScraper):
```
jamwalla: Looks like this is from #1026, so we'll be sure to merge that in before this one.

recidiviz/tests/ingest/scrape/regions/us_tn_bledsoe/us_tn_bledsoe_scraper_test.py
```diff
@@ -0,0 +1,62 @@
...
+        task = Task(
+            task_type=constants.TaskType.GET_MORE_TASKS,
+            endpoint=None,
+        )
+        expected_result = []
+
+        self.validate_and_return_get_more_tasks(content, task, expected_result)
+
+
+    def test_populate_data(self):
```
jamwalla: The goal here is to add a fixture HTML file in your test directory, containing an anonymized example of what the roster could look like. The test should call `validate_and_return_populate_data` with the sample page and the ingest info we expect as a result. You didn't have to do this for some of the other scrapers because the vendors already have test implementations that the child scrapers call.

See [Unit Tests](https://github.com/Recidiviz/recidiviz-data/tree/master/recidiviz/tools/scraper_template#unit-tests) from the scraper_template readme or look to the other scrapers for examples.

recidiviz/tests/ingest/scrape/regions/us_tn_bledsoe/us_tn_bledsoe_scraper_test.py
```diff
@@ -0,0 +1,62 @@
...
+
+class TestUsTnBledsoeScraper(BaseScraperTest, unittest.TestCase):
+
+    def _init_scraper_and_yaml(self):
+        # The scraper to be tested. Required.
+        self.scraper = UsTnBledsoeScraper()
+        # The path to the yaml mapping. Optional.
+        self.yaml = None
+
+    def test_get_more_tasks(self):
```
jamwalla: I think we could safely leave this off, since get_more_tasks shouldn't be called.

recidiviz/ingest/scrape/regions/us_tn_bledsoe/us_tn_bledsoe_scraper.py
```diff
@@ -0,0 +1,47 @@
...
+                      ingest_info: IngestInfo) -> Optional[ScrapedData]:
+        names = content.xpath('//*[@class="inmate_name"]')
+
+        for name in names:
+            ingest_info.create_person(full_name=name.text)
+
+        return ScrapedData(ingest_info=ingest_info, persist=True)
+
+    def get_more_tasks(self, content, task: Task) -> List[Task]:
+        return []
```
jamwalla: does `pass` work here? I wonder if this shouldn't be abstract!

recidiviz/ingest/scrape/regions/us_tn_bledsoe/manifest.yaml
```diff
@@ -0,0 +1,5 @@
+agency_name: Bledsoe County Sheriff's Department
+agency_type: jail
+base_url: https://www.tbejail.com/inmate_deposit.php?jail_id=5&inmate_letter=
```
jamwalla: Interesting, there are a bunch of jails at the base url, so this is sort of like a 'vendor'. You don't have to do it in this PR, but we should pull the scraping logic out into a `TbeJailScraper` or something!

recidiviz/ingest/scrape/regions/us_tn_bledsoe/us_tn_bledsoe_scraper.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+class UsTnBledsoeScraper(BaseScraper):
+    """Scraper implementation for us_tn_bledsoe."""
+    def __init__(self):
+        super(UsTnBledsoeScraper, self).__init__('us_tn_bledsoe')
+
+    def populate_data(self, content, task: Task,
+                      ingest_info: IngestInfo) -> Optional[ScrapedData]:
+        names = content.xpath('//*[@class="inmate_name"]')
```
jamwalla: Can you scrape the 'Inmate ID's as well? It's true that we don't need names *and* person_ids, but might as well store both in case we are later able to avoid storing these names at all.

jdulotr: Thanks for the suggestion! In this case, is it OK if I store it in the field external_id, or should I create another field?

recidiviz/tests/ingest/scrape/regions/us_tn_bledsoe/us_tn_bledsoe_scraper_test.py
jdulotr: I added a fixture from first two people on the html page.

recidiviz/tests/ingest/scrape/regions/us_tn_bledsoe/fixtures/report.html
```diff
@@ -0,0 +1,170 @@
...
+						<table class='datagrid1'>
+							<thead>
+								<th>Last Name, First Name - Middle Name</th>
+								<th>Last Name, First Name - Middle Name</th>
+							</thead>
+							<tbody>
+								<tr>
+									<td id='inmate_id_33821'style='font-size:18px;font-weight:bold;height:30px'>
+										<span class='inmate_name'>
+											BANKS,
```
jamwalla: can you change this name and the one below? Either use `PERSON ONE` or make up a name like `BART SIMPSON`

jamwalla:
Merging this with an issue filed to vendor-ize this later. Thanks for this!

PR #1042, opened 2019-02-20
Update travis to use docker

zacharylawrence:
Note: We acknowledge that the time needed to download the cached docker image is nearly the same as the time needed to download the dependencies directly. Through testing we've determined that using the cached image is slightly faster than rebuilding the image entirely (~180 sec vs ~215 sec). If this changes, we should stop using the Travis cache.

If we wanted to further reduce the time to build our docker image, we should consider moving off the base Ubuntu image.


.travis.yml
```diff
@@ -1,59 +1,35 @@
...
+  - docker save recidiviz-image | gzip -c > ~/docker-images/recidiviz-image.tgz
+
+  # Finally, run our docker application
+  - docker run --name recidiviz -d -t recidiviz-image
+
 script:
-- pipenv run pylint *.py recidiviz
-- pipenv run pytest recidiviz/tests/ --with-emulator
-- pipenv run mypy recidiviz
-after_success:
```
colincadams: we lost everything after this line, we should still be able to use travis for deployment when using the docker image, right?

The only one that seems like it could be trouble is running the `build_queue_config` command in python, and having the output file be accessible outside of the docker container for the `gae deploy queue.yaml` to work. Is this feasible?

zacharylawrence: Yeah, great point. Updated to include deploying to staging.

.travis.yml
```diff
@@ -1,59 +1,76 @@
...
+  - docker run --name recidiviz -d -t recidiviz-image
+
 script:
-- pipenv run pylint *.py recidiviz
-- pipenv run pytest recidiviz/tests/ --with-emulator
-- pipenv run mypy recidiviz
+  # Order test commands with fastest first to ensure we fail fast
+  - docker exec -it recidiviz pipenv run mypy recidiviz
+  - docker exec -it recidiviz pipenv run pylint recidiviz
+  - docker exec -it recidiviz pipenv run pytest
```
colincadams: `--with-emulator` makes pytest run the "integration" tests that rely on the emulator

Dockerfile
```diff
@@ -17,10 +17,23 @@ ENV TZ America/New_York
 
 RUN pip3 install pipenv
 
-ADD . /app
+# First, add only the Pipfiles to ensure we cache `pipenv sync` when
+# application code is updated but the Pipfiles remain untouched
+RUN mkdir /app
```
colincadams: do we have to make the directory? looks like we didn't have this line before

.travis.yml
```diff
@@ -1,59 +1,76 @@
...
-  config: staging.yaml
-- provider: gae
-  keyfile: client-secret-staging.json
-  project: recidiviz-staging
-  on:
-    all_branches: true
-    condition: $TRAVIS_TAG =~ ^v[0-9]+\.[0-9]+\.[0-9]+
-  config: cron.yaml
-notifications:
-  slack:
```
colincadams: should we keep slack notifications? I'm not even in that channel but presumably some people find it useful?

jessex: I'd like to keep it for now. I do find it somewhat useful though I acknowledge the noise it creates. But I think the noise will decrease post-fellowship so I'd like to keep it in.

zacharylawrence: SGTM

Dockerfile
```diff
@@ -17,10 +17,34 @@ ENV TZ America/New_York
...
-RUN pipenv sync
+RUN if [ "$DEV_MODE" = "True" ]; \
+    then pipenv sync --dev; \
+    else pipenv sync; \
+    fi
+
+# Add the rest of the application code once all dependencies are installed
+ADD . /app
 
 EXPOSE 8080
```
colincadams: did we ever figure out if there was a way to have the docker container stay up without running a command?

.travis.yml
```diff
@@ -1,59 +1,83 @@
...
-- coveralls
+  - coveralls
+
 before_deploy:
-- git stash --all
-- python -m recidiviz.tools.build_queue_config --environment staging
-- openssl aes-256-cbc -K $encrypted_f7b0b16e1068_key -iv $encrypted_f7b0b16e1068_iv
-  -in client-secret-staging.json.enc -out client-secret-staging.json -d
+  # Generate queue.yaml
+  - docker exec -it recidiviz pipenv run python -m recidiviz.tools.build_queue_config --environment staging
```
colincadams: nit: we probably don't need `-it` here but not positive

zacharylawrence: As discussed offline, we'll leave `-it` in here to ensure any errors in `build_queue_config` are printed to the Travis UI output.

Dockerfile
zacharylawrence: As discussed offline, we'll leave this here to ensure that the docker image stays alive when exec-ing other processes (eg. bash)

Issue #752, opened 2019-01-28
New world: Expected number of charges >= number of bonds


colincadams:
https://pantheon.corp.google.com/errors/COKkzfLNwOrUlQE?time=PT6H&project=recidiviz-123

colincadams:
#861 has an example request

Issue #696, opened 2019-01-22
handle sites where bonds don't match our schema


jamwalla:
Two vendors, NewWorld and IML, have separate tables for charges and bonds. There is no clear link between the bonds and charges, and there may be more charges than bonds or more bonds than charges.

Our schema defines a bond as the child of a charge, but for these sites it seems like bonds and charges are both children of a booking.

- IML provides case numbers that are shared with charges and bonds. But not all bonds have a case number.
- NewWorld provides no links, we guess that the charges and bonds are listed in matching order.

We've discussed summing bond amounts to at least have accurate total bond amounts, but we'd lose BondType information, which may be different for different bonds in the same booking.

jamwalla:
Done for New World, needs to be done for IML.

PR #1049, opened 2019-02-20
simplify new world vendor

jamwalla:
create dummy charges for each bond, since we have no relationship between bond and charge


recidiviz/ingest/scrape/vendors/newworld/newworld_scraper.py
```diff
@@ -17,96 +17,94 @@
...
-
-        for charge in ingest_info.get_all_charges():
-            if charge.offense_date == "No data":
-                charge.offense_date = None
-            if charge.bond and charge.bond.bond_id == "No data":
-                charge.bond = None
+        yaml_template = os.path.join(os.path.dirname(__file__),
+                                     'newworld_{}.yaml')
+
+        person_extractor = HtmlDataExtractor(yaml_template.format('person'))
```
colincadams: look at merging some of these (i.e. one for person/booking/charge and then one for bonds) in case it makes it more readable

recidiviz/tests/ingest/scrape/regions/us_fl_nassau/us_fl_nassau_scraper_test.py
```diff
@@ -130,6 +112,11 @@ def test_populate_data(self):
             offense_date="12/9/2015 4:50 PM",
             charge_class="Misdemeanor"
         )
+        booking.create_charge().create_bond(
```
colincadams: will this work with persistence? should we instead change our schema to support not knowing how bonds and charges relate?

jamwalla: This should work with persistence (I think it was @zdg2102's suggestion!). But I'm definitely interested in allowing bonds and charges to not be related. Is that too large a schema overhaul, @zdg2102?

recidiviz/ingest/scrape/vendors/newworld/newworld_scraper.py
jamwalla: The booking and charge extractors were definitely worth merging! I kind of like having the booking and person extractors separate, since we have to explicitly loop through the "booking" HTML elements. Otherwise we'd extract the whole page minus bonds in one go, and then loop through the bond tables and trust that they line up with the extracted bookings, which seems more error-prone. (But possibly more readable!) What do you think?

colincadams: this looks good to me!

recidiviz/tests/ingest/scrape/regions/us_fl_nassau/us_fl_nassau_scraper_test.py
zdg2102: I prefer to keep bonds and charges linked because it's a better representation of the real-world relationship. Is this pattern (of bonds we can't properly associate with any charges) something we're running into frequently, or is this scraper the only one?

jamwalla: It's definitely not infrequent - @ohinds has seen this before and NewWorld is a vendor so some portion of our data has this issue. But the majority of sites don't have this problem, I think it's safe to say. I'm in favor of using this workaround in these cases.

zdg2102: Okay, if that's the case, I'd be in favor of continuing to use dummy charges like this unless it becomes a more prevalent problem. With this current setup, what is getting set as the charge status on those dummy charges?

jamwalla: Huh, yeah, the default is UNKNOWN_FOUND_IN_SOURCE. Can we do better?

zdg2102: We could create a status like `DUMMY_CHARGE` or something like that? If we did that, what would the behavior be if the bond that requires the dummy charge gets removed?

jamwalla: Regardless of what we put in it, if the bond gets removed on the website the charge will be marked `UNKNOWN_REMOVED_FROM_SOURCE`, which is why I was thinking it would be ok to stick with the default. But our analysis would work slightly better over `DUMMY_CHARGE`s, we'd just have to note that they're still dummies once they're removed and only live in historical tables.

jamwalla: talked offline, we're not going to sweat this. Charge counts don't really matter and anyone reading this would be able to see what's going on really easily!

jamwalla:
#696 #752

PR #1050, opened 2019-02-20
data extractor multi-keys indexed by grandparent

jamwalla:
If multi-key classes exist alongside their grandparents in a table, the index may be used to refer to the index of the grandparent. For example, in a table with booking dates, charge names and bond amounts, the 3rd bond amount refers to the third booking's first charge's bond.

This also updates the NewWorldScraper to scrape `charge.charging_entity` rather than `arrest.arrest_agency` because scraping arrest data as multi-keys produces "undefined behavior" (what is the arrest of the third charge?). The extractor will attempt to index off the booking and may overwrite earlier arrest data.


recidiviz/ingest/extractor/data_extractor.py
```diff
@@ -109,15 +109,37 @@ def _get_parent(
...
+        if is_multi_key and class_to_set != 'person':
             parent_cls_to_set = HIERARCHY_MAP[class_to_set][-1]
-            if index != 0 and parent_cls_to_set in self.multi_key_classes:
-                grandparent = self._find_parent_ingest_info(
-                    ingest_info, HIERARCHY_MAP[parent_cls_to_set], index)
+            grandparent_cls_to_set = HIERARCHY_MAP[parent_cls_to_set][-1] if \
+                HIERARCHY_MAP[parent_cls_to_set] else None
+            grandparent = self._find_parent_ingest_info(
+                ingest_info, HIERARCHY_MAP[parent_cls_to_set], index)
+            recent = getattr(grandparent, 'get_recent_' + parent_cls_to_set)()
```
arian487: I know this is not all you but I think at this stage we should create helper functions that return the get_recent_* function as well as the create_* functions.

Issue #469, opened 2019-01-03
Investigate if we can use gevent workers


colincadams:
Gevent is [supposedly much faster than eventlet](https://stackoverflow.com/questions/12669971/what-does-make-significant-difference-of-performance-between-eventlet-and-gevent) but [doesn't seem to work on appengine](https://stackoverflow.com/questions/54030237/can-gevent-workers-be-used-in-the-python-3-app-engine-standard-environment?noredirect=1#comment94897359_54030237).

Let's see how my SO question does and update our usage if we get any feedback!

PR #1053, opened 2019-02-21
Make infer release called per region instead of serially in one task

terinpw:


Pipfile.lock
```diff
@@ -310,8 +338,16 @@
                 "sha256:d634a7ea1fc3380ff96f9e44d8d22f38418c1c381d5fac680b272d7d90883720",
                 "sha256:d97b0661e1aead761f0ded3b769044bb00ed5d33e1ec865e891a8b128bf7c656"
             ],
+            "markers": "platform_python_implementation == 'CPython'",
             "version": "==0.4.15"
         },
+        "grequests": {
```
colincadams: [A line was added to the `grequests` readme](https://github.com/kennethreitz/grequests/commit/bdf7387a4aa5ede387d5638c405aa5aad490d5de) in 2017 recommending to use other packages for this. It isn't clear to me why we should prefer other packages over this on, any thoughts here?

In general I'm fine using this library, and depending on `gevent` (which we want to do for #469 anyway), the note just makes it seem like this library may no longer be maintained.

recidiviz/ingest/scrape/infer_release.py
```diff
@@ -19,36 +19,64 @@
...
 infer_release_blueprint = Blueprint('infer_release', __name__)
+_INFER_RELEASE_URL_PREFIX = \
+    'https://recidiviz-123.appspot.com/infer_release/release/'
 
 
 @infer_release_blueprint.route('/release')
 @authenticate_request
 def infer_release():
-    regions = get_supported_regions()
+    region_args = get_values('region', request.args)
```
colincadams: Here you can use `ingest_utils.validate_regions` which will handle checking for all as well as checking for the environment each region runs in.

Then I think you could just check `len(region) > 1`, if so get the urls and make the async calls, otherwise actually infer_release for the given region.

recidiviz/utils/regions.py
```diff
@@ -102,6 +102,7 @@ def get_queue_name(self):
         return self.shared_queue if self.shared_queue \
             else '{}-scraper'.format(self.region_code.replace('_', '-'))
 
+
```
colincadams: unrelated to this PR, we should probably just add this two line requirement to our pylint config

Pipfile.lock
colincadams: Actually now it looks like requests has built-in support? http://docs.python-requests.org/en/v0.10.6/user/advanced/#asynchronous-requests

colincadams: Just kidding that's a very old version

colincadams: It looks like we could [just use concurrent futures](https://stackoverflow.com/a/46144596/5648860), or we could use [aiohttp](https://aiohttp.readthedocs.io/en/stable/) but then we have two different networking libraries

jessex: @colincadams I also am a bit confused as to why Kenneth added that recommendation. One is a self-confessed experiment and the other appears to also be an extension of sorts for `requests`.

terinpw: Chatted offline with @arian487 and @colincadams, and we're going to create a flow by which batch_writes and infer release are triggered after a region's scrape is over. At this point, we won't need to support queuing infer_release for all regions at once. 

Because of that the thought was to - for now - use grequests, knowing it's only going to be used for a week or two.

recidiviz/ingest/scrape/infer_release.py
terinpw: Good point with `validate_regions` -- done.

recidiviz/utils/regions.py
terinpw: Filed #1098, good point.

recidiviz/ingest/scrape/infer_release.py
```diff
@@ -19,36 +19,65 @@
...
+        return '', HTTPStatus.OK
+
+    _infer_release_for_region(more_itertools.one(regions))
+    return '', HTTPStatus.OK
+
+
+def _queue_region_tasks(urls: List[str]) -> None:
+    rs = (grequests.get(u) for u in urls)
+    grequests.map(rs,
+                  exception_handler=lambda r, e:
```
colincadams: should this retry or is logging the error enough and catch 'em next time?

terinpw: Good point -- I think if our argument is that we're going to replace this flow with a region specific one, we should just leave it as a log statement now.

PR #1060, opened 2019-02-22
upgrade opencensus (and others)

colincadams:


Pipfile
```diff
@@ -27,7 +27,7 @@ more-itertools = "*"
 lxml = "*"
 cattrs = "*"
 # TODO(569): Upgrade to next pypi opencensus version that includes necessary exporter fixes
```
jessex: @colincadams were we able to address this with the upgrade? It looks like that issue number may be mistaken, though.

colincadams: I'm not sure yet but I don't believe so. I attempted to run this in production today to see where it got stuck but ran into database related issues, so it never actually tried to export metrics. Will try again next week, or just add more metrics earlier to see if I can debug further.

PR #1063, opened 2019-02-22
clean up db->entity converter to use general convert/convert_all

terinpw:


recidiviz/persistence/database/database_utils.py
```diff
@@ -49,73 +49,18 @@ def for_cls(src_cls):
...
-        src: A schema Person or entity Person object
-    Returns:
-        The converted object, a schema or entity object.
-    """
-    return _convert(src)
-
-
-def convert_bookings(bookings_src):
+def convert_all(bookings_src):
     """Converts the given list of bookings to the correct objects
```
jamwalla: can you update the parameter name and comment?

terinpw: Updated!

terinpw:
Chatted with @jamwalla offline, going to add a separate PR for typing and another small change.

PR #1064, opened 2019-02-22
pushing bell

eitalse:


kevinaud:
https://omsweb.public-safety-cloud.com/jtclientweb/jailtracker/index/BELL_COUNTY_BELL takes me to an invalid facility page. Are we sure this jailtracker index is right?

jamwalla:
Good find! I uploaded a fix (#1273).

PR #1068, opened 2019-02-22
Change scraper example with a county to use a two-letter state name.

markfickett:
This is consistent with other examples, and with scraper names in the repo.


jamwalla:
And just as a note to future readers - the create_scraper script can parse spelled out state names using the `us` module - but using the two-letter names is definitely clearer!

markfickett:
Oh, cool that it handles both!

PR #1076, opened 2019-02-22
remove links to parents in schema

terinpw:


jessex:
What's the context for this one?

terinpw:
@jessex - we wanted to remove these because it's a bit of weird structure to have children point back to their parent entities; and we don't currently have any use case for this bidirectional link (unless you have one in mind). 

From talking with @zdg2102, we were planning on removing these early on, but never got around to it until now.

jessex:
Thanks, @terinpw! So we added these in the beginning just because we thought the bidirectional link might eventually be useful, but a use case never materialized? I don't have a use case myself. We can add those links back in the future if the use case shows up.

terinpw:
@jessex - I believe we originally created the links because they were present in the example (in SQL Alchemy docs) that we were referencing.

But yeah, it's easy enough to add back if we need.

Issue #1055, opened 2019-02-21
Batch up all writes for a region and write only if enough tasks pass


arian487:
We want to write to the database if and only if a large enough percentage of the tasks pass AND there were no protected class failures.

- [x] Store metadata for a scrape session and accumulate failure/success metrics

- [x] Create a pubsub queue and publish serialized entities 

- [ ] Add an endpoint which runs at the end of a scrape session and writes data from the queue based on the failure/success metrics of the metadata

PR #1079, opened 2019-02-25
Batch em all

arian487:
This is the first step to batching up requests.  This adds the code to actually do the batching in a message type we care about, as well as the changes in persistence to not fail fast and instead count the error types.

Note that we for now leave the actual pubsub message commented because we need to complete the second half which is the endpoint to actually subscribe to the queue and write everyhthing at once.  We add the task to the BatchMessage because tasks can fail and retry many times so we need to dedupe tasks so we don't double count failures, or count successes as failures.

https://github.com/Recidiviz/recidiviz-data/issues/1055


recidiviz/persistence/entity_matching.py
```diff
@@ -32,38 +32,66 @@
...
+        updated on the ingested person.
+        TODO: document how matches are determined in docstring or README.
+        Args:
+            session: (Session)
+            region: (str)
+            ingested_people: List[entities.Person]
+        """
+        with_external_ids = []
+        without_external_ids = []
+        self.db_people_ext = None
```
colincadams: I'm not a huge fan of keeping this state around in here, but I'll defer to @jamwalla and @zacharylawrence

recidiviz/ingest/scrape/base_scraper.py
```diff
@@ -154,88 +160,131 @@ def _generic_scrape(self, request: QueueRequest):
...
+                # batch_message = BatchMessage(
+                #     scraper_start_time=request.scraper_start_time,
+                #     task=task,
+                #     ingest_info=scraped_data.ingest_info
+                # )
+                # scrape_key = ScrapeKey(
+                #     self.region.region_code, request.scrape_type)
+                # TODO 1055: Actually publish when the endpoint is correctly
+                #   set up
+                # self.publish_batch_message(batch_message, scrape_key)
```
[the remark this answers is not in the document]
arian487: Done.  It turns out base_scraper doesn't even have a test for it which is bad...I've added a todo to write this in a separate PR.

https://github.com/Recidiviz/recidiviz-data/issues/1086

recidiviz/ingest/scrape/base_scraper.py
```diff
@@ -154,88 +160,132 @@ def _generic_scrape(self, request: QueueRequest):
...
+            scraper_errors = [
+                ScraperGetMoreTasksError,
+                ScraperPopulateDataError,
+                ScraperPopulateDataError
+            ]
+            # If we caught some other unexpected error, we want to catch it
+            # and remap it to our catch all scraper error.
+            error_type = type(e)
+            if error_type not in scraper_errors:
+                e = ScraperError(str(e))
```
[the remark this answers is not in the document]
arian487: I onestly don't remember why I remapped the error, removed.

recidiviz/ingest/scrape/base_scraper.py
```diff
@@ -154,88 +160,131 @@ def _generic_scrape(self, request: QueueRequest):
...
+                # self.publish_batch_message(batch_message, scrape_key)
+                persistence.write(
+                    ingest_utils.convert_ingest_info_to_proto(
+                        scraped_data.ingest_info), metadata)
+            return None
+        except Exception as e:
+            scraper_errors = [
+                ScraperGetMoreTasksError,
+                ScraperPopulateDataError,
+                ScraperPopulateDataError
```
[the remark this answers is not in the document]
arian487: This is removed now.

recidiviz/ingest/scrape/base_scraper.py
```diff
@@ -154,88 +160,132 @@ def _generic_scrape(self, request: QueueRequest):
...
+            ]
+            # If we caught some other unexpected error, we want to catch it
+            # and remap it to our catch all scraper error.
+            error_type = type(e)
```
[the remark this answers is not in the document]
arian487: Removed

recidiviz/ingest/scrape/base_scraper.py
```diff
@@ -324,3 +374,19 @@ def get_initial_task(self) -> Task:
             task_type=constants.TaskType.INITIAL_AND_MORE,
             endpoint=self.get_region().base_url,
         )
+
+    def publish_batch_message(
```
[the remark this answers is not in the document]
arian487: Agreed, I've rewrote this into a batch_persistence file

recidiviz/ingest/scrape/base_scraper.py
```diff
@@ -324,3 +374,19 @@ def get_initial_task(self) -> Task:
...
+            self, batch_message: BatchMessage, scrape_key: ScrapeKey):
+        """Publishes the ingest info BatchMessage.
+
+        Args:
+            batch_message: Tahe BatchMessage to publish on the queue.
+            scrape_key: The ScrapeKey of the region
+        """
+        def inner():
+            serialized = batch_message.to_serializable()
```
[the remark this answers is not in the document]
arian487: Done

recidiviz/ingest/scrape/task_params.py
```diff
@@ -113,3 +113,33 @@ def to_serializable(self):
     @classmethod
     def from_serializable(cls, serializable):
         return cattr.structure(serializable, cls)
+
+
+@attr.s(frozen=True)
+class BatchMessage:
```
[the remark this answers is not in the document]
arian487: Done

recidiviz/ingest/scrape/base_scraper.py
```diff
@@ -154,88 +160,123 @@ def _generic_scrape(self, request: QueueRequest):
...
+            if self.should_scrape_data(task.task_type):
+                # If we want to scrape data, we should either create an
+                # ingest_info object or get the one that already exists.
+                logging.info('Scraping data for %s and endpoint: %s',
+                             self.region.region_code, task.endpoint)
+                try:
+                    scraped_data = self.populate_data(
+                        content, task, request.ingest_info or IngestInfo())
+                except Exception as e:
+                    raise ScraperPopulateDataError(str(e))
```
colincadams: same as above

recidiviz/ingest/scrape/base_scraper.py
```diff
@@ -154,88 +160,123 @@ def _generic_scrape(self, request: QueueRequest):
...
+
+                # Only send along ingest info if it will not be persisted now.
+                ingest_info_to_send = None
+                if scraped_data is not None and not scraped_data.persist:
+                    ingest_info_to_send = scraped_data.ingest_info
+
+                try:
+                    next_tasks = self.get_more_tasks(content, task)
+                except Exception as e:
+                    raise ScraperGetMoreTasksError(str(e))
```
colincadams: same as above

recidiviz/ingest/scrape/base_scraper.py
```diff
@@ -154,88 +160,123 @@ def _generic_scrape(self, request: QueueRequest):
...
+                    logging.info(scraped_data.ingest_info.people[i])
+                logging.info('Last seen time of person being set as: %s',
+                             request.scraper_start_time)
+                metadata = IngestMetadata(self.region.region_code,
+                                          request.scraper_start_time,
+                                          self.get_enum_overrides())
+                if self.BATCH_WRITES:
+                    scrape_key = ScrapeKey(
+                        self.region.region_code, request.scrape_type)
+                    # TODO 1055: Actually publish when the endpoint is correctly
```
colincadams: nit: this comment is no longer necessary

recidiviz/ingest/scrape/base_scraper.py
```diff
@@ -154,88 +160,123 @@ def _generic_scrape(self, request: QueueRequest):
...
+                        task=task,
+                        scrape_key=scrape_key,
+                    )
+                else:
+                    persistence.write(
+                        ingest_utils.convert_ingest_info_to_proto(
+                            scraped_data.ingest_info), metadata)
+            return None
+        except Exception as e:
+            if self.USE_PUBSUB:
```
colincadams: `self.USE_PUBSUB` -> `self.BATCH_WRITES`?

recidiviz/ingest/scrape/errors.py
```diff
@@ -18,4 +18,16 @@
...
+class ScraperGetMoreTasksError(Exception):
+    """Raised when the get_more_tasks of a scraper fails."""
+
+
+class ScraperPopulateDataError(Exception):
+    """Raised when the populate_data of a scraper fails."""
+
+
+class ScraperFetchError(Exception):
+    """Raised when a transient error occurs in fetching a webpage."""
```
colincadams: nit: do we know if the error is transient?

recidiviz/ingest/scrape/base_scraper.py
```diff
@@ -154,88 +160,123 @@ def _generic_scrape(self, request: QueueRequest):
...
+                # default value of None if this scraper doesn't set it.
+                try:
+                    content, cookies = self._fetch_content(
+                        task.endpoint, task.response_type, headers=task.headers,
+                        cookies=task.cookies, params=task.params,
+                        post_data=post_data, json_data=task.json)
+                    if content == -1:
+                        return -1
+                except Exception as e:
+                    raise ScraperFetchError(str(e))
```
colincadams: `raise ScraperFetchError from e`?

recidiviz/ingest/scrape/base_scraper.py
```diff
@@ -154,88 +160,123 @@ def _generic_scrape(self, request: QueueRequest):
...
+
+                # We always fetch some content before doing anything.
+                # Note that we use get here for the post_data to return a
+                # default value of None if this scraper doesn't set it.
+                try:
+                    content, cookies = self._fetch_content(
+                        task.endpoint, task.response_type, headers=task.headers,
+                        cookies=task.cookies, params=task.params,
+                        post_data=post_data, json_data=task.json)
+                    if content == -1:
```
colincadams: `_fetch_content` catches some of its errors and returns `-1` instead. We either need to fix that, or raise an exception here when `content == -1`. I'd opt for fixing it in a separate PR and then merging from master if you're okay with that!

recidiviz/persistence/persistence.py
```diff
@@ -136,15 +140,67 @@ def _should_persist():
...
+
+
+def _abort_or_continue(
+        total_people,
+        enum_parsing_errors=0,
+        entity_matching_errors=0,
+        protected_class_errors=0,
+        data_validation_errors=0):
+    # TODO: fill this in with actual logic
+    if protected_class_errors > 0:
```
colincadams: nit: `> 0` is unnecessary

recidiviz/ingest/scrape/base_scraper.py
colincadams: jk fixed

recidiviz/ingest/scrape/errors.py
arian487: Removed.

recidiviz/ingest/scrape/base_scraper.py
arian487: Removed

recidiviz/ingest/scrape/base_scraper.py
arian487: Done.

recidiviz/ingest/scrape/base_scraper.py
arian487: Done.

recidiviz/ingest/scrape/base_scraper.py
arian487: Sounds good

recidiviz/persistence/persistence.py
arian487: Done.

recidiviz/ingest/scrape/base_scraper.py
arian487: Added traceback information

recidiviz/persistence/entity_matching.py
jamwalla: I think there's use for this as a class, but I kind of agree. What could we get out of this class by just providing the session and region on init? I think since `match_all` is only called by tests we could restructure this.

What if we could initialize an `EntityMatcher(Session(), REGION)` that performed a read of all open bookings on init (*all*, where we used to only read open bookings that matched the pool of names) and then just matches ingested entities one at a time as they come in? This would definitely require refactoring @terinpw and I would do separately.

recidiviz/persistence/persistence.py
```diff
@@ -136,15 +140,67 @@ def _should_persist():
...
 
+def _convert_and_count_errors(ingest_info, metadata):
+    people = []
+    protected_class_errors = 0
+    enum_parsing_errors = 0
+    ii_converter = converter.get_converter(ingest_info, metadata)
+    while ingest_info.people:
+        try:
+            people.append(ii_converter.convert_and_pop())
+        except Exception as e:
```
jamwalla: could we subclass our errors so that we can catch exceptions more narrowly in this file? I imagine everything we're thinking about in this file could be either a `ScraperError`, `PersistenceError` or `EntityMatchingError`. (Maybe PersistenceErrors include `ConverterError` and `EnumParsingError`).

recidiviz/persistence/persistence.py
```diff
@@ -136,15 +140,67 @@ def _should_persist():
...
+            logging.error(str(e))
+            if e.entity_type in PROTECTED_CLASSES:
```
jamwalla: I think this will raise an `AttributeError` if `e` is one of the `ValueError`s raised by the converter, for example if a date can't be parsed.

recidiviz/persistence/entity_matching.py
arian487: Discussed offline, but we can think about moving the error counting responsibility to the separate classes (converter, entity_matching) in a separate PR.

recidiviz/common/constants/person.py
```diff
@@ -57,6 +57,9 @@ def _get_default_map():
         return _ETHNICITY_MAP
 
 
+PROTECTED_CLASSES = [Race, Ethnicity, Gender]
```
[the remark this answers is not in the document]
arian487: Done.

recidiviz/ingest/scrape/base_scraper.py
```diff
@@ -154,88 +160,122 @@ def _generic_scrape(self, request: QueueRequest):
...
+            scraped_data = None
+            if self.should_scrape_data(task.task_type):
+                # If we want to scrape data, we should either create an
+                # ingest_info object or get the one that already exists.
+                logging.info('Scraping data for %s and endpoint: %s',
+                             self.region.region_code, task.endpoint)
+                try:
+                    scraped_data = self.populate_data(
+                        content, task, request.ingest_info or IngestInfo())
+                except Exception as e:
```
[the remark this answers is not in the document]
arian487: Awesome!  Done.

recidiviz/persistence/entity_matching.py
```diff
@@ -32,38 +32,66 @@
...
+    def __init__(
+            self, session: Session,
+            region: str, ingested_people: List[entities.Person]):
+        """
+        Finds all people in the given |region| and database |session| and
+        attempts to match them to the |ingested_people|. For any ingested
+        person, if a matching person exists in the database, the primary key is
+        updated on the ingested person.
+        TODO: document how matches are determined in docstring or README.
+        Args:
```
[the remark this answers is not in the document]
arian487: Done.

recidiviz/persistence/persistence.py
arian487: This is related to the counter re-org that we talked about.  I do think something needs to be doing a generic catch all because you can always get unexpected errors and we don't want those breaking our entire loop.

recidiviz/persistence/entity_matching.py
```diff
@@ -32,38 +32,66 @@
...
+        if with_external_ids:
+            self.db_people_ext = database.read_people_by_external_ids(
+                session, region, with_external_ids)
+            self.ingested_people_ext = with_external_ids
+        if without_external_ids:
+            self.db_people_no_ext = database.read_people_with_open_bookings(
+                session, region, without_external_ids)
+            self.ingested_people_no_ext = without_external_ids
+
+    def match_and_pop(self):
```
terinpw: Chatted offline with @arian487 @jamwalla and @zacharylawrence, but I don't think we should expose an API that allows the caller to call match on an incomplete list of people. Entity matching is only guaranteed to be correct if we've matched everyone ingested (because we check that no 2 ingested people match the same db person and vice versa), so if the caller has the option to stop midway through they might be incorrect in assuming entity matching worked. 

We should keep that responsibility in entity matching itself.

recidiviz/persistence/persistence.py
arian487: Fixed.

recidiviz/persistence/entity_matching.py
arian487: Sounds good, lets do it in a separate PR.

recidiviz/persistence/batch_persistence.py
```diff
@@ -0,0 +1,109 @@
...
+    def from_serializable(cls, serializable):
+        return cattr.structure(serializable, cls)
+
+
+def _publish_batch_message(
+        batch_message: BatchMessage, scrape_key: ScrapeKey):
+    """Publishes the ingest info BatchMessage.
+
+    Args:
+        batch_message: Tahe BatchMessage to publish on the queue.
```
jessex: The

recidiviz/persistence/batch_persistence.py
```diff
@@ -0,0 +1,109 @@
...
+    # The task which published this message.  We use this to dedupe messages
+    # That failed some number of times before finally passing, or to not
+    # double count tasks that failed more than once.
+    task: Task = attr.ib()
+
+    # The time at which this scraper was started, used to associate the entities
+    # ingested during throughout the scrape
+    scraper_start_time: Optional[datetime.datetime] = attr.ib()
+
+    # In the case that a person's information is spread across multiple pages
```
jessex: Is this the only scenario when `IngestInfo` is specified? Is it not also specified for single-page scrapers where you are simply batching up the successful scrape to be written later?

recidiviz/persistence/entity_matching.py
jessex: It sounds like a couple of points in this class specifically are being delegated to follow-on PRs. Can we file issues and slap on some TODOs?

recidiviz/persistence/persistence.py
```diff
@@ -29,14 +29,16 @@
 from recidiviz.common.constants.bond import BondStatus
 from recidiviz.common.constants.booking import CustodyStatus
 from recidiviz.common.constants.charge import ChargeStatus
+from recidiviz.common.constants.entity_enum import EnumParsingError
+from recidiviz.common.constants.person import PROTECTED_CLASSES
```
jessex: Nit but this import is no longer in alphabetical order.

recidiviz/persistence/persistence.py
```diff
@@ -136,15 +141,70 @@ def _should_persist():
...
+        protected_class_errors=0,
+        data_validation_errors=0):
+    # TODO: fill this in with actual logic
+    if protected_class_errors:
+        raise PersistenceError(
+            'Aborting because there was an error regarding a protected class')
+    if (enum_parsing_errors + entity_matching_errors +
+            data_validation_errors) / total_people >= ERROR_THRESHOLD:
+        raise PersistenceError(
+            'Aborting because we exceeded the error threshold')
```
jessex: We should include the relevant numbers in this error message.

recidiviz/persistence/persistence.py
```diff
@@ -157,28 +217,50 @@ def write(ingest_info, metadata):
...
+    total_people = len(ingest_info.people)
     with monitoring.measurements(mtags) as measurements:
-        people = converter.convert(ingest_info, metadata)
-        validate_one_open_booking(people)
+
+        # Convert the people one at a time and count the errors as they happen.
+        people, enum_parsing_errors, protected_class_errors =\
+            _convert_and_count_errors(ingest_info, metadata)
+        people, data_validation_errors = validate_one_open_booking(people)
         logging.info(
```
jessex: Should we delete this log statement now? Seems redundant to the new one just below.

recidiviz/persistence/persistence.py
```diff
@@ -157,28 +217,50 @@ def write(ingest_info, metadata):
...
         persisted = False
         session = Session()
         try:
             logging.info('Starting entity matching')
-            entity_matching.match_entities(session, metadata.region, people)
-            logging.info('Successfully completed entity matching')
+            entity_matching_errors = _entity_match_and_count_errors(
+                session, metadata.region, people)
+            logging.info(
+                'Successfully completed entity matching with %s errors',
```
jessex: It's not necessarily successful anymore, right? Now it could fail completely, and this message would give false confidence to anyone traversing the logs.

recidiviz/persistence/entity_matching.py
arian487: Done.

https://github.com/Recidiviz/recidiviz-data/issues/1094

recidiviz/persistence/batch_persistence.py
arian487: Fixed.

recidiviz/persistence/batch_persistence.py
arian487: Ah, this comment is out of date, fixed.

recidiviz/persistence/persistence.py
arian487: Fixed.

recidiviz/persistence/persistence.py
arian487: Done.

recidiviz/persistence/persistence.py
arian487: Done.

recidiviz/persistence/persistence.py
arian487: I disagree, at least for this PR entity matching does not and should not care about the error threshold, and by that metric is successfully ran.  The parent later on decides if the number of errors was acceptable.  But I suppose to be more clear I'll just remove successfully :P

jessex: Fair! Good compromise.

recidiviz/persistence/batch_persistence.py
```diff
@@ -0,0 +1,109 @@
...
+    This is the object that is serialized and put on the pubsub queue.
+    """
+    # The task which published this message.  We use this to dedupe messages
+    # That failed some number of times before finally passing, or to not
+    # double count tasks that failed more than once.
+    task: Task = attr.ib()
+
+    # The time at which this scraper was started, used to associate the entities
+    # ingested during throughout the scrape
+    scraper_start_time: Optional[datetime.datetime] = attr.ib()
```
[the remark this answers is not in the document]
arian487: When there is an error, we don't care to add this.

recidiviz/persistence/batch_persistence.py
```diff
@@ -0,0 +1,109 @@
...
+
+def write_error(error, task, scrape_key):
+    """Batches up the errors using pubsub
+
+    Args:
+        error: (str) the error that the scraper failed with.
+        task: (Task) the task which queued up this write
+        scrape_key: (ScrapeKey) information on the region
+    """
+    batch_message = BatchMessage(
```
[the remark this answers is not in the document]
arian487: In the batch_endpoint PR i've tested this and it doesn't seem to complain.

recidiviz/persistence/batch_persistence.py
```diff
@@ -0,0 +1,109 @@
...
+
+    # In the case that a person's information is spread across multiple pages
+    # this is used to pass it along to the next page.
+    ingest_info: Optional[IngestInfo] = attr.ib(default=None)
+
+    # The error type of the task if it ended in failure.
+    error: Optional[str] = attr.ib(default=None)
+
+    def to_serializable(self):
+        return cattr.unstructure(self)
```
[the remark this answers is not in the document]
arian487: Thats a good question, it seems like it works because it was imported from task_params, I moved this to a top level init.

recidiviz/persistence/persistence.py
```diff
@@ -58,6 +59,9 @@
 monitoring.register_views([people_persisted_view, errors_persisted_view])
 
 
+ERROR_THRESHOLD = 0.5
```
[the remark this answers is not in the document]
arian487: Not yet, I forgot about this I'll set up something this week.  We can have this as a placeholder for now, but if we want variable cutoffs that would require a completely different approach than what I have now.  I think we agreed that it would be pretty complicated to handle that stuff on export time, hence this approach.

arian487:
Everything passes now, please take a look.

recidiviz/__init__.py
```diff
@@ -43,3 +47,16 @@
     db_engine = sqlalchemy.create_engine(sqlalchemy_url)
     Base.metadata.create_all(db_engine)
     Session.configure(bind=db_engine)
+
+cattr.register_unstructure_hook(datetime.datetime,
```
zacharylawrence: Nit: Do you mind adding a quick comment about why you needed to move these here?

recidiviz/persistence/batch_persistence.py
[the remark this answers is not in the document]
zacharylawrence: Does this mean there's really 2 types of messages we want to put on the queue? One of `(task, scraper_start_time, ingest_info)` and one of `(task, error)`?

recidiviz/persistence/batch_persistence.py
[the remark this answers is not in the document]
zacharylawrence: SGTM!

recidiviz/persistence/converter/converter.py
```diff
@@ -38,6 +38,15 @@ def convert(ingest_info, metadata):
     return _Converter(ingest_info, metadata).convert()
 
 
+def get_converter(ingest_info, metadata):
```
zacharylawrence: Should this converter be refactored to just expose `_Converter` directly? Maybe it could have an API that lets you call `.convert(person_id)` or something that lets you convert one person at a time and catch errors as you go?

```
converter = Converter(ingest_info)
for person in ingest_info.people:
  try:
    entity_person = converter.convert(person)
  catch ...
```

Thoughts on this?

recidiviz/__init__.py
arian487: Done.

recidiviz/persistence/converter/converter.py
arian487: I think that could work, but I also think that would become moot if we go with the other API of letting the respective classes/modules handle their own error counting and return that to the caller.  I have an issue out for that (https://github.com/Recidiviz/recidiviz-data/issues/1094) so lets tackle that when a final decision is made on that.

recidiviz/persistence/batch_persistence.py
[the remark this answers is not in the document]
arian487: Pretty much.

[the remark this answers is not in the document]
colincadams: `scraper_start_time` seems like it could be useful to keep either way, especially if we want to look at persisting messages between scrapes in the future. Is there any reason you think it shouldn't be set?

recidiviz/persistence/converter/converter.py
zacharylawrence: Even if we plan to restructure this code in the future, I think it's still worth making sure we don't return a private class now. Thoughts on removing this method and just making `_Converter` public?

recidiviz/persistence/batch_persistence.py
```diff
@@ -0,0 +1,108 @@
...
+    @classmethod
+    def from_serializable(cls, serializable):
+        return cattr.structure(serializable, cls)
+
+
+def _publish_batch_message(
+        batch_message: BatchMessage, scrape_key: ScrapeKey):
+    """Publishes the ingest info BatchMessage.
+
+    Args:
```
zacharylawrence: I think we can probably remove these Args pydoc comments since that information is already in the type information? Is this something we can do across our codebase?

recidiviz/persistence/batch_persistence.py
```diff
@@ -0,0 +1,108 @@
...
+        batch_message: The BatchMessage to publish on the queue.
+        scrape_key: The ScrapeKey of the region
+    """
+    def inner():
```
zacharylawrence: Is there a better name we could use for this? If not, I think `inner` is fine since it seems simple enough.

recidiviz/persistence/batch_persistence.py
```diff
@@ -0,0 +1,108 @@
...
+        serialized = batch_message.to_serializable()
+        return pubsub_helper.get_publisher().publish(
+            pubsub_helper.get_topic_path(scrape_key,
+                                         pubsub_type=PUBSUB_TYPE),
+            data=json.dumps(serialized).encode())
+    pubsub_helper.retry_with_create(scrape_key, inner, PUBSUB_TYPE)
+
+
+def write(ingest_info, scraper_start_time, task, scrape_key):
```
zacharylawrence: Could you add typing information to this method and below?

recidiviz/persistence/converter/converter.py
```diff
@@ -55,6 +64,9 @@ def __init__(self, ingest_info, metadata):
     def convert(self):
         return [self._convert_person(p) for p in self.ingest_info.people]
 
+    def convert_and_pop(self):
+        return self._convert_person(self.ingest_info.people.pop())
```
zacharylawrence: Does this actually mutate the ingest_info object? Should we make a copy of it in the `__init__`?

recidiviz/persistence/persistence.py
```diff
@@ -136,15 +141,79 @@ def _should_persist():
                 strtobool((os.environ.get('PERSIST_LOCALLY', 'false'))))
 
 
+# TODO 1094: Consider moving the error counting to the respective converter
+#   and entity matching code.
+
+
+def _convert_and_count_errors(ingest_info, metadata):
```
zacharylawrence: Could you add typing information here too? I think it'll be especially helpful since we're returning a tuple of multiple items.

recidiviz/persistence/batch_persistence.py

recidiviz/persistence/batch_persistence.py

recidiviz/persistence/persistence.py
arian487: Done.

recidiviz/persistence/batch_persistence.py

recidiviz/persistence/converter/converter.py
arian487: Done.

recidiviz/persistence/persistence.py
```diff
@@ -136,15 +141,79 @@ def _should_persist():
...
+    return entity_matching_errors
+
+
+def _abort_or_continue(
+        total_people,
+        enum_parsing_errors=0,
+        entity_matching_errors=0,
+        protected_class_errors=0,
+        data_validation_errors=0):
+    # TODO: fill this in with actual logic
```
jessex: What's the deal with this one? Outdated?

recidiviz/persistence/converter/converter.py
arian487: Done.

recidiviz/persistence/persistence.py
zacharylawrence: @arian487 Do we still need this TODO?

recidiviz/persistence/converter/converter.py
```diff
@@ -55,6 +56,14 @@ def __init__(self, ingest_info, metadata):
     def convert(self):
```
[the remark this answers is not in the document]
arian487: Yea I think that makes sense, done.

recidiviz/persistence/persistence.py
arian487: I think I'll leave it because the logic is not finalized.  I reworded it though.

PR #1080, opened 2019-02-25
UsCaKingsScraper

jamwalla:


recidiviz/ingest/extractor/html_data_extractor.py
```diff
@@ -304,7 +304,9 @@ def _below(self, cell: HtmlElement) -> Iterator[HtmlElement]:
         grand_parent = parent.getparent()
         # If |cell| is inside a <thead>, the |next_row| is inside a <tbody>.
         if grand_parent is not None and grand_parent.tag == 'thead':
-            next_row = grand_parent.getnext()[0]
+            tbody = grand_parent.getnext()
+            if tbody and tbody.getchildren():
```
colincadams: here if there are no rows in the `<tbody>` we want `next_row` to get set to `None` correct? If so I would do `next_row = next(tbody.iter_children(), None)`

jamwalla: yes amazing!!

PR #1081, opened 2019-02-25
Us ky franklin

olleicua:


recidiviz/tests/ingest/scrape/regions/us_ky_franklin/.#us_ky_franklin_scraper_test.py
```diff
@@ -0,0 +1 @@
+root@e6ff5e5ac631.4785
```
colincadams: what editor do you use? I assume this file is one that it generated? If so let us know so we can set up our repo to ignore files of this form.

recidiviz/ingest/scrape/regions/us_ky_franklin/manifest.yaml
```diff
@@ -0,0 +1,6 @@
+agency_name: $agency
```
colincadams: set the agency

recidiviz/tests/ingest/scrape/regions/us_ky_franklin/.#us_ky_franklin_scraper_test.py
olleicua: my bad, those come from emacs. I removed them.

colincadams: great thanks! I just sent out #1083 so hopefully they'll all be ignored in the future!

recidiviz/ingest/scrape/regions/us_ky_franklin/manifest.yaml
olleicua: agency set

olleicua:
ping

PR #1082, opened 2019-02-25
converter_utils.parse_datetime: call dateparser.parse in english

jamwalla:
dateparser does language detection and so (obviously) had the following behavior:
```
>>> dateparser.parse('NA')
datetime.datetime(2019, 2, 25, 17, 24, 32, 444917)
>>> dateparser.parse('N/A')
datetime.datetime(2019, 2, 21, 0, 0)
```
We should assume we're scraping dates in english, for now.


recidiviz/persistence/converter/converter_utils.py
```diff
@@ -72,7 +72,7 @@ def parse_datetime(date_string):
     """
     if date_string == '' or date_string.isspace():
         return None
-    parsed_date = dateparser.parse(date_string)
+    parsed_date = dateparser.parse(date_string, languages=['en'])
```
colincadams: should we move this method out to a more global util, so that it can then be called from aggregate_ingest as well?

jamwalla: Probably, but I filed #1084 for now because this is more fraught than I thought (`converter_utils.parse_days` should do more/better than dateparser does).

Issue #974, opened 2019-02-12
We should detect when a scrape session for a region is complete


ohinds:

colincadams:
One possible idea for how to handle this:

We could keep a `pending_task_count` on the `session` object in datastore, and increment it by `tasks_added - 1` at the end of each scrape. When it hits zero we can call out to cloud composer that the scrape has finished, or trigger the next step another way.

I'm not sure that we can fully guarantee consistency between datastore and cloud tasks this way, e.g. our instance dies right after we decrement the count but before we return 200 so cloud tasks retries the task. Worth thinking about further whether there is a good solution to handle that.

jessex:
Could we have a background process that polls a particular subscription path to check for the existence of messages for a given region? We'd want to be smart about it so we weren't querying for every single region repeatedly around the clock. For example, we could have a single background process that checks every, say, 10 minutes to find the list of regions that are currently active and compares it against the list from the previous check. If something dropped out then we can reason that it's no longer active, though there could be edge cases to think through there depending on the exact delivery semantics of Pub/Sub.

colincadams:
My understanding of the docket (which admittedly isn't great even though I migrated it to pub/sub) is that for our typical background scrape, an item gets added to the docket on scraper start. This item is then immediately pulled and added to the session, but not acked, and then the scrape starts. AFAIK this docket item is never acked so is returned to the docket after 5 mins, but we also never pull from the docket again. Next time that scraper is started the docket is purged and a new item is added.

I think that part of the core of this problem is that we don't know when to ack the docket item either. I don't totally remember how us_ny used to use the docket, if we went back to a model like that would that make the solution easier?

colincadams:
Now we could potentially query the cloud task queue for a given region to see if a scrape is done, assuming that query includes currently leased tasks. But this doesn't work in the model of shared queues without inspecting each message. I can look into that to see if it is possible.

jessex:
>AFAIK this docket item is never acked so is returned to the docket after 5 mins

Never acking the docket item would be news to me. At least when the docket was a pull-queue, I was under the impression that it was acked by virtue of the task that pulled the item from the queue returning a 2XX-space response. I assumed that we had an equivalent mechanism in the migration to Pub/Sub?

>but we also never pull from the docket again.

[...]

PR #1092, opened 2019-02-27
Batch endpoint

### LINKED ISSUE #1055: Batch up all writes for a region and write only if enough tasks pass

arian487:
Adds an endpoint for persisting the batched writes.  This should be queued by the scrape tracker and is invoked once a scraper is complete.  Additionally in this PR, we generate a UUID for the generated id instead of using the memory address because we need a stronger guarantee that the generated ids don't collide when they get serialized onto the pubsub queue.

Fixes #1055


recidiviz/ingest/scrape/ingest_utils.py
```diff
@@ -285,3 +288,13 @@ def ingest_info_from_serializable(serializable: Dict[Any, Any]) \
         -> ingest_info.IngestInfo:
     proto = json_format.ParseDict(serializable, ingest_info_pb2.IngestInfo())
     return convert_proto_to_ingest_info(proto)
+
+
+def append_to_proto(base, proto_to_append):
```
jessex: Does this need to be here? I don't like extracting a public method out into a utility/helper module unless there are multiple uses for the logic, because now we need to be testing and documenting it and making sure it makes sense within the context of the utility/helper class, which is now a bit larger. Seems easier all around to keep it private down in `batch_persistence`.

recidiviz/persistence/batch_persistence.py
```diff
@@ -107,3 +173,39 @@ def write_error(error, task, scrape_key):
         task=task,
     )
     _publish_batch_message(batch_message, scrape_key)
+
+
+@batch_blueprint.route('/read_and_persist', methods=['POST'])
+@authenticate_request
+def read_and_persist():
```
jessex: I don't think this should be down in the `persistence` package. It's an API endpoint. This blurs the lines of separation between the layers in the app more than is necessary/useful.

recidiviz/persistence/batch_persistence.py
```diff
@@ -107,3 +173,39 @@ def write_error(error, task, scrape_key):
...
+    """
+    json_data = request.get_data(as_text=True)
+    data = json.loads(json_data)
+
+    region = data['region']
+    scrape_type = data['scrape_type']
+    scraper_start_time = data['scraper_start_time']
+
+    scraper = regions.get_region(region).get_scraper()
+    overrides = scraper.get_enum_overrides()
```
jessex: This feels ripe for a lil' convenience method. `region.get_enum_overrides()` feels more readable/direct to me than instantiating a scraper instance and then retrieving overrides from there as though it's stateful to the scraper itself.

recidiviz/persistence/batch_persistence.py
```diff
@@ -107,3 +173,39 @@ def write_error(error, task, scrape_key):
...
+        logging.info('Too many scraper tasks failed(%s), aborting write',
+                     len(failed_tasks))
+        return '', HTTPStatus.OK
+
+    metadata = IngestMetadata(
+        region=region, last_seen_time=scraper_start_time,
+        enum_overrides=overrides)
+
+    persistence.write(proto, metadata)
+    # TODO: queue up infer release before returning
```
jessex: Does this kind of do an end-around around #974? CC @colincadams

recidiviz/persistence/batch_persistence.py
```diff
@@ -77,18 +83,78 @@ def inner():
...
+    for message in messages:
+        batch_message = BatchMessage.from_serializable(message.message.data)
+        # We do this because dicts are not hashable in python and we want to
+        # avoid an n2 operation to see which tasks have been seen previously
+        # which can be on the order of a million operations.
+        task_hash = hash(json.dumps(batch_message.task.to_serializable(),
+                                    sort_keys=True))
+        if not batch_message.error:
+            successful_tasks.add(task_hash)
+            if task_hash in failed_tasks:
```
jessex: This occurs when a scrape task failed, was retried, and then succeed?

recidiviz/ingest/scrape/ingest_utils.py
arian487: Done.

recidiviz/persistence/batch_persistence.py

recidiviz/persistence/batch_persistence.py
arian487: Yes, colin is currently working on a scraper tracker which will detect when a scrape is complete and then start the batching process (which will in turn start the infer release process)

recidiviz/persistence/batch_persistence.py
arian487: I'm not sure what you mean exactly?  This endpoint will not be called until all messages are published and there are no more tasks left so this race condition could not happen.  The state of the queue should in theory be complete by the team we go and read it.

jessex: Ah, I didn't know under what conditions the messages would be read from the PubSub, i.e. if they would only ever be accessed through your endpoint below, and also who would be clients of that endpoint. Do we need to ensure that the endpoint is not called until all scrapes for the session are complete? Given what you said about Colin's work, it sounds like that's what we're aiming for.

recidiviz/persistence/batch_persistence.py
```diff
@@ -89,3 +171,46 @@ def write_error(error: str, task: Task, scrape_key: ScrapeKey):
...
+    metadata = IngestMetadata(
+        region=region, last_seen_time=scraper_start_time,
+        enum_overrides=overrides)
+
+    persistence.write(proto, metadata)
+
+
+@batch_blueprint.route('/read_and_persist', methods=['POST'])
+@authenticate_request
+def read_and_persist():
```
jessex: I never saw a response to my previous comment about preferring not to have an API endpoint down in the persistence layer.

PR #1093, opened 2019-02-27
added 11 jailtracker counties in KY

olleicua:


recidiviz/ingest/scrape/regions/us_ky_hardin/us_ky_hardin_scraper.py
```diff
@@ -0,0 +1,44 @@
...
+"""Scraper implementation for us_ky_hardin."""
+
+from recidiviz.common.constants.enum_overrides import EnumOverrides
+from recidiviz.ingest.scrape.vendors.jailtracker.jailtracker_scraper import \
+    JailTrackerScraper
+
+
+class UsKyHardinScraper(JailTrackerScraper):
+    """Scraper implementation for us_ky_hardin."""
+    def __init__(self, _=None):
```
colincadams: these `_=None` can just be removed if you aren't using them. I'll send out a PR to remove it from the template as it is rarely used.

olleicua: I'm pretty sure I've seen argument errors when I do that.. possibly in the mypy linter.. I just recall that travis didn't pass.

jamwalla: Can you pull in the latest changes (#1095)?

PR #1099, opened 2019-02-27
separated new_with_defaults and builder into separate annotations instead of one joint class

terinpw:


recidiviz/common/attr_annotations.py
```diff
@@ -18,27 +18,48 @@
 import attr
 
 
-class BuildableAttr:
-    """Abstract base class for Attr objects that can be built with a Builder."""
+def default_attr(cls):
+    """Annotation to add method to attr class to create object with default
```
jessex: I was also in the habit of calling this an annotation because of my Java background, but these are actually called [_decorators_](https://github.com/gto76/python-cheatsheet/blob/master/README.md#decorator) in Python. ([Decorators vs Annotations](https://stackoverflow.com/questions/15347136/is-a-python-decorator-the-same-as-java-annotation-or-java-with-aspects))

terinpw: Great point, done!

terinpw:
Makes sense, renamed the file!

terinpw:
So just worked with @zacharylawrence and @colincadams - and neither mypy nor pycharm plays super well with these decorators. 

MyPy doesn't know about the attributes that are added by the decorators. This means whenever we use new_with_defaults or Builder, MyPy throws an error. 

Smaller issue - PyCharm can't find the method declarations that are added via decorator. This just means we can't jump into the implementation like we normally can. 

We were inclined to use two separate mixins instead of decorators, if only so that it plays nicely with MyPy. What do you think @jessex?

jessex:
That's totally fine by me. The main point is to make for easier development so let's follow that goal. :+1:

recidiviz/tests/common/buildable_attr_test.py
```diff
@@ -20,7 +20,7 @@
```
zacharylawrence: Should this test file also be renamed?

Also, I know this wasn't tested before, but should we also test `new_with_defaults`? I think there's no need to do this here since it's not a regression.

recidiviz/tests/common/buildable_attr_test.py
```diff
@@ -63,9 +63,5 @@ def testBuild_ExtraField_RaisesException(self):
         subject.not_a_real_field = "TEST_2"
 
         # Act + Assert
-        with self.assertRaises(FakeBuildableAttr.BuilderException):
+        with self.assertRaises(BuilderException):
             subject.build()
-
-    def testInstantiateAbstractBuildableAttr_RaisesException(self):
```
zacharylawrence: Why'd we need to delete this?

recidiviz/common/attr_mixins.py
```diff
@@ -18,22 +18,40 @@
 import attr
 
 
-class BuildableAttr:
-    """Abstract base class for Attr objects that can be built with a Builder."""
+class DefaultAttr:
```
zacharylawrence: Consider renaming to something else like `DefaultableAttr` but as you pointed out offline without the "Defaultable Loan" connotation.

recidiviz/common/attr_mixins.py
```diff
@@ -18,22 +18,40 @@
...
+
+            # Ignore Factories to allow them to render into a default value
+            if isinstance(default, attr.Factory):
+                continue
+
+            kwargs[field] = None if default is attr.NOTHING else default
 
-    # Consider BuildableAttr abstract and only allow instantiating subclasses
-    # pylint: disable=unused-argument
-    def __new__(cls, *args, **kwargs):
```
zacharylawrence: Is there a reason to remove this? Should it also exist on `DefaultAttr`?

terinpw: Put it back, and updated to check that cls is an attr

terinpw: Yup and added.

recidiviz/common/attr_mixins.py
terinpw: Changed!

recidiviz/tests/common/buildable_attr_test.py
terinpw: Yup renamed.

recidiviz/common/attr_mixins.py
```diff
@@ -18,9 +18,15 @@
 import attr
 
 
-class DefaultAttr:
+class DefaultableAttr:
     """Mixin to add method to attr class that creates default object"""
 
+    # DefaultAttr can only be mixed in with an attr class
```
zacharylawrence: I think a typo here: `DefaultableAttr`

recidiviz/common/attr_mixins.py
```diff
@@ -18,9 +18,15 @@
...
+    def __new__(cls, *_args, **_kwargs):
+        if not attr.has(cls):
+            raise Exception('Base class must be an attr class')
```
zacharylawrence: I think `Base class` here means this class. Maybe you mean `Parent class`?

PR #1105, opened 2019-02-27
Entity match charges with children if possible

### LINKED ISSUE #1012: charge matching
Because we are ordering db_charges with more information first, we ensure that charges with less information (fewer children) never take the spot of a better match (matches children as well as own info).
```
order db_charges: charges with more children are processed first
for db_charge in db_charges:
  potential_matches_without_comparing_children = get them()
  potential_matches_with_comparing_children = get them()
  if potential_matches_without_comparing_children > 1:
    if potential_matches_with_comparing_children:
       match db_charge to first potential_matches_with_comparing_children
    else:
      match db_charge to first potential_matches_without_comparing_children
```
  cc @terinpw

terinpw:
Fixes #1012 

When possible, we should attempt to match charges based on charge fields and sentence/bond fields. Only if we find no matches while comparing sentence/bond fields do we attempt to match based on the charge fields alone. 

Co Author - jamwalla


recidiviz/persistence/entity_matching.py
```diff
@@ -287,6 +274,10 @@ def _drop_hold(hold: entities.Hold):
         hold.status = HoldStatus.INFERRED_DROPPED
 
 
+def _charge_relationship_count(charge: entities.Charge):
```
jessex: Can you add a small comment explaining exactly what's being counted here? I think it's non-obvious.

recidiviz/persistence/entity_matching.py
```diff
@@ -296,15 +287,24 @@ def match_charges(
     |db_booking|, the primary key is updated on the ingested charge. All
     db charges that are not matched to an ingested charge are marked dropped and
     added to the |ingested_booking|.
-    Args:
-        ingested_booking: (entities.Booking)
-        db_booking: (entities.Booking)
+
+    Note: When matching charges, we attempt to match considering children (
+    bonds and sentences) as well as charge information. If we cannot match while
```
jessex: Can we add a bit more detail to this? For example, what happens if two charges look the same but have different bonds or sentences? What happens if they look the same, but the db charge has bonds or sentences and the ingested charge has none, or vice versa? I think we should very much err on the side of over-documenting entity matching, given how critical it is and how complex it is.

recidiviz/persistence/entity_matching.py
```diff
@@ -296,15 +287,24 @@ def match_charges(
...
+    comparing children, we match on just charge information.
     """
     dropped_charges = []
-    for db_charge in db_booking.charges:
-        ingested_charge = _get_next_available_match(db_charge,
-                                                    ingested_booking.charges,
-                                                    utils.is_charge_match)
+    sorted_db_charges = sorted(
```
zdg2102: Can you add a comment explaining the impact of sorting here?

recidiviz/tests/persistence/entity_matching_test.py
```diff
@@ -411,6 +411,93 @@ def test_matchCharges(self):
...
+        ingested_charge = attr.evolve(db_charge, charge_id=None, bond=None)
+
+        expected_matched_charge = attr.evolve(
+            ingested_charge, charge_id=db_charge.charge_id)
+
+        db_booking = entities.Booking.new_with_defaults(charges=[db_charge])
+        ingested_booking = entities.Booking.new_with_defaults(
+            charges=[ingested_charge])
+
+        # import ipdb; ipdb.set_trace()
```
zdg2102: Is this line supposed to be here?

recidiviz/tests/persistence/entity_matching_test.py
```diff
@@ -411,6 +411,93 @@ def test_matchCharges(self):
...
+            db_booking=db_booking, ingested_booking=ingested_booking)
+        self.assertCountEqual(ingested_booking.charges,
+                              [expected_matched_charge])
+
+    def test_matchCharges_disambiguateByChildren(self):
+        db_bond = entities.Bond.new_with_defaults(
+            bond_id=_ID, external_id=_EXTERNAL_ID_ANOTHER,
+            status=BondStatus.PENDING)
+        db_bond_another = entities.Bond.new_with_defaults(
+            bond_id=_ID_ANOTHER, external_id=_EXTERNAL_ID,
```
zdg2102: I might not be reading this right, but wouldn't these disambiguate by having different external IDs without even needing to look at their children?

terinpw: This is in the bond though, which is the child. So the charges are the same, so only when comparing the bonds do we know how to match.

recidiviz/tests/persistence/entity_matching_test.py
terinpw: Nah, removed it.

recidiviz/tests/persistence/entity_matching_test.py
zdg2102: Oh whoops... my bad

recidiviz/persistence/entity_matching.py
terinpw: Good idea, done

recidiviz/persistence/entity_matching.py
terinpw: Yeah great point. Can you take a look and let us know if you think the doc should be updated?

recidiviz/persistence/entity_matching.py
terinpw: Addressed this in the docstring above - I think that should suffice, but let us know if you think it's still not clear. 

Also updated the variable name to be more descriptive.

zdg2102: Makes sense to me

recidiviz/persistence/entity_matching.py
```diff
@@ -296,15 +288,55 @@ def match_charges(
...
+    be marked dropped.
+
+    Because we can match two charges whose children are not equal, we sort our
+    db_charges, attempting to match those with more children before those with
+    fewer or none.
+
+    Examples:
+        1. Two identical charges are in our DB, one with a bond (A) and one
+        without (B). The same data is scraped the next day so that we have
+        corresponding ingested bonds C (with a bond) and D (without a bond).
```
jessex: "bonds C" should be "charges C", no?

terinpw: Good catch!

PR #1107, opened 2019-02-27
Scraper for Ohio County, KY

sliuu:


recidiviz/ingest/scrape/regions/us_ky_ohio/us_ky_ohio_scraper.py
```diff
@@ -0,0 +1,68 @@
...
+        This value can either be text or a number. In either case, this method
+        should return the value as a string.
+        """
+        return "OHIO_COUNTY_KY"
+
+    def get_enum_overrides(self) -> EnumOverrides:
+        overrides_builder = super(UsKyOhioScraper,
+                                  self).get_enum_overrides().to_builder()
+        overrides_builder.ignore('V', ChargeClass)
+        overrides_builder.add("COURT ORDERED",
```
terinpw: same comment here for `external unknown`. I think several of these values should be mapped to different enums, but I'll let @ohinds comment on the exact mapping.

recidiviz/ingest/scrape/regions/us_ky_ohio/us_ky_ohio_scraper.py
```diff
@@ -0,0 +1,68 @@
...
+                              ChargeStatus.EXTERNAL_UNKNOWN)
+        overrides_builder.add("NO BOND",
```
terinpw: nit here and below - our codebase isn't consistent, but we're trying to move to using single quotes here.

recidiviz/ingest/scrape/regions/us_ky_mason/manifest.yaml
```diff
@@ -0,0 +1,6 @@
+agency_name: $agency
```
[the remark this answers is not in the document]
sliuu: done

recidiviz/ingest/scrape/regions/us_ky_ohio/us_ky_ohio_scraper.py
sliuu: done

PR #1114, opened 2019-02-28
make BaseScraper.get_more_tasks not abstract

jamwalla:
Some scrapers don't have more tasks, just a single `SCRAPE_DATA` task returned in `get_initial_task`. These scrapers shouldn't have to return more tasks, BaseScraper will raise a `NotImplementedError` if its own method is called.


recidiviz/tools/scraper_template/region_scraper.txt
```diff
@@ -31,6 +31,3 @@ class ${class_name}(BaseScraper):
     def populate_data(self, content, task: Task,
                       ingest_info: IngestInfo) -> Optional[ScrapedData]:
         pass
-
-    def get_more_tasks(self, content, task: Task) -> List[Task]:
```
colincadams: since most scrapers need this do you think we should leave it here, with a comment that it can be deleted if not used? I could be convinced either way, not sure what would be easiest for a scraper writer.

jamwalla: no, you're totally right! we should keep this.

PR #1115, opened 2019-02-28
Add support for name suffix

colincadams:
Happy for this to be a discussion, not sure if everyone agrees this should be supported this way?


recidiviz/persistence/converter/person.py
```diff
@@ -54,14 +54,17 @@ def _parse_name(proto) -> Optional[str]:
...
+    if full_name and (given_names or middle_names or surname or name_suffix):
+        raise ValueError(
+            'Cannot have full_name and surname/middle/given_names/name_suffix')
 
     if full_name:
         return full_name
     if given_names or middle_names or surname:
-        return _to_csv([given_names, middle_names, surname])
+        # skip this if only `name_suffix` was set.
+        return _to_csv([given_names, middle_names, surname, name_suffix])
```
colincadams: If we submit this as is, it will break entity matching on names as all existing names will change to have a ',' at the end. I could instead hack around this, to skip the `suffix` entirely if it's not set, therefore skipping the comma, but that could allow for collisions where one of the other names contained a comma.

jamwalla: Happy to break entity matching for a more correct product later!

jessex: What would it mean to break entity matching on names right now? That the next scrapes for every county after this went into production would immediately create brand new persons and bookings, and inferred release would call all previous persons and bookings released? Would we do anything to try to clean up that data situation or just let it happen?

colincadams: Indeed, it basically introduces a discontinuity across the board. So we have to decide how valuable it is for the data we have now to be continuous.

After chatting with @zacharylawrence and the fact that serializing with csv makes it so we can't add new fields, we decided maybe it's a good idea to use a more futureproof scheme. This will still introduce the discontinuity now, but should prevent it in the future. Is this crazy? 😬

jessex: I don't have an issue with creating the discontinuity now, this early in the game, as long as we have a plan to ensure this won't occur going forward. Ideally it would be a comprehensive plan that protects us on any fields we entity match across. @ohinds and Vera might disagree on the data side.

recidiviz/tests/persistence/converter/person_test.py
```diff
@@ -52,7 +52,7 @@ def testParsesPerson(self):
 
         # Assert
         expected_result = entities.Person.new_with_defaults(
-            full_name='FULL_NAME',
+            full_name='{{"full_name": "{}"}}'.format('FULL_NAME'),
```
zacharylawrence: Do you have thoughts on making an immutable attr for this with a `__repl__` that does this json formatting for you? This class could also contain the validation logic.

I'm not sure if this is best or really makes sense but do you have thoughts?

colincadams: I changed it to use an attr - I don't think it's a huge win but I do think it's nice that we can break out validation in it's own method without having to pass all of the fields.

As discussed offline, this string formatting is messy but I think I prefer to do it manually here because then we know if the formatting implementation changes which would cause entity matching to break again.

PR #1118, opened 2019-02-28
run_scraper --region all

jamwalla:
hopefully you don't want to do this very often, but might as well have it!

### DIFF
--- a/recidiviz/tools/run_scraper.py
+++ b/recidiviz/tools/run_scraper.py
@@ -43,6 +43,9 @@ from recidiviz.utils import regions
 
 
 # This function acts as a bound method to the scraper instance.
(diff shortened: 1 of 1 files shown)


recidiviz/tools/run_scraper.py
```diff
@@ -65,7 +65,10 @@ def start_scrape(queue, self, scrape_type):
 
 
 def run_scraper(args):
-    region_codes = args.region.split(',')
```
colincadams: should we use `ingest_utils.validate_regions` here? we would have to add a flag to skip the environment check

jamwalla: love it, do you have a good way to exit here? since validate_regions already produces the logging.error

PR #1135, opened 2019-03-01
Update read_enum_errors to read from correct module

terinpw:
Also small doc updates.


recidiviz/tools/read_enum_errors.py
```diff
@@ -1,10 +1,13 @@
 """Parses enum errors from logs.
 
-Run the following to generate an input file:
+Run the following to generate an input generated from production logs:
```
colincadams: -generated

PR #1137, opened 2019-03-01
added 14 jailtracker scrapers for KY

olleicua:


recidiviz/common/constants/charge.py
```diff
@@ -115,7 +115,8 @@ def _get_default_map():
     'PAROLE_VIOLATION': ChargeClass.PAROLE_VIOLATION,
     'PROBATION VIOLATION': ChargeClass.PROBATION_VIOLATION,
     'PROBATION_VIOLATION': ChargeClass.PROBATION_VIOLATION,
-    'UNKNOWN': ChargeClass.EXTERNAL_UNKNOWN
+    'UNKNOWN': ChargeClass.EXTERNAL_UNKNOWN,
+    'V': ChargeClass.EXTERNAL_UNKNOWN
```
terinpw: @ohinds for further comment, but we only want to use `EXTERNAL UNKNOWN` when the website itself says "unknown"

recidiviz/ingest/scrape/regions/us_ky_johnson/manifest.yaml
```diff
@@ -0,0 +1,6 @@
+agency_name: Big Sandy Regional Detention Center
+agency_type: jail
+base_url: $url
```
terinpw: I filed #1140, but for now can you remove the `base_url: $url` line from all the generated yamls?

recidiviz/common/constants/charge.py
ohinds: Yup, @terinpw is right. Also, 'V' means different things in different places, so it shouldn't live in the global map. I've been trying to determine from the charge name whether 'V' is a probation or parole violation, or whether it refers to some sort of minor 'violation' that should be mapped to `INFRACTION`. What was the charge name here?

olleicua: Hm.. there are several charges with this charge class for this one person:  'SPEEDING 26 MPH OR > SPEED LIMIT', 'DISREGARDING STOP SIGN', and 'IMPROPER PASSING'. It could mean vehicle related..  There's no way to link directly to the person in Jailtracker but you can search for them by name: "SHELLEY BRAY" at http://www.logancojail.com/Logan_inmate_list.html

recidiviz/ingest/scrape/regions/us_ky_johnson/manifest.yaml
olleicua: done

olleicua: So.. if I do that then I can't run the scrapers which makes it hard to test..

recidiviz/common/constants/charge.py
olleicua: I'm going to guess this is INFRACTION but also that maybe it shouldnt be in the global mat how to I put it in a map just for this scraper?

recidiviz/ingest/scrape/regions/us_ky_johnson/manifest.yaml
jamwalla: yikes, good point, leave those in for now. I filed #1168. good catch!

olleicua: I have a commit I can cherry-pick into another PR later.

recidiviz/common/constants/charge.py
olleicua: I'm guessing this one should be "infraction" so I'm adding it to the map for just that region. In general what should I do with ambiguous values? I've started a thread in scraper-writers-discuss@googlegroups.com with a bunch that I've found

recidiviz/common/constants/charge.py
```diff
@@ -102,6 +102,7 @@ def _get_default_map():
 # space. Add mappings here using a single space between words and numbers.
 # For example, `N/A` can be written as `N A` and `(10%)` can be written as `10`.
 _CHARGE_CLASS_MAP = {
+    'CO': ChargeClass.CIVIL,
```
olleicua: My best guess is this stands for 'civil offence'. Nothing else here makes sense.

olleicua:
Ok, all the enum parse errors are resolved, afaict this is ready to merge

recidiviz/common/constants/charge.py
```diff
@@ -115,7 +116,8 @@ def _get_default_map():
     'PAROLE_VIOLATION': ChargeClass.PAROLE_VIOLATION,
     'PROBATION VIOLATION': ChargeClass.PROBATION_VIOLATION,
     'PROBATION_VIOLATION': ChargeClass.PROBATION_VIOLATION,
-    'UNKNOWN': ChargeClass.EXTERNAL_UNKNOWN
+    'UNKNOWN': ChargeClass.EXTERNAL_UNKNOWN,
+    # 'V': ChargeClass.EXTERNAL_UNKNOWN
```
terinpw: nit let's get rid of this commented out line

recidiviz/common/constants/charge.py
```diff
@@ -187,12 +190,14 @@ def _get_default_map():
 # space. Add mappings here using a single space between words and numbers.
 # For example, `N/A` can be written as `N A` and `(10%)` can be written as `10`.
 _COURT_TYPE_MAP = {
+    'CC': CourtType.EXTERNAL_UNKNOWN,
```
terinpw: Same comment about `EXTERNAL_UNKNOWN`. @ohinds for what the mapping should be

recidiviz/common/constants/charge.py
olleicua: good catch

recidiviz/common/constants/charge.py
olleicua: It's hard to know whether this is Civil Court or Circuit Court but I set it to ignore for this region for now so that we can merge it if nobody knows but if it makes it easier to disambiguate here is rest of the IngestObject for that person:

INFO:root:Person:                                                                                                                                             
   surname: FERGUSON                                                                                                                                          
   given_names: SHELLEY                                                                                                                                       
   gender: F                                                                                                                                                  
   age: 38                                                                                                                                                    
   race: WHITE                                                                                                                                                
   bookings[0]: Booking:                                                                                                                                      
      booking_id: 110966                                                                                                                                      
      admission_date: 10/24/2018 12:59:00 AM                                                                                                                  
      arrest: Arrest:                                                                                                                                         
         arrest_date: 10/24/2018                                                                                                                              
         officer_name: UNREADABLE                                                                                                                             
         officer_id: 49769                                                                                                                                    
         agency: FAYETTE CO                                                                                                                                   
      charges[0]: Charge:                                                                                                                                     
         charge_id: 117631
         offense_date: 2018-10-23  
         statute: 02668                                                                                                                                       
         name: PROBATION VIOLATION (FOR FELONY OFFENSE)
         charge_class: O                                                                                                                                      
         status: REVOKED                                                                                                                    
         number_of_counts: 1
         court_type: CC                                                                                                                                       
         case_number: 16CR184

recidiviz/common/constants/charge.py
ohinds: Sounds good to me!

recidiviz/common/constants/charge.py
ohinds: Yeah, since a probation violation is on a criminal offense, I'd have a strong guess that CC is Circuit Court.

olleicua: thanks :)

olleicua:
ping

recidiviz/common/constants/charge.py
ohinds: Yes, this seems right. However, can we put this in the region scraper where it's seen? I'm nervous that CO might mean something else other places?

olleicua: So unfortunately I didn't keep track of which parse errors were from which regions so I just ran the run_scraper script for each regain and it looks like who ever had the charge class of CO is no longer there. Unrelatedly I found another one for court type but it was pretty circuit.

olleicua:
Thanks, replied above

PR #1139, opened 2019-03-01
Revert "Make infer release called per region instead of serially in o…

colincadams:
…ne task (#1053)"

This reverts commit 8405f07fee2a1a8bd04d9909d2aa527b37df0c08.


colincadams:
We believe this was interacting poorly with `eventlet` which caused infinite recursion on certain requests

PR #1143, opened 2019-03-01
TN Aggregate Ingest

ohinds:


recidiviz/ingest/aggregate/regions/tn/tn_aggregate_ingest.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+_MANUAL_FACILITY_TO_COUNTY_MAP = {
+    'Johnson City (F)': 'Washington',
+    'Kingsport City': 'Sullivan',
+}
+
+
+def parse(filename: str) -> Dict:
+
```
jessex: Extra newline.

recidiviz/ingest/aggregate/regions/tn/tn_aggregate_ingest.py
```diff
@@ -0,0 +1,128 @@
...
+from recidiviz.persistence.database.schema import TnFacilityAggregate
+
+
+_MANUAL_FACILITY_TO_COUNTY_MAP = {
+    'Johnson City (F)': 'Washington',
+    'Kingsport City': 'Sullivan',
+}
+
+
+def parse(filename: str) -> Dict:
```
zacharylawrence: Could you pass the generic type information of the keys/values into the `Dict`?

Like: `Dict[DeclarativeMeta, pd.DataFrame]`

recidiviz/ingest/aggregate/regions/tn/tn_aggregate_ingest.py
```diff
@@ -0,0 +1,128 @@
...
+        r'Pre- trial Felony': 'pretrial_felony_population',
+        r'Pre- trial Misd\.': 'pretrial_misdemeanor_population',
+        r'Total Jail Pop\.': 'total_jail_population',
+        r'Total Beds\*\*': 'total_beds',
+    }
+
+    df = aggregate_ingest_utils.rename_columns_and_select(
+        df, rename, use_regex=True)
+
+    df = df.dropna(how='all')
```
zacharylawrence: Could you add a quick comment about how some rows are parsed into 2 lines but none of the columns we care about are split over the two lines (& that's why we need to do this)?

recidiviz/persistence/database/schema.py
```diff
@@ -928,3 +928,24 @@ class PaFacilityPopAggregate(Base, _AggregateTableMixin):
     work_release_adp = Column(Integer)
     admissions = Column(Integer)
     discharge = Column(Integer)
+
+
+class TnFacilityAggregate(Base, _AggregateTableMixin):
+    """TN state-provided aggregate population statistics."""
+    __tablename__ = 'tn_facility_pop_aggregate'
```
zacharylawrence: Could you update this tablename to match the class name? `tn_facility_aggregate`

recidiviz/ingest/aggregate/regions/tn/tn_aggregate_ingest.py
zacharylawrence: This is a super nit so feel free to ignore it: In the other aggregate parses we separate the "parse what's on the pdf" logic from the "add stuff to the parsed data" logic. I think it's minor, but you can see in [ga_aggregate_ingest.py](https://github.com/Recidiviz/recidiviz-data/blob/2cbd41f67e5941121f0d29eddb181389e26cc578/recidiviz/ingest/aggregate/regions/ga/ga_aggregate_ingest.py#L36) how we have a main `parse` method which adds fips/report_date/report_granularity and knows about the schema vs the `_parse_table` which just operates on the data in the pdf and nothing else.

If you like it more as is then I'm fine with it. Thoughts?

recidiviz/ingest/aggregate/regions/tn/tn_aggregate_ingest.py
```diff
@@ -0,0 +1,128 @@
...
+
+    df['federal_and_other_population'] = df[
```
zacharylawrence: Nit: Could you add a quick comment here about how tabula parses 2 columns into 1 so we have to manually split?

recidiviz/ingest/aggregate/regions/tn/tn_aggregate_ingest.py
```diff
@@ -0,0 +1,128 @@
...
+    'Johnson City (F)': 'Washington',
+    'Kingsport City': 'Sullivan',
+}
+
+
+def parse(filename: str) -> Dict:
+
+    table = tabula.read_pdf(filename, pages=[2, 3, 4], multiple_tables=True)
+
+    formatted_dfs = []
```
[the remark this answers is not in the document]
ohinds: Nice, done.

recidiviz/ingest/aggregate/regions/tn/tn_aggregate_ingest.py
ohinds: Done.

recidiviz/ingest/aggregate/regions/tn/tn_aggregate_ingest.py
ohinds: Done.

recidiviz/persistence/database/schema.py
ohinds: Whoops!

recidiviz/ingest/aggregate/regions/tn/tn_aggregate_ingest.py
ohinds: Good idea. Done.

recidiviz/ingest/aggregate/regions/tn/tn_aggregate_ingest.py
jessex: I _think_ you're misinterpreting that. I think it's saying two lines between the end of a top-level definition and the start of a new one, not being, say, the signature of a top-level method and the first line of its implementation. So two lines between the end of `_MANUAL_FACILITY_TO_COUNTY_MAP` and the start of `parse`, not an extra line between the start of `parse` and `table = parse_table(filename)`. Does that make sense?

ohinds: Haha, I just missed which line you were calling extra. Removed.

PR #1160, opened 2019-03-04
Wa king enums

colincadams:


recidiviz/ingest/scrape/regions/us_wa_king/us_wa_king_scraper.py
```diff
@@ -158,10 +163,17 @@ def get_enum_overrides(self):
         ob.ignore('TRANSFER OF CUSTODY', ChargeStatus)
 
         # Court Type
+        # https://www.courts.wa.gov/newsinfo/resources/?fa=newsinfo_jury.brochure_guide&altMenu=Citi
+        # https://en.wikipedia.org/wiki/Courts_of_Washington_(state)
```
colincadams: After looking more into the the washington court system, it isn't totally clear to me how to map them to our court types.
* Should county district courts and federal district courts map to the same type?
* What should municipal courts that have jurisdiction over minor crimes be mapped to?
* What should the state supreme court be mapped to?

ohinds: - Should county district courts and federal district courts map to the same type?
Yup!

- What should municipal courts that have jurisdiction over minor crimes be mapped to?
Circuit, I think.

- What should the state supreme court be mapped to?
Oh! I guess we should have a Supreme enum value.

PR #1111, opened 2019-02-28
Detect scraper finished

colincadams:
part of #974

PR #1161, opened 2019-03-04
Run stop, infer release when scraper is finished

### LINKED ISSUE #974: We should detect when a scrape session for a region is complete

colincadams:
fixes #974 

I want to wait for #1111 to be merged and released so that we can see how it does before we wholly switch over to this


recidiviz/ingest/scrape/scrape_stage.py
```diff
@@ -0,0 +1,24 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+# ============================================================================
+"""Defines the order of processes to run once a scrape has finished"""
+
+next_process = {
+    'scraper_status.check_for_finished_scrapers':
+        'scraper_control.scraper_stop',
+    'scraper_control.scraper_stop': 'infer_release.infer_release',
```
colincadams: This depends on infer_release adding support to be called per region

recidiviz/ingest/scrape/queues.py
```diff
@@ -96,6 +96,17 @@ def create_task(*, region_code, queue_name, url, body):
     queue_path = format_queue_path(queue_name)
     client().create_task(queue_path, task)
 
+def create_control_task(*, region_code, url):
```
jessex: Can we add a docstring here explaining what the task will eventually be consumed by, and where its format comes from?

recidiviz/ingest/scrape/scraper_control.py
```diff
@@ -166,13 +167,18 @@ def scraper_stop():
...
     def _stop_scraper(region):
         for scrape_type in scrape_types:
             sessions.end_session(ScrapeKey(region, scrape_type))
 
         region_scraper = regions.get_region(region).get_scraper()
         region_scraper.stop_scrape(scrape_types)
 
+        queues.create_control_task(
+            region_code=region,
+            url=url_for(scrape_stage.next_process[request.endpoint]))
```
arian487: Can we add a comment here saying exactly what is called next?  Just so people don't have to click through (infer_release right now, batch_endpoint later).

recidiviz/ingest/scrape/scrape_stage.py
```diff
@@ -0,0 +1,24 @@
+# Recidiviz - a platform for tracking granular recidivism metrics in real time
```
jessex: Minor point but I had to think about the name for a second to realize it was talking about stage as in a step in a sequence, and not stage as in the environment. What do you think about changing the nomenclature here to `phase`?

recidiviz/ingest/scrape/scrape_stage.py
```diff
@@ -0,0 +1,24 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+# ============================================================================
+"""Defines the order of processes to run once a scrape has finished"""
+
+next_process = {
```
jessex: I'd suggest adding a little function for encapsulating the logic to retrieve the next phase `get_next_phase(current_endpoint: str)`. I think it's a better experience for callers, and makes it easier to evolve this in the future.

recidiviz/ingest/scrape/queues.py
colincadams: I think I understand what you meant - is this helpful / anything else to add?

recidiviz/ingest/scrape/scrape_stage.py
colincadams: yeah I like that better!

recidiviz/ingest/scrape/scrape_stage.py
colincadams: done.

nb: `request.endpoint` has type `Optional[str]`, so I added None-check at the top that throws an Error. it feels a little funny given that we could just have the param be `str`, but it makes it so we don't have to make the check at each call site.

recidiviz/ingest/scrape/scraper_control.py
colincadams: I think I prefer the task not knowing what the next task is, and it seems like those comments could easily become stale. I don't feel that strongly though so happy to add if you think it would be helpful.

recidiviz/ingest/scrape/scraper_control.py
```diff
@@ -166,13 +167,19 @@ def scraper_stop():
...
-    def _stop_scraper(region):
+    @copy_current_request_context
+    def _stop_scraper(region: str):
         for scrape_type in scrape_types:
             sessions.end_session(ScrapeKey(region, scrape_type))
 
         region_scraper = regions.get_region(region).get_scraper()
         region_scraper.stop_scrape(scrape_types)
 
+        next_phase = scrape_phase.next_phase(request.endpoint)
```
colincadams: we could move this `if phase then add_task` pattern into a convenience method in `scrape_phase` as well. I was hesitant because it would require making `scrape_phase` aware of and use `flask.url_for`, but interested to see what you all think as it is repeated code.

colincadams:
> A high level comment, since we call stop on a region, is it possible that that the scraper_status returns a completed region after we've already called stop on it from the cron? Furthermore, whats to stop the scraper_status from calling stop over and over since the first stop will show up as a completed scraper

good catch! I added in the sessions shortcut in `_check_finished`. Later on if it seems doing a datastore query for every region every ten minutes is too heavyweight, we could add a new datastore index and just do one query for all open sessions.

recidiviz/tools/build_queue_config.py
```diff
@@ -39,24 +40,26 @@ def ignore_aliases(self, data):
         return True
 
 def build_queues(environment: str):
-    queues: List[Dict[str, Any]] = []
+    qs: List[Dict[str, Any]] = []
```
colincadams: if you hate this name happy to import SCRAPER_CONTROL_QUEUE directly

recidiviz/ingest/scrape/scraper_control.py
arian487: Personally I don't mind it as is since it is a pretty narrow use case.

recidiviz/ingest/scrape/queues.py
```diff
@@ -96,6 +96,25 @@ def create_task(*, region_code, queue_name, url, body):
     queue_path = format_queue_path(queue_name)
     client().create_task(queue_path, task)
 
+SCRAPER_CONTROL_QUEUE = 'scraper-control'
+
+def create_control_task(*, region_code, url):
```
arian487: I would change the name.  When I first read it, I thought create_control_task was specific to endpoints in scraper_control but that is not the case.  We should rename this IMO.

colincadams: Updated this to use the 'phase' nomenclature as well. wdyt?

PR #1167, opened 2019-03-04
Added jids where possible

terinpw:
Added jurisdiction_ids to all counties where there is 1 id per FIPS.
This excludes the following: 'berrien', 'coweta', 'brown', 'autauga', 'morgan', 'columbia', 'jefferson', 'lonoke', 'dekalb', 'jackson', 'dale', 'cherokee', 'pike', 'lawrence', 'king', 'cape girardeau', 'monroe', 'gwinnett', 'franklin'

Also updated regions.py to temporarily remove parsing of
jurisdiction_ids.


jessex:
How does a scraper writer identify the jurisdiction id when they're starting a new jurisdiction?

terinpw:
@jessex - we're going to add functionality to create_scraper that attempts to automatically populate this information (county name and state -> fips -> jurisdiction_id). 

If there is more than one jurisdiction_id for the found fips, create_scraper will leave the jurisdiction_id field blank, and the reviewer should supply the jurisdiction_id (as we don't want scraper writers to have to interpret our csv).

PR #1170, opened 2019-03-04
Scrub pii

### LINKED ISSUE #1141: Remove PII after inferred release for places without person IDs
Related to #622

arian487:
Fixes #1141


recidiviz/persistence/persistence_utils.py
```diff
@@ -0,0 +1,36 @@
...
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+# =============================================================================
+
+"""Exposes an endpoint to scrape all of the county websites."""
+import datetime
+
+from recidiviz.persistence import entities
+
+
+def remove_pii_for_person(person: entities.Person):
```
terinpw: nit add `-> None` as a return type

recidiviz/persistence/converter/converter.py
```diff
@@ -72,7 +72,14 @@ def _convert_person(self, ingest_person):
 
         person_builder.bookings = converted_bookings
 
-        return person_builder.build()
+        converted_person = person_builder.build()
+
+        # If every booking for the person has a release date, than scrub pii
```
terinpw: nit - should be "then" not "than"

recidiviz/persistence/converter/converter.py
```diff
@@ -72,7 +72,14 @@ def _convert_person(self, ingest_person):
...
+        if all(converted_booking.release_date for converted_booking in
```
terinpw: Not necessarily we need to do in this PR, but we should consider consolidating how we determine if a booking is open. 

We do this for loop in a bunch of places, and if our definition ever changes we may miss updating one.

terinpw: Filed #1176.

recidiviz/persistence/converter/converter.py
```diff
@@ -84,7 +84,14 @@ def _convert_person(self, ingest_person):
 
         person_builder.bookings = converted_bookings
 
-        return person_builder.build()
+        converted_person = person_builder.build()
+
+        # If every booking for the person has a release date, then scrub pii
+        if all(converted_booking.release_date for converted_booking in
```
jessex: If we're planning to change the check of this logic from release date to status, let's put a TODO on this line.

recidiviz/persistence/persistence.py
```diff
@@ -89,6 +89,7 @@ def infer_release_on_open_bookings(region, last_ingest_time, custody_status):
             'Found %s people with bookings that will be inferred released',
             len(people))
         for person in people:
+            persistence_utils.remove_pii_for_person(person)
```
jessex: Can we add a note about this to the docstring?

Issue #739, opened 2019-01-25
Refactor endpoint names to be more consistent


arian487:
Right now our endpoint names are very inconsistent, we should standardize them, an example as per josh:

/api/v1/ingest/aggregates
/api/v1/ingest/records
/api/v1/ingest/scraper/{action}
/api/v1/ingest/infer_release

jessex:
We also want to place all of our endpoints in a dedicated `api` package of some sort, e.g. `recidiviz.ingest.api`.

PR #1173, opened 2019-03-04
Batch endpoint

### LINKED ISSUE #1055: Batch up all writes for a region and write only if enough tasks pass

arian487:
Create the endpoint to batch requests.  

Fixes #1055


recidiviz/persistence/batch_persistence.py
```diff
@@ -15,19 +15,33 @@
...
+from recidiviz.ingest.scrape import ingest_utils
 from recidiviz.ingest.scrape.task_params import Task
-from recidiviz.utils import pubsub_helper
+from recidiviz.persistence import persistence
+from recidiviz.utils import pubsub_helper, regions
+from recidiviz.utils.auth import authenticate_request
 
 PUBSUB_TYPE = 'scraper_batch'
+BATCH_READ_SIZE = 500
+FAILED_TASK_THRESHOLD = 1
```
jessex: So we're not writing anything if we have even a single task failure? Is this a temporary measure?

recidiviz/persistence/batch_persistence.py
```diff
@@ -64,19 +74,91 @@ def _publish_batch_message(
...
             pubsub_helper.get_topic_path(scrape_key,
                                          pubsub_type=PUBSUB_TYPE),
             data=json.dumps(serialized).encode())
+        response.result()
     pubsub_helper.retry_with_create(scrape_key, publish, PUBSUB_TYPE)
 
 
-def write(ingest_info: IngestInfo, scraper_start_time: datetime.datetime,
-          task: Task, scrape_key: ScrapeKey):
+def _get_batch_messages(scrape_key):
```
jessex: Are we planning to add MyPy types to the various new functions in here? Or docstrings?

recidiviz/persistence/batch_persistence.py
```diff
@@ -89,3 +171,46 @@ def write_error(error: str, task: Task, scrape_key: ScrapeKey):
...
+    metadata = IngestMetadata(
+        region=region, last_seen_time=scraper_start_time,
+        enum_overrides=overrides)
+
+    persistence.write(proto, metadata)
+
+
+@batch_blueprint.route('/read_and_persist', methods=['POST'])
+@authenticate_request
+def read_and_persist():
```
jessex: Can we move this back out to a higher layer instead of having an API endpoint in the persistence layer?

recidiviz/persistence/batch_persistence.py
```diff
@@ -89,3 +171,46 @@ def write_error(error: str, task: Task, scrape_key: ScrapeKey):
...
+    """Reads all of the messages on the pubsub queue for a region and persists
+    them to the database.
+    """
+    overrides = regions.get_region(region).get_enum_overrides()
+    scrape_key = ScrapeKey(region, scrape_type)
+
+    messages = _get_batch_messages(scrape_key)
+    proto, failed_tasks = _get_proto_from_messages(messages)
+    # Only acknowledge if the above code passed.
+    _ack_messages(messages, scrape_key)
```
jessex: What if we fail to actually persist in the database, even for a transient issue?

recidiviz/persistence/batch_persistence.py
arian487: These are failures in actually running scraper tasks.  If you notice, we batch things up before we call persistence.write and what we batch is the ingest_info object.  Failures in persistence have their own thresholds (which was submitted in the last PR).  

The reason we are choosing to fail if a single task fails on the scraper task side is because we don't know what the task is, and these failures are generally a lot more rare.  The failure could have been an entire page failing to load in get_more_tasks resulting in us missing a lot of data, and it is hard to know the nature of the failure.

recidiviz/persistence/batch_persistence.py
arian487: Not for the ones that talk to pubsub and get pubsub messages.  The problem is they return types that are google pubsub type that are not easily retrieved/exposed by the client.  Mypy doesn't let us do partial typing (if you type one, you have to type all).  

I have added docstrings though to be clear about what is happening.

recidiviz/persistence/batch_persistence.py
arian487: We talked about leaving failed messages unacked and have an N-day TTL (which gives us time to fix and push the issue) so it could write it later but that has some other design work that needs to get done before that cleanly works (if we leave an unacked message for later processing, what happens if the same person gets scraped again a day later and puts a message on the queue?  Some type of in memory entity matching?).  Either way, I think the current format is ok, and transient errors in persistence are very very unlikely now given that we count and catch the errors.

recidiviz/persistence/batch_persistence.py
arian487: Discussed offline, we will do that as part of the endpoint refactor:

https://github.com/Recidiviz/recidiviz-data/issues/739

I've added that to the end of fellowship milestone.

recidiviz/persistence/batch_persistence.py
jessex: Sounds reasonable.

recidiviz/persistence/batch_persistence.py
```diff
@@ -83,7 +83,15 @@ def publish():
...
 
 def _get_batch_messages(scrape_key):
-    """Reads all of the messages from pubsub"""
+    """Reads all of the messages from pubsub for the scrape key.
+
+    Args:
+        scrape_key: (ScrapeKey): The scrape key to tell us which queue to
+            retrieve from.
+    Returns:
+        A list of messages (ReceivedMessaged) which containing the message data
```
jessex: `which contain` or `containing`

recidiviz/persistence/batch_persistence.py
jessex: Sounds good to me. Thanks for the explanation!

recidiviz/persistence/batch_persistence.py
jessex: Discussed offline. Understood!

recidiviz/persistence/batch_persistence.py

PR #1174, opened 2019-03-04
pipe jurisdiction_id through our system from ingest -> db

terinpw:


recidiviz/persistence/converter/person.py
```diff
@@ -46,6 +46,10 @@ def copy_fields_to_builder(person_builder, proto, metadata):
     # TODO(769): Store place_of_residence only if it can be anonymized
     new.place_of_residence = None
 
+    new.jurisdiction_id = fn(normalize, 'jurisdiction_id', proto)
+    if not new.jurisdiction_id:
```
zacharylawrence: Can we use the `default` parameter on `fn`?

recidiviz/utils/regions.py
```diff
@@ -128,11 +129,8 @@ def get_region_manifest(region_code: str) -> Dict[str, Any]:
             as region_manifest:
         yaml_dict = yaml.load(region_manifest)
         # TODO(#386): Require jurisdiction_id once supported in codebase
-        try:
-            yaml_dict.pop('jurisdiction_id')
-        except KeyError:
-            pass
-
+        if 'jurisdiction_id' not in yaml_dict.keys():
+            yaml_dict['jurisdiction_id'] = ''
```
zacharylawrence: As discussed offline, let's go ahead and fill in each `manifest.yaml` with a `jurisdiction_id` so we don't have to write an empty string to the database.

recidiviz/persistence/converter/person.py
terinpw: Good look.

recidiviz/ingest/scrape/base_scraper.py
```diff
@@ -248,9 +248,10 @@ def _generic_scrape(self, request: QueueRequest):
                 logging.info('Last seen time of person being set as: %s',
                              request.scraper_start_time)
                 metadata = IngestMetadata(self.region.region_code,
-                                          request.scraper_start_time,
-                                          self.get_enum_overrides())
-                if self.BATCH_WRITES:
+                                          self.region.jurisdiction_id,
+                                      request.scraper_start_time,
```
jessex: Are these three de-dents correct?

recidiviz/ingest/models/ingest_info.proto
```diff
@@ -42,6 +42,7 @@ message Person {
     optional string full_name = 10;
     optional string middle_names = 11;
     optional string name_suffix = 13;
+    optional string jurisdiction_id = 14;
```
colincadams: Can you make sure you have `mypy-protobuf` installed (it's in the --dev packages) and then re run the proto compiler including the `--mypy_out flag`:
`protoc recidiviz/ingest/models/ingest_info.proto  --python_out . --mypy_out .`

This makes sure that the type hints also get updated. I'll add a comment to the proto with this info and also look into an ifttt linter.

recidiviz/common/ingest_metadata.py
```diff
@@ -27,9 +27,12 @@
 class IngestMetadata(DefaultableAttr):
```
colincadams: what do you think about requiring keywords on init? (`kw_only=True`)

recidiviz/persistence/persistence.py
```diff
@@ -63,15 +63,16 @@
 ERROR_THRESHOLD = 0.5
 
 
-def infer_release_on_open_bookings(region, last_ingest_time, custody_status):
+def infer_release_on_open_bookings(
+        region_code, last_ingest_time, custody_status):
```
colincadams: add type `str` to ensure it's the `region_code`?

recidiviz/ingest/scrape/base_scraper.py
terinpw: I have no idea how these happened -- but good catch!

recidiviz/persistence/persistence.py
terinpw: Yup! Done

recidiviz/common/ingest_metadata.py
terinpw: I'm into that -- but I would say let's do it in a follow up PR, as it's pretty widely used.

recidiviz/ingest/models/ingest_info.proto
terinpw: Ooh didn't know about this. Done!

recidiviz/utils/regions.py
terinpw: Agreed, will hold off on this PR until we've done that.

Issue #1138, opened 2019-03-01
Write all documentation required before end of fellowship


zacharylawrence:
Engineering documents on architecture for each of the following:
- Overview of our GCP project (including continuous integration) 
- Data ingestion (from a scraper, aggregate reports, and direct county ingest)
- Scraper request control (rate limiting, proxies, queues)
- Matching entities across scrape sessions
- Database architecture 
- Analytics and visualizations workflow
- Monitoring documentation 

Documentation for developers
- Setting up a development environment
- Writing a scraper (vendor and region)

How to maintain the system
- Routine scraper errors (ex. enum mapping errors)
- Deploying into production
- Exporting data into BigQuery 
- Performing schema updates

Troubleshooting documentation
- Recovering from catastrophic errors
- Restoring from database backup
- Querying the database directly
- Running individual region scrapers

Extending the system:
- Scraping a new field
- Adding a new enum value
- Creating a new BigQuery table view
- New features we didn’t have time for (snapshot scraping, etc)

For non-developers: how to use the data
- How to query BigQuery
- Tableau

PR #1177, opened 2019-03-04
ensure proto files are regenerated 

colincadams:


recidiviz/ingest/models/ingest_info.proto
```diff
@@ -19,6 +19,15 @@ syntax = "proto2";
 
 package recidiviz.ingest.models;
 
+// To regenerate, ensure `mypy-protobuf` is installed:
```
jessex: Do we have notes on the Proto in a readme somewhere, or is that being deferred to #1138?

colincadams: not currently as far as I know. Right now the readme in the `models` directory is dedicated for documenting the meaning of fields in the ingest info. I would argue that this should just be done in line in the class, but I think that battle is long lost.

Where do you think proto related documentation would be most useful? We could put it in a top-level readme, along with deployment related information, etc. Right now given that we only have one proto it makes sense to me to put it right in that file since that is the file someone would be modifying, but if we have more protos in the future it should live in a common place.

jessex: Good question. I think documentation about all development workflow, except very niche things, should generally live together. To me, having to regenerate a core data model is not niche, but something that should be contained _or at least referenced_ within the mainline dev workflow docs. I think it's important to document not just how to regenerate but when you should.

colincadams: added a note to the top-level readme about updating ingest info as well as the database schema

colincadams:
I've also added a tool to check that if one ingest info file is modified, all of them are modified. There are some small cases where this isn't necessary (i.e. me adding a comment to the proto now) but I'm hoping it will be useful most of the time. Maybe there are other sets of files where this would be useful as well?

For the times when the check doesn't need to be met, starting a commit message with `[skip validation]` will skip the check.

recidiviz/ingest/models/ingest_info.proto
jessex: Looks great, thank you!

PR #1178, opened 2019-03-04
catch named errors in persistence

jamwalla:


jamwalla:
My thinking here is that oddities in the data can cause us to raise `ConverterError`s or `EntityMatchingError`s, which we can handle gracefully during batching. Other errors would represent bugs in our code, which we want to actually raise rather than catching in an `except Exception` statement.

jamwalla:
ping @colincadams, I wonder if you have thoughts here - I didn't want to make your date parsing raise a `ConverterError` since it's used elsewhere, but we'd need to wrap converter-specific functions around it if we wanted this PR to be complete.

recidiviz/persistence/converter/errors.py
```diff
@@ -0,0 +1,21 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+# =============================================================================
+"""Contains errors for the converter directory."""
+
+
+class ConverterError(ValueError):
+    """Raised when an field in an IngestInfo cannot be converted."""
```
zdg2102: nit: "a field"

recidiviz/persistence/converter/converter.py
```diff
@@ -182,7 +183,8 @@ def _charges_pointing_to_total_bond(
         return [inferred_charge]
 
     if any(c.bond is not None for c in charges):
-        raise ValueError("Can't use total_bond and create a bond on a charge")
+        raise ConverterError("Can't use total_bond and create a bond on a "
```
zdg2102: nit: can we do single quotes for consistency?

PR #1180, opened 2019-03-05
implemented scraper; needs tests

olleicua:
This one was tricky. I still need to write tests but I'm creating the PR now so I can get feedback on the enum parse issues as well as my adding a field that was documented but not implemented to the ingest_info schema.


recidiviz/ingest/scrape/regions/us_ky_lyon/us_ky_lyon_scraper.py
```diff
@@ -0,0 +1,131 @@
...
+            if el.get('title') == ' (DateTime)':
+                return el.text_content()
+        return ''
+
+    def get_booking_projected_release_date(self, elements):
+        # There seems to be an 'Actual Completion Date' entry for each charge.
+        # I'm not sure how to interpret this but they seem to appear in
+        # chronological order so I'm taking the last one. Like many of the
+        # fields the value precedes the label hence i - 1.
+        for i, el in reversed(list(enumerate(elements))):
```
jamwalla: I think the Sentence IngestInfo object should have `completion_date` and `projected_completion_date` fields, ~~which we could store or translate into `min` and `max_length` fields if we prefer~~. Filed issue #1185.
cc @ohinds

olleicua: so in that case should I still consider the last one to be the booking_projected_release_date and just add all of them to the sentence as well?

jamwalla: yep, that's perfect!

olleicua: ok, that's done

jamwalla:
The [ingest_info.proto](https://github.com/Recidiviz/recidiviz-data/blob/master/recidiviz/ingest/models/ingest_info.proto) has instructions for regenerating the proto, that should fix some of the travis errors!

olleicua:
Ok, so assuming that my enum parse overrides are fine here this is good but travis isn't passing because I messed with the schema. It's possible I need to make addition changes to the schema so travis will pass any advice appreciated.

jamwalla:
If you run `protoc recidiviz/ingest/models/ingest_info.proto --python_out . --mypy_out .` the ingest_info_pb2.py and ingest_info_pb2.pyi files will be regenerated, that plus some pylint cleanup should fix the travis build!

olleicua:
$ protoc recidiviz/ingest/models/ingest_info.proto --python_out . --mypy_out .
zsh: command not found: protoc

$ apt-get install protoc
Reading package lists... Done
Building dependency tree       
Reading state information... Done
E: Unable to locate package protoc

jamwalla:
Ah, sorry. I think you can `pip install mypy-protobuf` or get it in the virtual environment by installing `pipenv install --dev`? Not sure because I got the binary separately; you could also try instructions like these: http://google.github.io/proto-lens/installing-protoc.html

recidiviz/ingest/scrape/regions/us_ky_lyon/us_ky_lyon_scraper.py
```diff
@@ -0,0 +1,142 @@
...
+                current = [None] * 4
+        return charges
+
+    def sanitize_string(self, string):
+        return re.sub(r"\s+", ' ', string.strip())
+
+    def get_enum_overrides(self) -> EnumOverrides:
+        overrides_builder = EnumOverrides.Builder()
+
+        # Ignore 'N/A' for all values.
```
jamwalla: outdated comment?

olleicua: good catch

olleicua:
should be good (assuming travis passes)

PR #1182, opened 2019-03-05
Build docker on Travis instead of cloud build when deploying

colincadams:
This will allow us to skip building the image at all, since it is
already built, right?


zacharylawrence:
> This will allow us to skip building the image at all, since it is
> already built, right?

Yeah, makes sense to me! Nice change!

PR #1191, opened 2019-03-05
add region specific support for infer release

terinpw:


recidiviz/tests/ingest/scrape/infer_release_test.py
```diff
@@ -69,21 +69,23 @@ def setup_method(self, _test_method):
...
-            mock_region, mock_infer_release):
+            self, mock_get_region, mock_validate_regions,
+            mock_get_most_recent_session,
+            mock_infer_release):
         headers = {'X-Appengine-Cron': "test-cron"}
+        mock_validate_regions.return_value = ['us_ut', 'us_wy']
+        mock_get_region.side_effect = [_REGIONS[0], _REGIONS[1]]
+
         time = datetime(2014, 8, 31)
-        mock_supported_regions.return_value = [region.region_code
```
colincadams: should we keep this? that way it tests the `agency_type == 'jail'` requirement.

recidiviz/ingest/scrape/infer_release.py
```diff
@@ -20,22 +20,24 @@
...
+from recidiviz.utils.regions import Region, RemovedFromWebsite, get_region
 
 infer_release_blueprint = Blueprint('infer_release', __name__)
 
 
 @infer_release_blueprint.route('/release')
 @authenticate_request
 def infer_release():
-    regions = get_supported_regions()
+    region_codes = validate_regions(get_values('region', request.args))
```
colincadams: even though I'm removing it, we should modify the `cron.yaml` to call `release?region=all` just in case this intermediate state gets deployed

recidiviz/tests/ingest/scrape/infer_release_test.py
terinpw: Agreed, updated.

recidiviz/ingest/scrape/infer_release.py
terinpw: agreed!

PR #1192, opened 2019-03-05
 Shortcut scraper stop if no open sessions

colincadams:


recidiviz/ingest/scrape/sessions.py
```diff
@@ -136,15 +137,16 @@ def end_session(scrape_key):
                                  include_closed=False,
                                  scrape_type=scrape_key.scrape_type)
 
+    closed_sessions = []
```
jessex: Can we quickly update this docstring with `Returns: list of the scrape sessions which were closed`?

colincadams: Done

PR #1193, opened 2019-03-05
Create main() for create_scraper

zacharylawrence:
Co-Authored-By: Terinpw <terinpw@users.noreply.github.com>


jamwalla:
Looks good to me! At this point do you wanna factor out a _create_parser()
function too?

zacharylawrence:
> Looks good to me! At this point do you wanna factor out a _create_parser() function too?

I don't think this will work since other parts of `main()` require knowing the `optional_args`! I'd love to be wrong though...

zacharylawrence:
@jamwalla Feel free to chime in with more recommendations! We're going to go ahead and merge for now.

PR #1203, opened 2019-03-05
Update create_scraper to set jid

zacharylawrence:


recidiviz/common/jid.py
```diff
@@ -0,0 +1,79 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+# =============================================================================
+"""
+Common utility functions used to manipulate fips.
```
terinpw: This description is for fips.py, we should update it for this new file

recidiviz/tools/create_scraper.py
```diff
@@ -170,5 +172,23 @@ def _populate_file(template_path, target_path, substitutions):
...
 
+def _get_state(state_arg) -> us.states:
+    state = us.states.lookup(state_arg)
+    if state is None:
+        raise ValueError('Couldn\'t parse state "%s"' % state_arg)
+
+    return state
+
+
+def _get_region_parts(county_name: str, state: us.states) -> Tuple[str]:
```
terinpw: Would it be clearer to instead have a method `_generate_full_region_name(county_name, state, delimiter)`?

Then we could directly call `generate_region_name(county_name, state, "-")` above. I think it's pretty unclear what  `region_parts` are on their own.

recidiviz/common/jid.py
zacharylawrence: Nice catch, updated. PTAL @terinpw!

recidiviz/tools/create_scraper.py
zacharylawrence: Good idea, done!

PR #1212, opened 2019-03-06
IML scraper dummy charges for bonds with no matching charge

### LINKED ISSUE #696: handle sites where bonds don't match our schema

jamwalla:
fixes #696


arian487:
Looks good!

PR #1215, opened 2019-03-06
Upgrade opencensus

colincadams:
Once we move to a real release of `opencensus` we can remove the explicit dependency on `opencensus-correlation`


c24t:
Looks like this was an unintended consequence of https://github.com/census-instrumentation/opencensus-python/pull/523. @reyang any way to change this so an editable install of `opencensus` from a specific rev also installs `opencensus-correlation` from that rev?

reyang:
@c24t currently I don't know if there is a way, will investigate and report back here.

reyang:
We will track the progress in https://github.com/census-instrumentation/opencensus-python/issues/544.

PR #1216, opened 2019-03-06
Replace place_of_residence with resident_of_region and residency_status

zdg2102:
This does not include the conversion logic for new scrapes, which will be added in a separate PR


recidiviz/common/constants/person.py
```diff
@@ -58,6 +58,16 @@ def _get_default_map():
         return _ETHNICITY_MAP
 
 
+class ResidencyStatus(EntityEnum, metaclass=EntityEnumMeta):
```
jamwalla: what do you think about `IN_CUSTODY`?

recidiviz/common/constants/person.py
```diff
@@ -58,6 +58,16 @@ def _get_default_map():
...
+    HOMELESS = enum_strings.residency_status_homeless
+    PERMANENT = enum_strings.residency_status_permanent
+    TRANSIENT = enum_strings.residency_status_transient
+
+    @staticmethod
+    def _get_default_map():
+        raise RuntimeError('ResidencyStatus is not mapped directly')
```
jamwalla: We could start filling this in with just the enum names as string values

recidiviz/common/constants/person.py
jamwalla: also, `EXTERNAL_UNKNOWN`?

recidiviz/persistence/database/migrations/versions/2019_03_06_1027_0b32488901d4_split_residence_field.py
```diff
@@ -0,0 +1,148 @@
+"""split_residence_field
```
jessex: Is a follow-up PR going to move much of this logic into production to process incoming residency text?

recidiviz/persistence/database/migrations/versions/2019_03_06_1027_0b32488901d4_split_residence_field.py
```diff
@@ -0,0 +1,148 @@
...
+            possible_county_names.add(
+                normalized_residence_county[:-(suffix_length + 1)])
+    for county_name in possible_county_names:
+        if normalized_region_county == county_name:
+            return True 
+    return False 
+
+
+# Returns residency status if it can be determined, and None otherwise
+def get_residency_status(place_of_residence):
```
jessex: If we did add `IN_CUSTODY` as @jamwalla suggests, we need to search for that in here, as well.

recidiviz/common/constants/person.py
jessex: `IN_CUSTODY` is an interesting one. If the goal is to perform analysis based on where a person was a resident immediately prior to their incarceration, then `IN_CUSTODY` doesn't feel quite apt. Furthermore, if `IN_CUSTODY` is a valid residency status and we don't have pre-incarceration residency information for a significant majority of the population, then should all of those people who are currently incarcerated by given this as a residency status? 

tl;dr: what do we think `IN_CUSTODY` as a residency status means, and what do we plan to use residency status for? CC @ohinds

recidiviz/persistence/database/migrations/versions/2019_03_06_1027_0b32488901d4_split_residence_field.py
```diff
@@ -0,0 +1,148 @@
...
+
+    person_results = connection.execute(
+        'SELECT person_id, place_of_residence, region '
+        'FROM person WHERE place_of_residence IS NOT NULL;').fetchall()
+    person_history_results = connection.execute(
+        'SELECT person_history_id, place_of_residence, region '
+        'FROM person_history '
+        'WHERE place_of_residence IS NOT NULL;').fetchall()
+
+    for person_id, place_of_residence, region in person_results:
```
jessex: Not sure how perfect we are trying to be with our migration code, but it looks like these two `for` blocks could be extracted into a shared helper without much trouble.

recidiviz/persistence/database/migrations/versions/2019_03_06_1027_0b32488901d4_split_residence_field.py
```diff
@@ -0,0 +1,148 @@
...
+                'UPDATE person SET residency_status = \'{}\' WHERE '
+                'person_id = {};'.format(
+                    residency_status, person_id))
+                
+    for person_history_id, place_of_residence, region in \
+            person_history_results:
+        is_resident = get_is_resident(place_of_residence, region)
+        residency_status = get_residency_status(place_of_residence)
+        if is_resident is not None:
+            connection.execute(
```
jessex: I guess we are making such few queries that we're not worried about performance and don't need to try to batch this up?

recidiviz/common/constants/person.py
zdg2102: @ohinds and I talked about this yesterday and landed on excluding `IN_CUSTODY`, at least for now. We can't very easily infer what each region intended by marking someone's residence as a carceral facility, and it also overlaps confusingly with our `IN_CUSTODY` status on `Booking.custody_status`. It also only represents a very small handful of rows, so we're not losing much information by not including it.

recidiviz/common/constants/person.py
zdg2102: The issue is this enum doesn't really convert the same way the other ones do, in that we don't do exact string matching, we do substring matching. So `'HOMELESS 10011'` would need to match to `HOMELESS`. That will probably end up becoming a map anyway, but I don't want to include it here because it violates the contract expected by `EntityEnum`

recidiviz/persistence/database/migrations/versions/2019_03_06_1027_0b32488901d4_split_residence_field.py
zdg2102: Yep, I'm going to essentially copy most of this logic to the converter 👍

recidiviz/common/constants/person.py
zdg2102: Also I'm generally a fan of not including `EXTERNAL_UNKNOWN` until we actually need it. We haven't found anywhere yet that lists `place_of_residence` as "unknown", so I'd rather not add it until we do.

recidiviz/persistence/database/migrations/versions/2019_03_06_1027_0b32488901d4_split_residence_field.py
zdg2102: Yeah, since the migrations only run once and are autogenerated 90% of the time anyway, most of the ones we have are already pretty ugly. I can change it if you feel particularly strongly about it, but I don't think there's much gain

recidiviz/persistence/database/migrations/versions/2019_03_06_1027_0b32488901d4_split_residence_field.py
zdg2102: Since they're updates, the only way to batch them would be to pass them in to a temporary table and then merge those changes over onto the real table, which would be a lot easier to mess up than just running a bunch of `UPDATE` calls. Since it only has to run once, the performance shouldn't be an issue

recidiviz/persistence/database/migrations/versions/2019_03_06_1027_0b32488901d4_split_residence_field.py
jessex: If we're planning to do that, should we be doing the migration _after_ we've started capturing the data in the database? We risk missing a slice of data between when the migration runs and when the converter logic goes live.

Then again, we're going to have another Drop Day so maybe this is irrelevant.

zdg2102: We'd have to do the migration first regardless, since that creates the columns we'd be capturing the data in. But yeah, since we'll be dropping it anyway it shouldn't create any problems.

recidiviz/tests/persistence/entity_matching_test.py
```diff
@@ -71,7 +71,7 @@ def test_matchPerson_duplicateMatch_throws(self):
 
         schema_person = schema.Person(
             person_id=_PERSON_ID, external_id=_EXTERNAL_ID,
-            full_name=_FULL_NAME, birthdate=_DATE, place_of_residence=_PLACE_1,
+            full_name=_FULL_NAME, birthdate=_DATE,
```
ohinds: Should we not use resident of region and residency status in entity matching? Or is that work deferred?

recidiviz/persistence/converter/person.py
```diff
@@ -44,7 +44,8 @@ def copy_fields_to_builder(person_builder, proto, metadata):
     new.gender = fn(Gender.parse, 'gender', proto, metadata.enum_overrides)
     new.gender_raw_text = fn(normalize, 'gender', proto)
     # TODO(769): Store place_of_residence only if it can be anonymized
```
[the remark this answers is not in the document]
zdg2102: I didn't update the text because I was being lazy, but this TODO is now actually a TODO for copying the parsing logic from the migration to the converter and calling it here. I'll delete the TODO when I do that in the followup PR

recidiviz/tests/persistence/entity_matching_test.py
zdg2102: I'm not sure why `place_of_residence` was included in this test at all, because right now as far as I can tell we only use `external_id`, `full_name`, and `birthdate` for matching people. I would personally be hesitant to use `resident_of_region` or `residency_status` because they seem less stable (e.g. it's easy for me to imagine someone getting booked and added to the system, and then a couple days later their residency information being added, in which case we'd erroneously treat them as a different person).

@terinpw @jamwalla does that all sound right to you guys?

terinpw: Yes -- I don't think the place_of_residence has any importance with this test. So it's fine to be deleted.

terinpw: With regards to adding it to the entity matching logic, I don't have a super strong feeling either way. 

I am concerned about what @zdg2102 outlined, but without looking at sites themselves, I'm not sure how common updating the person info after the booking is. Without that knowledge, I'd lean towards not adding it into entity matching unless we're having too many entity matching collisions.

PR #1218, opened 2019-03-06
fix deploy script

arian487:


deploy_production.sh
```diff
@@ -13,5 +13,6 @@ echo "Starting deploy of queue.yaml"
 python -m recidiviz.tools.build_queue_config --environment production
 gcloud app deploy queue.yaml --project=recidiviz-123
 
+VERSION=$(echo $1 | tr '.' '-')
```
colincadams: this can be inlined (it's only not in Travis because those lines aren't executed in bash)

Issue #1086, opened 2019-02-26
Write a test for base_scraper


arian487:
Currently it is only tested in run_scraper

PR #1223, opened 2019-03-07
Add base scraper test:

arian487:
Fixes https://github.com/Recidiviz/recidiviz-data/issues/1086


recidiviz/tests/ingest/scrape/base_scraper_test.py
```diff
@@ -0,0 +1,367 @@
...
+        self.tasks.append(task)
+
+# pylint: disable=protected-access
+
+
+class TestBaseScraper(TestCase):
+    """Tests for base_scraper"""
+
+    def setup_method(self, _):
+        ii_info = IngestInfo()
```
jamwalla: dumb nit - 'ii' is ingest_info so this says ingest_info_info. also use the kwargs in the create_ methods

recidiviz/tests/ingest/scrape/base_scraper_test.py
```diff
@@ -0,0 +1,367 @@
...
+        self.tasks = []
+
+    def populate_data(self, content, task, ingest_info):
+        return super().populate_data(content, task, ingest_info)
+
+    #  pylint: disable=arguments-differ
+    def add_task(self, _, task: QueueRequest):
+        self.tasks.append(task)
+
+# pylint: disable=protected-access
```
jamwalla: nit: spacing here

recidiviz/tests/ingest/scrape/base_scraper_test.py
```diff
@@ -0,0 +1,367 @@
...
+
+        # Should send the ii_info since we chose not to persist.
+        expected_tasks = [QueueRequest(
+            scrape_type=constants.ScrapeType.BACKGROUND,
+            next_task=TEST_TASK,
+            scraper_start_time=start_time,
+            ingest_info=self.ii_info
+        )]
+
+        self.assertEqual(mock_get_more.call_count, 1)
```
jamwalla: [assert_called_once](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called_once) / [assert_not_called](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_not_called)?

recidiviz/tests/ingest/scrape/base_scraper_test.py
jamwalla: actually - any reason this couldn't be a top-level 'constant' instead of in this setup method?

arian487: We call convert_ingest_info_to_proto which modifies it.  Also fixed the name

recidiviz/tests/ingest/scrape/base_scraper_test.py
arian487: Done.

recidiviz/tests/ingest/scrape/base_scraper_test.py
arian487: Did assert called once with, but pylint complains about these.  Thinks that the line has no affect...

arian487:
The plan to test BATH_WRITES is to just test when I actually launch that (probably today)

jamwalla:
all sounds good!

PR #1225, opened 2019-03-07
Add SentenceRelationship to IngestInfo

zdg2102:
Also clarify the comments on SentenceRelationship in `schema.py`


recidiviz/ingest/models/ingest_info.py
```diff
@@ -349,8 +349,37 @@ def __init__(
...
+
+    def create_sentence_relationship(
+            self, other_sentence, relationship_type) -> 'SentenceRelationship':
+        relationship = SentenceRelationship(
+            self, other_sentence, relationship_type)
+        self.sentence_relationships.append(relationship)
+        return relationship
+
+
+# TODO(1145): add SentenceRelationship to proto (and also add normalization
```
zacharylawrence: Is there a reason you updating the `ingest_info_pb2.pyi` file but not updating the `ingest_info.proto` or `ingest_info_pb2.py` file?

recidiviz/ingest/models/ingest_info_pb2.py
```diff
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
```
jessex: Why this?

recidiviz/ingest/models/ingest_info_pb2.pyi
```diff
@@ -86,6 +95,12 @@ class Person(google___protobuf___message___Message):
     def FromString(cls, s: bytes) -> Person: ...
     def MergeFrom(self, other_msg: google___protobuf___message___Message) -> None: ...
     def CopyFrom(self, other_msg: google___protobuf___message___Message) -> None: ...
+    if sys.version_info >= (3,):
```
jessex: Where did all of this come from? A protoc version upgrade sometime since the last regeneration of the proto?

recidiviz/ingest/models/ingest_info_pb2.py
zdg2102: I have no idea... just happened when I ran `protoc`

recidiviz/ingest/models/ingest_info_pb2.pyi
zdg2102: Could be? I installed protoc 3.6.1 from tar immediately before generating this, so I assume it would have been the very latest version

recidiviz/ingest/models/ingest_info.py
zdg2102: The point of this change was to only update the part that scraper writers would touch. As long as the `add_sentence_relationship` method is there, anyone writing a new scraper for a site that provides sentence relationship information will include the necessary calls, even if at the moment they don't actually do anything. That way it's easy for someone to come in later and add the proto, etc. to get those relationships actually written to the database (as opposed to if nothing was exposed to the scraper writers at all, in which case any scrapers written before the method was added would necessarily ignore any sentence relationship information)

zacharylawrence: Got it, I think @colincadams just wrote a test that will cause this to fail since a change in `ingest_info.py` will force a change in `ingest_info.proto` and `ingest_info.pyi`. Does it make sense to go ahead and add it to our proto even if we're not using the value on the persistence side?

recidiviz/ingest/models/ingest_info_pb2.py
zacharylawrence: If you're not adding anything to `ingest_info.py` then you shouldn't need to run `protoc`, right?

recidiviz/ingest/models/ingest_info.py
jessex: +1 to @zacharylawrence. Also, why are we not yet persisting this?

zdg2102: @zacharylawrence good point, I added the proto although nothing's using it here

@jessex we aren't persisting it yet because the persistence logic will be a bit more complicated than it is for the other entity types, because we have to be careful around the fact that the real relationships are undirected, but the relationships in the database are directed, so every sentence can be related either A->B or B->A. I just wanted to be sure that at least the part of the code that scraper writers will interact with was available to be used, per the reasoning in my comment above

recidiviz/ingest/models/ingest_info_pb2.py
zdg2102: I updated the proto also now

recidiviz/ingest/models/ingest_info.py
jessex: Thanks for the insight, @zdg2102.

PR #1227, opened 2019-03-07
add batch to the phase process

arian487:
Also update the tests


recidiviz/persistence/batch_persistence.py
```diff
@@ -221,20 +222,24 @@ def persist_to_database(region, scrape_type, scraper_start_time):
     persistence.write(proto, metadata)
 
 
-@batch_blueprint.route('/read_and_persist', methods=['POST'])
+@batch_blueprint.route('/read_and_persist')
```
colincadams: Out of scope of this PR but brings up the question that we could change scraper phases to be `POST` methods. Them being `GET` is not very restful, but I assume they were setup this way originally so that we could easily hit them from our browsers which does still seem useful. Maybe @jessex has an opinion?

recidiviz/persistence/batch_persistence.py
```diff
@@ -221,20 +222,24 @@ def persist_to_database(region, scrape_type, scraper_start_time):
...
     them to the database.
     """
-    json_data = request.get_data(as_text=True)
-    data = json.loads(json_data)
-
-    region = data['region']
-    scrape_type = data['scrape_type']
-    scraper_start_time = data['scraper_start_time']
+    region = request.args.get('region')
+    session = sessions.get_most_recent_completed_session(
```
colincadams: Note: I think some potential future work would be to make the whole scraper phase pipeline work on a per session basis, where we detect if a particular scrape session is complete (instead of a region) and then pass a session identifier (likely {region, scrape_type, start_time}) on to each of the following phases. Do you think that makes sense? It would fix any edge cases for multiple scrape sessions running for the same region at once I believe. Might not be that valuable, but I think it might make it easier to reason about?

In the meantime I think this is totally fine. I probably should've implemented it the other way in the first place.

recidiviz/server.py
```diff
@@ -57,6 +57,8 @@
 if logger.handlers:
     logger.handlers[0].setFormatter(fr)
 
+
```
colincadams: nit: extra blank lines

recidiviz/tests/ingest/scrape/scraper_control_test.py
```diff
@@ -33,6 +34,7 @@ def client():
     app.register_blueprint(scraper_control.scraper_control)
     # Include so that flask can get the url of `infer_release`.
     app.register_blueprint(infer_release.infer_release_blueprint)
```
colincadams: you should be able to remove the infer release blueprint now

recidiviz/persistence/batch_persistence.py
```diff
@@ -221,20 +222,24 @@ def persist_to_database(region, scrape_type, scraper_start_time):
...
+        region, ScrapeType.BACKGROUND)
+    scrape_type = session.scrape_type
+    scraper_start_time = session.start
 
     persist_to_database(region, scrape_type, scraper_start_time)
 
+    next_phase = scrape_phase.next_phase(request.endpoint)
+    if next_phase:
+        queues.enqueue_scraper_phase(
```
colincadams: Can you modify the `batch_persistence_test` to verify that `infer_release` is enqueued, the same way that the `scraper_control_test` ensures that `read_and_persist` is enqueued?

recidiviz/persistence/batch_persistence.py
arian487: I think for now hitting them from the browser is quite useful, but I agree in the future they should be changed.

recidiviz/persistence/batch_persistence.py
arian487: I agree that I think we should be operating on scrape sessions and not regions.  The edge case is not a worry though since I don't think we'll really be running two sessions of the same region.

recidiviz/server.py
arian487: Done.

recidiviz/tests/ingest/scrape/scraper_control_test.py
arian487: Done.

recidiviz/persistence/batch_persistence.py

PR #1230, opened 2019-03-07
BigQuery workflow

rasmi:
Various BigQuery workflow tools, including:

- Cloud SQL Export to GCS
- GCS import to BigQuery
- Export configuration
- Export manager
- View definitions
- View manager

Can break this into multiple PRs if that helps!


rasmi:
Draft PR, tests still incoming...

recidiviz/calculator/bq/views.py
```diff
@@ -0,0 +1,42 @@
...
+
+@attr.s(frozen=True)
+class BigQueryView:
+    """View which consists of a name (view_id) and query (view_query)"""
+    view_id: str = attr.ib()
+    view_query: str = attr.ib()
+
+
+"""Example View that computes a count of all unique people in the database."""
+PERSON_COUNT_VIEW = BigQueryView(
```
arian487: Should these be different files?  I could see some views having more complex logic that I'd want to live in its own files.

recidiviz/calculator/bq/bq_load.py
```diff
@@ -0,0 +1,136 @@
...
+                     table_ref.dataset_id,
+                     table_ref.table_id)
+
+        destination_table = bq_utils.client().get_table(table_ref)
+        logging.info('Loaded %d rows in table %s.%s.%s',
+                     destination_table.num_rows,
+                     destination_table.project,
+                     destination_table.dataset_id,
+                     destination_table.table_id)
+    except exceptions.NotFound:
```
jessex: Is this raised by both of the calls above or only one of them?

jessex: Are there other exceptions that could be raised, that we should be catching, logging, and maybe propagating?

recidiviz/calculator/bq/bq_load.py
```diff
@@ -0,0 +1,136 @@
...
+        table_ref: bigquery.table.TableReference):
+    """Wait for a table LoadJob to finish, and log its status.
+
+    Args:
+        load_job: BigQuery LoadJob whose result to wait for.
+        table_ref: TableReference to retrieve final table status.
+    """
+    try:
+        # Wait for table load job to complete.
+        load_job.result()
```
jessex: I haven't checked the API docs yet, but how is this wait being executed? Is it just a busy wait on the main thread?

recidiviz/calculator/bq/bq_load.py
```diff
@@ -0,0 +1,136 @@
...
+        table_name: str) -> Optional[Tuple[bigquery.job.LoadJob,
+                                           bigquery.table.TableReference]]:
+    """Loads a table from CSV data in GCS to BigQuery.
+
+    Given a table name, retrieve the export URI and schema from export_config,
+    then load the table into BigQuery.
+
+    This starts the job, but does not wait until it completes.
+
+    Tables are created if they do not exist, and overwritten if they do exist.
```
jessex: Might be worth being explicit that this function is not just exporting table metadata but the actual contents of each row in the table. So when it says the table is overwritten, it means that any and all data currently in the BQ table is overwritten.

recidiviz/calculator/bq/bq_load.py
```diff
@@ -0,0 +1,136 @@
...
+def load_all_tables_concurrently(
+        dataset_ref: bigquery.dataset.DatasetReference,
+        tables: List[sqlalchemy.Table]):
+    """Start all table LoadJobs concurrently.
+
+    Wait until completion to log results."""
+
+    # Kick off all table LoadJobs at the same time.
+    load_jobs = [
+        start_table_load(dataset_ref, table.name) for table in tables
```
jessex: I don't know how long the process of actually kicking off a load job takes (effectively instant, tens of seconds?) but if there is lots of overhead in that it would be nice to order in descending order by table size to squeeze out some extra efficiency.

recidiviz/calculator/bq/bq_utils.py
```diff
@@ -0,0 +1,66 @@
...
+
+def create_dataset_if_necessary(dataset_ref: bigquery.dataset.DatasetReference):
+    """Create a BigQuery dataset if it does not exist."""
+    try:
+        client().get_dataset(dataset_ref)
+    except exceptions.NotFound:
+        logging.info(
+            'Dataset %s does not exist. Creating...', str(dataset_ref))
+        dataset = bigquery.Dataset(dataset_ref)
+        client().create_dataset(dataset)
```
jessex: Similar to my previous comment: could this raise another exception we should be catching or do we want to just let these classes of potential IO exceptions just bubble?

recidiviz/calculator/bq/bq_utils.py
```diff
@@ -0,0 +1,66 @@
...
+def table_exists(
+        dataset_ref: bigquery.dataset.DatasetReference,
+        table_id: str) -> bool:
+    """Check whether or not a BigQuery Table or View exists in a Dataset."""
+    table_ref = dataset_ref.table(table_id)
+
+    try:
+        client().get_table(table_ref)
+        return True
+    except exceptions.NotFound:
```
jessex: Is this the suggested pattern from the BQ Python client, catching `NotFound` exceptions to determine if a particular resource exists? There are no boolean checks in the client we can use instead?

recidiviz/calculator/bq/bq_utils.py
```diff
@@ -0,0 +1,66 @@
...
+        logging.info(
```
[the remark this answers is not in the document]
jessex: I was originally going to say that on the first run this _shouldn't_ be a warning because it's expected. But the flip side of that is that this would be warning-worthy for _every single run after the first_, so I agree.

recidiviz/calculator/bq/bq_load.py
```diff
@@ -0,0 +1,136 @@
+"""Helper functions to create and update BigQuery Views."""
```
jessex: Make sure to fill in the license information before exiting draft mode.

recidiviz/calculator/bq/cloudsql_export.py
```diff
@@ -0,0 +1,167 @@
...
+            project=metadata.project_id(), operation=operation_id)
+        operation = get_operation.execute()
+        operation_status = operation['status']
+
+        if operation_status in {'PENDING', 'RUNNING', 'UNKNOWN'}:
+            time.sleep(3)
+        elif operation_status == 'DONE':
+            operation_in_progress = False
+
+        logging.info('Operation %s status: %s', operation_id, operation_status)
```
jessex: This feels more like a debug-level statement to me. Seeing the frequent "not quite yet" messages for each export isn't super helpful. But if we do make this debug-level, having a single "done" info-level would be a good idea.

recidiviz/calculator/bq/cloudsql_export.py
```diff
@@ -0,0 +1,167 @@
...
+    project_id = metadata.project_id()
+    instance_id = cloudsql_instance_id()
+    export_request = client().instances().export(
+        project=project_id,
+        instance=instance_id,
+        body=export_context)
+    logging.info('Starting export: %s', str(export_request.to_json()))
+    response = export_request.execute()
+
+    # We need to block until the operation is done because
```
jessex: Blocking in this thread only, though. Someone could launch another export request and trigger concurrent exports. What would the failure mode for that look like?

recidiviz/calculator/bq/cloudsql_export.py
```diff
@@ -0,0 +1,167 @@
...
+                 'in database "%s" in project "%s"',
+                 operation_id, table_name, instance_id, project_id)
+    operation_success = wait_until_operation_finished(operation_id)
+
+    return operation_success
+
+
+def export_all_tables(tables: List[sqlalchemy.Table]):
+    for table in tables:
+        export_table(table.name)
```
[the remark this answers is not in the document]
jessex: Rasmi noted above in the `export_table` function that no this cannot be parallelized because CloudSQL only supports one export at a time.

recidiviz/calculator/bq/bq_load.py
```diff
@@ -0,0 +1,136 @@
...
+    # Kick off all table LoadJobs at the same time.
+    load_jobs = [
+        start_table_load(dataset_ref, table.name) for table in tables
+    ]
+
+    # Wait for all jobs to finish, log results.
+    for load_job_started in load_jobs:
+        if load_job_started:
+            load_job, table_ref = load_job_started
+            wait_for_table_load(load_job, table_ref)
```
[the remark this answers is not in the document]
rasmi: Yes. Also, when the cron is implemented, each table will be exported/loaded separately, so this function will not be used.

recidiviz/calculator/bq/export_manager.py
```diff
@@ -0,0 +1,83 @@
...
+    export_success = cloudsql_export.export_table(table)
+    if export_success:
+        bq_load.start_table_load_and_wait(
+            dataset_ref, table)
+    else:
+        logging.error('Skipping BigQuery load of table "%s", '
+                      'which failed to export.', table)
+
+
+def export_then_load_all_sequentially():
```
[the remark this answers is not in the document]
rasmi: Cloud SQL has a one-export-at-a-time limit, so that is why we export sequentially. There is no reason to load sequentially -- this is just a convenience method around `export_table_then_load_table`. Added comment!

recidiviz/calculator/bq/bq_utils.py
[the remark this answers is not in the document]
rasmi: Done!

recidiviz/calculator/bq/cloudsql_export.py
```diff
@@ -0,0 +1,167 @@
...
+    operation_success = False
+
+    while operation_in_progress:
+        get_operation = client().operations().get(
+            project=metadata.project_id(), operation=operation_id)
+        operation = get_operation.execute()
+        operation_status = operation['status']
+
+        if operation_status in {'PENDING', 'RUNNING', 'UNKNOWN'}:
+            time.sleep(3)
```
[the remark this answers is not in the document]
rasmi: Done!

recidiviz/calculator/bq/bq_load.py
rasmi: Yes, it just hangs until the job returns.

recidiviz/calculator/bq/bq_load.py
rasmi: This was the only one I encountered, but I can imagine some ValueErrors or others sneaking in if tables or dataset names are not formatted correctly. I can catch all exceptions and surface them if you prefer?

rasmi: (I left it at NotFound because I figured it is okay for yet-unknown exceptions to fail loudly)

recidiviz/calculator/bq/bq_load.py
rasmi: Each load job takes ~15 seconds total. The request to start it is <1 second, and the job load time is dominated by misc. BQ overhead (presumably managing resources). So loading a 10GB table takes about as long as loading a 10MB table. May be worth doing this if/when we get to ~100GB-1TB tables and loads start approaching 1 minute.

But also worth noting that this function won't be used in practice, since the cron will export and load each table individually rather than export all tables and load them all parallel.

recidiviz/calculator/bq/bq_utils.py
rasmi: I think it's fine to let them bubble! Not sure what other exceptions may be raised, but I think it's fine to wait and find out. I can catch all exceptions if you'd prefer, but if there's other unexpected behavior I think it should fail loudly here.

recidiviz/calculator/bq/bq_utils.py
rasmi: The closest alternative would be to list all datasets in the project (or a single dataset if it is known which one we're using), then list all tables in that dataset and check if the table name is present. But I think this is simpler.

recidiviz/calculator/bq/cloudsql_export.py
rasmi: Good catch, I added a try-except for that case! Ideally these methods will only be called from the parent helper functions, so this wouldn't happen in practice, but now we should fail more gracefully and get a helpful error message if so.

recidiviz/calculator/bq/cloudsql_export.py
rasmi: Done!

recidiviz/calculator/bq/views.py
rasmi: Done!

recidiviz/calculator/bq/bq_load.py
rasmi: Done!

rasmi:
Integrated all requested changes and ready to merge! @jessex  -- let me know if my responses to remaining open questions are satisfactory!

recidiviz/calculator/bq/bq_utils.py
jessex: Agreed, let's stick with this.

recidiviz/calculator/bq/bq_load.py
jessex: Does this have any sort of timeout associated, or is there a different signature that allows for setting a timeout? My worry is some runaway failure where this never terminates, leaving resources just spinning away in the ether, and potentially multiple such processes if the issue isn't resolved.

recidiviz/calculator/bq/bq_load.py
jessex: Awesome, thanks for the numbers. Very helpful!

recidiviz/calculator/bq/cloudsql_export.py
```diff
@@ -0,0 +1,192 @@
...
+_client = None
+def client() -> discovery.Resource:
+    global _client
+    if not _client:
+        credentials, _ = google.auth.default()
+        _client = discovery.build(
+            'sqladmin', 'v1beta4', credentials=credentials)
+    return _client
+
+# TODO(rasmi): If not in GAE, force user to prod or staging.
```
jessex: Is there an issue filed for this or are we tackling it before merging?

recidiviz/calculator/bq/bq_utils.py
```diff
@@ -0,0 +1,83 @@
...
+    try:
+        client().get_dataset(dataset_ref)
+    except exceptions.NotFound:
+        logging.info(
+            'Dataset %s does not exist. Creating...', str(dataset_ref))
+        dataset = bigquery.Dataset(dataset_ref)
+        client().create_dataset(dataset)
+
+
+#TODO(rasmi): Combine table_exists with create_or_update_view?
```
jessex: Is there an issue filed for this or are we tackling it before merging?

recidiviz/tests/calculator/bq/export_config_test.py
```diff
@@ -0,0 +1,162 @@
...
+            export_config.TABLES_TO_EXPORT.
+
+        2) Check that all columns are defined in their respective tables.
+        """
+
+        for table in export_config.COLUMNS_TO_EXCLUDE.keys(): # pylint: disable=consider-iterating-dictionary
+            self.assertTrue(
+                table in export_config.ALL_TABLE_COLUMNS.keys(),
+                msg='Table "{}" in `export_config.COLUMNS_TO_EXCLUDE`'
+                    ' not found in `export_config.TABLES_TO_EXPORT`.'
```
jessex: I don't see it, but do we have any sort of test that asserts that all the tables we want to export are declared in `TABLES_TO_EXPORT`? I think it would be great if a unit test failed if someone added a new table in `schema.py` without either adding it to `TABLES_TO_EXPORT` or adding it to a list of explicitly not-exported tables in that test.

recidiviz/calculator/bq/bq_load.py
rasmi: Good point. This method will be run from a Cloud Task, where it will share our default Task timeout of 5 minutes. But it is good to have a timeout explicitly set here as well. I have set it to 5 minutes, since I've never seen BQ load run longer than a minute or so.

recidiviz/calculator/bq/bq_load.py
rasmi: I can capture the `load_job.result()` and log the `statistics` field fairly easily if you'd like.
It's also possible to get BQ job statistics for the past 6 months with the BQ API (or with `gcloud alpha bq jobs list`). I think this operation is not worth worrying about too much, since BQ is made to ingest 100s of GB in ~1 minute. It may be more worth monitoring for Cloud SQL exports, as I'm not sure how performant those are.

recidiviz/calculator/bq/bq_utils.py
rasmi: It's just a stylistic decision -- I like it as-is so I'll just remove this TODO.

recidiviz/tests/calculator/bq/export_config_test.py
rasmi: I think this is a good idea, but I don't know how to best capture "all the tables in `schema.py`" since there are a few classes in there that are `sqlalchemy.Table`s that we don't necessarily want to export. Opened #1306 as a TODO.

recidiviz/calculator/bq/bq_load.py
rasmi: Opened #1307.

recidiviz/calculator/bq/cloudsql_export.py
rasmi: Filed #1308. It's not a huge deal, since this will generally be run by GAE cron (or manually triggered by hitting an endpoint), but if someone runs this code locally and isn't authenticated properly, the export will fail.

rasmi:
Thanks for your reviews, @jessex @arian487! Happy to have this merged 😄

PR #1231, opened 2019-03-07
Add conversion for place of residence

zdg2102:


recidiviz/persistence/converter/person.py
```diff
@@ -118,3 +135,60 @@ def _parse_race_and_ethnicity(proto, enum_overrides):
         ethnicity = fn(Ethnicity.parse, 'ethnicity', proto, enum_overrides)
 
     return race, ethnicity
+
+
+def _parse_is_resident(proto, region):
```
zacharylawrence: Could you add type annotations here?

recidiviz/persistence/converter/person.py
```diff
@@ -118,3 +135,60 @@ def _parse_race_and_ethnicity(proto, enum_overrides):
...
+    """Returns (True) if person is resident of |region|, (False) if they are
+    not, and (None) if it cannot be determined
+    """
+
+    place_of_residence = fn(normalize, 'place_of_residence', proto)
+    if not place_of_residence:
+        return None
+
+    # Remove 'us_xx_' prefix
+    region_county = region[6:]
```
zacharylawrence: Does this correctly handle state-level scrapers, like `us_ny`?

recidiviz/common/constants/person.py
```diff
@@ -138,3 +138,10 @@ def _get_default_map():
     'N': Ethnicity.NOT_HISPANIC,
     'UNKNOWN': Ethnicity.EXTERNAL_UNKNOWN,
 }
+
+
+RESIDENCY_STATUS_SUBSTRING_MAP = {
```
zacharylawrence: The mappings from "string we scrape off a website" to "enum" is conversion logic right? Does it make sense to put this in `constants/person.py` instead of `converter/person.py`? Thoughts?

recidiviz/persistence/converter/person.py
```diff
@@ -118,3 +135,60 @@ def _parse_race_and_ethnicity(proto, enum_overrides):
...
+        if normalized_residence_county[-(suffix_length):] == suffix:
+            possible_county_names.add(
+                normalized_residence_county[:-(suffix_length + 1)])
+    for county_name in possible_county_names:
+        if normalized_region_county == county_name:
+            return True
+    return False
+
+
+def _parse_residency_status(proto):
```
zacharylawrence: Could you add type annotations here too?

recidiviz/persistence/converter/person.py
```diff
@@ -43,9 +61,8 @@ def copy_fields_to_builder(person_builder, proto, metadata):
     new.ethnicity_raw_text = fn(normalize, 'ethnicity', proto)
     new.gender = fn(Gender.parse, 'gender', proto, metadata.enum_overrides)
     new.gender_raw_text = fn(normalize, 'gender', proto)
-    # TODO(769): Store place_of_residence only if it can be anonymized
-    new.residency_status = None
-    new.resident_of_region = None
+    new.residency_status = _parse_residency_status(proto)
```
zacharylawrence: The idea behind the `fn` function was to help you not have to write the step:
```
if not place_of_residence:
    return None
```

I think this could be rewritten like this? Do you think that makes sense compared to what you have already?
```
new.residency_status = fn(_parse_residency_status, 'place_of_residence', proto)

def _parse_residency_status(place_of_residence):
    place_of_residence = normalize(place_of_residence)
    ...
```

recidiviz/persistence/converter/person.py
zdg2102: Done

recidiviz/persistence/converter/person.py
zdg2102: Updated it to handle state regions, and added tests for it

recidiviz/persistence/converter/person.py
zdg2102: Done

recidiviz/persistence/converter/person.py
zdg2102: Done

recidiviz/common/constants/person.py
zdg2102: This is almost the same as the other master conversion maps in this file, it just converts off of substrings instead of full strings. Since this might need to get updated down the line with more matching strings, the same as the other maps here, I think it makes sense to keep them all together.

zacharylawrence: Yeah, I think you're right. That sgtm!

recidiviz/persistence/converter/person.py
```diff
@@ -118,3 +137,70 @@ def _parse_race_and_ethnicity(proto, enum_overrides):
         ethnicity = fn(Ethnicity.parse, 'ethnicity', proto, enum_overrides)
 
     return race, ethnicity
+
+
+def _parse_is_resident(place_of_residence: str, region: str):
```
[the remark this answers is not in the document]
zdg2102: Done

PR #1233, opened 2019-03-08
add logs

arian487:


recidiviz/ingest/scrape/constants.py
```diff
@@ -34,6 +34,7 @@ class ScrapeType(enum.Enum):
     SNAPSHOT = "snapshot"
 
 MAX_PEOPLE_TO_LOG = 4
+PUBSUB_TYPE = 'scraper_batch'
```
colincadams: call this `BATCH_PUBSUB_TYPE`?

recidiviz/persistence/batch_persistence.py
```diff
@@ -122,6 +122,7 @@ def _ack_messages(messages, scrape_key):
             retrieve from.
     """
     ack_ids = [message.ack_id for message in messages]
+    logging.info('Got ')
```
colincadams: do you need anything else in this log?

PR #1237, opened 2019-03-08
Remove duplicate record trees from reads

zdg2102:


recidiviz/persistence/database/database.py
```diff
@@ -55,8 +55,7 @@ def read_people(session, full_name=None, birthdate=None):
         query = query.filter(Person.full_name == full_name)
     if birthdate is not None:
         query = query.filter(Person.birthdate == birthdate)
-
-    return database_utils.convert_all(query.all())
```
colincadams: can `database_utils.convert_all` be deleted?

recidiviz/persistence/database/database.py
```diff
@@ -143,6 +143,20 @@ def _query_people_and_open_bookings(session, region):
         .filter(Booking.release_date.is_(None))
 
 
+def _convert_and_normalize_record_trees(
```
colincadams: should this go in `database_utils` or does it make sense here?

recidiviz/persistence/database/database.py
zdg2102: No, `read_bookings` still uses it because it doesn't return full record trees

recidiviz/persistence/database/database.py
zdg2102: Since right now `database_utils` is just the persistence->schema converter, I think it's okay for this to stay in here

Issue #980, opened 2019-02-12
Consider having scraper tests compare against entities instead of ingest info


colincadams:
Currently entities just look at the ingest info returned from populate data. This is nice because they don't need to understand how the converter works, and just deal with what was returned from the extractor.

However, there are some cases where it would be nice for the scraper writer to know how the converter works. E.g. it can pull the bond type out of the bond amount automatically. If the scraper writer doesn't know this, they might be tempted to write custom logic in their scraper to handle this. As we continue to add more logic like this to the converter (like #941) I think this will be more useful.

jamwalla:
Let's thoroughly document in ingest_info.py the sorts of conversions that could happen on each field.

arian487:
I agree that documentation is the way to go here.

PR #1245, opened 2019-03-10
Teton, Wyoming county scraper

dmcc:


recidiviz/ingest/scrape/regions/us_wy_teton/us_wy_teton_scraper.py
```diff
@@ -0,0 +1,120 @@
...
+            elif len(charge_pieces) == 2:
+                charge_statute, charge_name = charge_pieces
+            else:
+                # Shouldn't get here since empty charges will cause 'charges' to
+                # be an list and the split prevents us from having more than 2
+                # pieces.
+                raise ValueError("Can't parse charge: %r" % charge)
+
+            charge_class = None
+            for possible_charge_class in _CHARGE_CLASSES:
```
colincadams: Note: charge class detection is already built in to the converter so this isn't necessary (see #980).

dmcc: Thanks, good to know! I've removed the charge detection from the scraper.

recidiviz/ingest/scrape/regions/us_wy_teton/us_wy_teton_scraper.py
```diff
@@ -0,0 +1,113 @@
...
+
+from recidiviz.common.constants.booking import Classification
+from recidiviz.common.constants.enum_overrides import EnumOverrides
+from recidiviz.ingest.extractor.html_data_extractor import HtmlDataExtractor
+from recidiviz.ingest.models.ingest_info import IngestInfo
+from recidiviz.ingest.scrape.base_scraper import BaseScraper
+from recidiviz.ingest.scrape import constants
+from recidiviz.ingest.scrape.task_params import ScrapedData, Task
+
+_CHARGE_CLASSES = ('felony', 'misdemeanor')
```
jamwalla: this is unused, right?

recidiviz/ingest/scrape/regions/us_wy_teton/us_wy_teton_scraper.py
```diff
@@ -0,0 +1,113 @@
...
+            elif len(charge_pieces) == 2:
+                charge_statute, charge_name = charge_pieces
+            else:
+                # Shouldn't get here since empty charges will cause 'charges' to
+                # be an list and the split prevents us from having more than 2
+                # pieces.
+                raise ValueError("Can't parse charge: %r" % charge)
+
+            charge_degree = None
+            for possible_degree in _DEGREES:
```
jamwalla: it's pretty likely that you covered the possible degrees here, but could we instead match on something like `.+ - (.+) DEGREE`?

recidiviz/ingest/scrape/regions/us_wy_teton/us_wy_teton_scraper.py
```diff
@@ -0,0 +1,113 @@
...
+                if '%s degree' % possible_degree in charge_name.lower():
+                    charge_degree = possible_degree
+                    break
+            booking.create_charge(statute=charge_statute,
+                                  name=charge_name,
+                                  degree=charge_degree)
+
+    def get_more_tasks(self, content, task: Task) -> List[Task]:
```
jamwalla: this could be the single task returned from `get_initial_task`

recidiviz/ingest/scrape/regions/us_wy_teton/us_wy_teton_scraper.py
```diff
@@ -0,0 +1,113 @@
...
+            task_type=constants.TaskType.SCRAPE_DATA,
+        )]
+
+    def get_enum_overrides(self):
+        overrides_builder = EnumOverrides.Builder()
+
+        # Booking classifications
+        overrides_builder.add('UNCLASSIFIED', Classification.EXTERNAL_UNKNOWN)
+        # This appears to be distinct from minimum/medium/maximum.
+        overrides_builder.ignore('INMATE WORKER')
```
jamwalla: @ohinds this is interesting, is it the same as Classification.WORK_RELEASE?

recidiviz/ingest/scrape/regions/us_wy_teton/us_wy_teton_scraper.py
dmcc: Oops, good catch! Fixed in latest push.

recidiviz/ingest/scrape/regions/us_wy_teton/us_wy_teton_scraper.py
dmcc: I'm not sure I've seen enough samples to know if the em-dash is always there as a separator (currently none of the listed inmates have any charges with degrees). I converted this it to use a regex. Let me know if you'd prefer the stricter regex.

recidiviz/ingest/scrape/regions/us_wy_teton/us_wy_teton_scraper.py
dmcc: Good point, fixed!

recidiviz/ingest/scrape/regions/us_wy_teton/us_wy_teton_scraper.py
dmcc: I don't have much real background on this topic and I haven't contacted the office directly, so can't be for sure. With those caveats, it seemed like this was a program where the inmates would still be housed in the jail but given work assignments during the day. Is this the same as `Classification.WORK_RELEASE`?

recidiviz/ingest/scrape/regions/us_wy_teton/us_wy_teton_scraper.py
jamwalla: that looks great!

recidiviz/ingest/scrape/regions/us_wy_teton/us_wy_teton_scraper.py
jamwalla: Checked with Oliver, one of our Vera teammates - you're doing the right thing here; since they could just be working inside it's fine to discard this information.

recidiviz/ingest/scrape/regions/us_wy_teton/manifest.yaml
```diff
@@ -3,3 +3,4 @@ agency_type: jail
 base_url: http://www.tetonsheriff.org/?page_id=28
 timezone: America/Denver
 environment: staging
+jurisdiction_id: 56039001
```
dmcc: Not sure if this is correct -- I added 001 to the FIPS code.

jamwalla: yep, that's what `create_scraper` gets too!

jamwalla:
great!! i'll merge this in now, thanks for this

PR #1248, opened 2019-03-11
Add logging for duplicate record trees

zdg2102:


recidiviz/persistence/database/database.py
```diff
@@ -149,11 +150,21 @@ def _convert_and_normalize_record_trees(
     any duplicates created by how SQLAlchemy handles joins
     """
     converted_people: List[entities.Person] = []
-    ids: Set[int] = set()
+    count_by_id: Dict[int, int] = defaultdict(lambda: 0)
     for person in people:
-        if person.person_id not in ids:
+        if count_by_id[person.person_id] == 0:
```
terinpw: nit just use `not count_by_id[person.person_id]`

recidiviz/persistence/database/database.py
```diff
@@ -149,11 +150,21 @@ def _convert_and_normalize_record_trees(
     any duplicates created by how SQLAlchemy handles joins
```
terinpw: Not really related to this PR, but can we update this comment to say `remove any duplicate people created...`. From the method description it sounds like we're looking at duplicates for all entities, but in reality we're only checking people .

recidiviz/persistence/database/database.py
zdg2102: Discussed offline, will leave as is

Issue #1045, opened 2019-02-20
Figure out how docket is used / if necessary


colincadams:
See discussion in #974

PR #1253, opened 2019-03-11
Delete unused name_lists

zacharylawrence:


zacharylawrence:
@jessex Speaking with @colincadams he pointed out that these CSVs are also used to test the SnapshotScrape functionality and was curious if we could also pull that out of the codebase? If we want to keep the SnapshotScrape functionality around then we could figure out another way to test that code without relying on the CSVs deleted in this PR.

jessex:
Hm, good question. We have a couple of open issues related to "multi-day scraping" which is the eventual successor to the prototypical snapshot scraping. I don't know what the state of that work is or whether we have designs in place. If we have made no progress on that whatsoever and have no sense yet for how we will do it, then I'm fine with removing snapshot scraping for now.

But I have heard that we have at least a few regions, at the moment, which are not completing within an overnight session and we do have `us_ny` in the codebase, so it seems like we should at least generate an idea for what to do within the next couple of weeks. If we have a plan and it relies on some of the existing snapshot scrape logic and infrastructure, I'd rather just rework the unit tests if it's easy enough.

colincadams:
just discussed offline with @zacharylawrence and we agreed that we should delete `name_list` specific code for now, but keep the docket and snapshot scrapes around. We should at some point tackle what snapshot scrapes will look like in the future, and whether the docket should be a part of that (#1045).

zacharylawrence:
This is a larger change than I realised, so I'm going to deprioritize this for now.

PR #1254, opened 2019-03-11
Delete unused calculator code

zacharylawrence:


zacharylawrence:
@jessex Is this code something we still want to keep around?

jessex:
Yes I do want to keep this around, actually. I plan to port this into Cloud Dataflow in the near future and want to keep it up so that  the team and I more easily can reference it while we design the new calculation pipeline. Thanks much for the diligence, though!

zacharylawrence:
> Yes I do want to keep this around, actually. I plan to port this into Cloud Dataflow in the near future and want to keep it up so that the team and I more easily can reference it while we design the new calculation pipeline. Thanks much for the diligence, though!

SGTM, thanks for the clarification

PR #1255, opened 2019-03-11
Delete already run `split_manifest` tool

zacharylawrence:


zacharylawrence:
@colincadams SGTM. Lets go ahead and delete this for now and if a scraper writer ends up needing something like this we can bring it back!

PR #1259, opened 2019-03-11
Use parsedatetime instead of dateparser

colincadams:
Use parsedatetime instead of dateparser

This allows us to capture more use cases, including the jailtracker `'{}y {}m {}d'` case (with a little munging).


colincadams:
@jessex this is copied from my other repo using copybara. Once we merge this I will see what the flow looks like on my other repo.

jessex:
So your other repo is the private and this is the public, and you're testing what it will look like to copy changes from public to private? Sounds great.

recidiviz/ingest/aggregate/regions/ca/ca_aggregate_site_scraper.py
```diff
@@ -72,11 +75,12 @@ def get_urls_to_download() -> List[Tuple[str, Dict]]:
...
-        date_from = dateparser.parse(match.group(1))
-        date_to = dateparser.parse(match.group(2))
-    else:
-        date_from = 1
-        date_to = 12
+        date_from = date.parse_date(match.group(1))
+        date_to = date.parse_date(match.group(2))
+
+    if not (match and date_from and date_to):
+        date_from = datetime.date(year=2002, month=1, day=1)
```
colincadams: @arian487 it's not totally clear to me what the default dates where supposed to be. WDYT?

recidiviz/persistence/converter/converter_utils.py
```diff
@@ -112,12 +78,13 @@ def calculate_birthdate_from_age(age):
...
 
 
-def parse_days(time_string):
+def parse_days(time_string: str, from_dt: Optional[datetime.datetime] = None):
     """
     Converts the given string into an int number number of days
 
     Args:
         time_string: The string to convert into int.
+        from_dt: Datetime to use as base for any relative dates.
```
jamwalla: do you think it's worth saying explicitly that return a date `time_string`days before `from_dt`?

Pipfile.lock
```diff
@@ -68,10 +68,10 @@
         },
```
jamwalla: should there also be Pipfile changes? Or is this unrelated

recidiviz/tests/common/date_test.py
```diff
@@ -0,0 +1,53 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+# =============================================================================
+"""Tests for date.py"""
```
jamwalla: we want to use unittest, right?

recidiviz/ingest/aggregate/regions/ca/ca_aggregate_site_scraper.py
arian487: Weird, I must have forgotten to write a test for this.  I don't think the 'else' is actually relevant but our default can be september 5th 1995 to day 1 of the previous month (so feb 1st for example if done today)

Pipfile.lock
colincadams: this is from adding and removing various packages, I'll revert it

recidiviz/tests/common/date_test.py
colincadams: I'm still not clear on the advantage of including unittest, given that we are using pytest as our testing framework and we depend on it for use of fixtures

recidiviz/persistence/converter/converter_utils.py
colincadams: I don't think that we can make that claim as it really depends on the string that we are given. For example the string `in 1 year` will return a date one year in the future:
```
>>> dateparser.parse('in 1 year')
datetime.datetime(2020, 3, 12, 10, 17, 3, 30627)
```

recidiviz/ingest/aggregate/regions/ca/ca_aggregate_site_scraper.py
colincadams: fixed

PR #1260, opened 2019-03-11
Add support for mirroring the codebase

colincadams:


mirror/copy.bara.sky
```diff
@@ -0,0 +1,53 @@
...
+
+core.workflow(
+    name = "importPrFromPublic",
+    origin = git.github_pr_origin(
+        url = publicUrl,
+        use_merge = True,
+    ),
+    destination = git.github_pr_destination(
+        url = privateUrl,
+        pr_branch = 'mirror/${CONTEXT_REFERENCE}',
```
jessex: To be clear, this creates a branch name with a `/` in the name, it isn't using some odd syntax to, say, create branches off of some `mirror` branch?

mirror/copy.bara.sky
```diff
@@ -0,0 +1,53 @@
...
+    ),
+    mode = "CHANGE_REQUEST",
+    set_rev_id = False,
+    destination_files = filesToPublish,
+    authoring = authoring.pass_thru(defaultAuthor),
+    transformations = [
+        metadata.save_author(),
+        metadata.squash_notes(use_merge = False),
+        metadata.replace_message("${GITHUB_PR_TITLE}\n\n${GITHUB_PR_BODY}"),
+        # does a fake merge of the PR when going from internal -> external
```
jessex: A "fake merge" of the PR against the public repository, right?

mirror/copy.bara.sky
```diff
@@ -0,0 +1,53 @@
...
+        integrates = [],
+    ),
+    mode = "CHANGE_REQUEST",
+    set_rev_id = False,
+    destination_files = filesToPublish,
+    authoring = authoring.pass_thru(defaultAuthor),
+    transformations = [
+        metadata.save_author(),
+        metadata.squash_notes(use_merge = False),
+        metadata.replace_message("${GITHUB_PR_TITLE}\n\n${GITHUB_PR_BODY}"),
```
jessex: What's this line doing exactly?

mirror/copy.bara.sky
```diff
@@ -0,0 +1,53 @@
...
+        pr_branch = 'mirror/${CONTEXT_REFERENCE}',
+        integrates = [],
+    ),
+    mode = "CHANGE_REQUEST",
+    set_rev_id = False,
+    destination_files = filesToPublish,
+    authoring = authoring.pass_thru(defaultAuthor),
+    transformations = [
+        metadata.save_author(),
+        metadata.squash_notes(use_merge = False),
```
jessex: What's this doing? `squash_notes` is squashing down the notes of a commit message, or is squashing the notes of the PR description, or is squashing commits themselves somehow?

mirror/copy.bara.sky
```diff
@@ -0,0 +1,53 @@
...
+        use_merge = True,
+    ),
+    destination = git.github_pr_destination(
+        url = privateUrl,
+        pr_branch = 'mirror/${CONTEXT_REFERENCE}',
+        integrates = [],
+    ),
+    mode = "CHANGE_REQUEST",
+    set_rev_id = False,
+    destination_files = filesToPublish,
```
jessex: So if someone, say, created a file in one of the excluded/hidden directories in a _public_ PR, then it would _not_ be copied over to the private PR?

jessex:
Thanks so much for this, @colincadams! Some workflow questions, expanding on our conversation last Friday:
* If an open source contributor issues a PR against the public repo, at what points is that repo copied to private? On merge? Does all of the review activity happen on the public repo PR?
* We talked about adding the PR copy from private back out to public. Is there any such similar functionality for _releases_? Ideally, if we create a new release on the private repo that includes changes to a non-excluded file, then we would auto-copy that release to the public repo. Not sure if that's viable.
* In the public repo, can we keep open the `regions` and `vendors` packages with README files that say something to the effect of, "This is where you would add your scrapers"? I like that as an explanation in case a future good actor is trying to build off of this repository and is wondering why there's an apparent hole.

mirror/copy.bara.sky
colincadams: correct, it just prepends the branch with 'mirror/' to hopefully make it more obvious to reviewers what those branches are from.

mirror/copy.bara.sky
colincadams: Yes that's right. I updated the comment but still not sure if it is clear?

mirror/copy.bara.sky
colincadams: I added some more comments and reordered them so that they work properly. It makes the commit message start with the PR title, followed by the PR body, so that these show up in the proper places in the new PR.

mirror/copy.bara.sky
colincadams: it gathers all the commit messages from the public branch. Only one commit is added to the private branch

mirror/copy.bara.sky
colincadams: correct. So this is not strictly necessary, but I thought it might prevent pushing scrapers to the public repo.

colincadams:
> * If an open source contributor issues a PR against the public repo, at what points is that repo copied to private? On merge? Does all of the review activity happen on the public repo PR?

I think all of the review activity should happen on the public PR and the reviewer assigned will run the copy over to private repo. If we don't have CI running directly on the public repo, then they will probably do this right at the beginning to get build results and we can potentially have red/green reported back tot he public pr (without logs). If we do have CI running on the public repo, then we can have this be a step once the public PR is approved (and can be enforced in copybara with `github_pr_origin.review_state`.

> * We talked about adding the PR copy from private back out to public. Is there any such similar functionality for _releases_? Ideally, if we create a new release on the private repo that includes changes to a non-excluded file, then we would auto-copy that release to the public repo. Not sure if that's viable.

AFAIK that is not supported. We could have copybara move tags over, but I don't think we could have the actual github release metadata moved automatically.
For tags I think it would be something like: 
```
origin {
    ref: "refs/tags/*",
}
destination {
    push: "refs/tags/*",
    fetch: "master",
}

> * In the public repo, can we keep open the `regions` and `vendors` packages with README files that say something to the effect of, "This is where you would add your scrapers"? I like that as an explanation in case a future good actor is trying to build off of this repository and is wondering why there's an apparent hole.

And we wouldn't want these READMEs to live in the private PR? That is doable, but the READMEs would have to be edited in PRs separate from files that we do want copied, as we would just merge that PR directly into public.

colincadams:
Okay I might've lied about the tag thing, I'll look into that more...

jessex:
>If we do have CI running on the public repo, then we can have this be a step once the public PR is approved (and can be enforced in copybara with github_pr_origin.review_state

I think we should have CI running on the public repo too, probably, but that's not necessary in the immediate term.

>And we wouldn't want these READMEs to live in the private PR? That is doable, but the READMEs would have to be edited in PRs separate from files that we do want copied, as we would just merge that PR directly into public.

Not PR-related, but repository-related. I want those very brief READMEs to live in both repos, public and private.

colincadams:
Alright I've added the READMEs

colincadams:
Merged the test pr and then mirrored the code, and this is what it looks like in my test repo:

Screen Shot 2019-03-12 at 2 08 26 PM
Pull Request
Screen Shot 2019-03-12 at 2 09 16 PM
Commit History

jessex:
Readmes and those test results look nice!

colincadams:
Alright, still trying to figure out the best tag solution. I tried a few ways and found one that I think should work but isn't. I filed an issue to see if I can get some help with it.

Are you okay moving forward with this with the caveat that tags may not work, or is that a dealbreaker?

jessex:
I don't think that's a non-starter, we can file an issue to resolve that later. Let's move forward with this and see how it works in the normal workflow.

colincadams:
Done! I assigned the issues for creating the public repo and moving to travis.com to you as I believe both require being a member of the recidiviz organization. Let me know if there's anything I can help with on those however!

PR #1274, opened 2019-03-13
Update README with setup for Github SSH on VM

zdg2102:


recidiviz/persistence/database/README.md
```diff
@@ -314,3 +314,33 @@ At the end of `/etc/bash.bashrc`, add the below line. This makes aliases availab
...
 ```
 
+Set up a shared Git user for pushes from the VM by creating `/etc/gitconfig` if it doesn't exist, and then running:
+
+```bash
+sudo git config --system user.name <shared user name>
+sudo git config --system user.email <shared user email>
+```
+
+In `/etc/ssh/ssh_config`, create a new host configuration for Github with the below content:
```
colincadams: flyby comment. Don't necessarily need to change it if it works, but I wonder if it would be simpler to just use a ~/.git-credentials file with the user and a personal access token for that user?

colincadams: I mention it because it's how I have copybara set up and it seems to work well

zdg2102: In that setup, would there be a separate SSH key for each user?

zdg2102: Discussed offline, will leave as-is just because we've already set it up and there's not much difference

PR #1275, opened 2019-03-13
some batch fixes

arian487:


recidiviz/persistence/batch_persistence.py
```diff
@@ -144,7 +144,6 @@ def _get_proto_from_messages(messages):
     successful_tasks = set()
     failed_tasks = set()
     for message in messages:
-        logging.info('Message has an ack id of %s', message.ack_id)
```
colincadams: 🙏

recidiviz/tests/persistence/batch_persistence_test.py
```diff
@@ -223,9 +223,8 @@ def test_persist_to_db_failed_no_write(self, mock_write):
         batch_persistence.write_error(TEST_ERROR, t2, scrape_key)
 
         start_time = datetime.datetime.now()
-        with self.assertRaises(batch_persistence.BatchPersistError):
-            batch_persistence.persist_to_database(
-                scrape_key.region_code, scrape_key.scrape_type, start_time)
+        batch_persistence.persist_to_database(
```
colincadams: assert False, and assert True in the other test

arian487: Done.

recidiviz/ingest/scrape/base_scraper.py
```diff
@@ -191,7 +192,7 @@ def _generic_scrape(self, request: QueueRequest):
                     if content == -1:
                         return -1
                 except Exception as e:
-                    raise ScraperFetchError() from e
+                    raise ScraperFetchError(str(e)) from e
```
colincadams: what's the advantage of passing this in? `raise ... from ...` will cause both errors to be printed including their tracebacks.

recidiviz/ingest/scrape/base_scraper.py
```diff
@@ -270,7 +271,8 @@ def _generic_scrape(self, request: QueueRequest):
                 scrape_key = ScrapeKey(
                     self.region.region_code, request.scrape_type)
                 batch_persistence.write_error(
-                    error=type(e).__name__,
+                    error=str(e),
+                    trace_id=flask.request.headers.get("X_CLOUD_TRACE_CONTEXT"),
```
colincadams: I found an example and it looks like this needs to be parsed a bit also:
```
def get_trace_id_from_flask():
    """Get trace_id from flask request headers.

    :rtype: str
    :returns: TraceID in HTTP request headers.
    """
    if flask is None or not flask.request:
        return None

    header = flask.request.headers.get(_FLASK_TRACE_HEADER)

    if header is None:
        return None

    trace_id = header.split("/", 1)[0]

    return trace_id
```

recidiviz/ingest/scrape/base_scraper.py
arian487: when I did str(e) in the line where I catch the error and pass it into write_error, the message was coming up as ''.  I wanted the message to actually get passed along so we could easily see all the failures.

recidiviz/ingest/scrape/base_scraper.py
arian487: Done.

recidiviz/ingest/scrape/base_scraper.py
colincadams: gotcha, e gets set as __cause__ on the new error, we can try that in a separate pr and see if that fixes it

PR #1281, opened 2019-03-13
All queues

colincadams:


deploy_production.sh
```diff
@@ -10,7 +10,7 @@ echo "Starting deploy of cron.yaml"
 gcloud app deploy cron.yaml --project=recidiviz-123
 
 echo "Starting deploy of queue.yaml"
-python -m recidiviz.tools.build_queue_config --environment production
+python -m recidiviz.tools.build_queue_config --environment all
```
arian487: Perhaps for production we should only deploy production queues?  Feel free to ignore if you feel otherwise

colincadams: agreed!

PR #1285, opened 2019-03-13
Add TN aggregate to the cron

ohinds:


cron.yaml
```diff
@@ -140,3 +140,12 @@ cron:
       min_backoff_seconds: 2.5
       max_doublings: 5
       job_age_limit: 9h
+
+  - description: Run tennessee aggregate report
+    url: /scrape_aggregate_reports/scrape_state?state=tennessee
+    schedule: every day 21:00
+    timezone: America/New_York
```
jessex: It matters almost not at all, but do we want to use the time zone for Tennessee (Central)?

ohinds: Tennessee is half in Eastern, half in Central. So is KY, and we use Eastern for that, too.

jessex: _The more you know_ 🌠

PR #1289, opened 2019-03-13
Add date imposed back to newworld scraper

zdg2102:
Removed in #705


zdg2102:
@jamwalla it was because of changes in the snapshot writer design and how it handles provided dates

PR #1296, opened 2019-03-14
Pass orphaned entities to the persistence layer after entity matching

### LINKED ISSUE #573: decide on a process for orphaned entities
If we drop entities besides those attached to a booking, nothing in the database will point to them. We could run a clean up job, or we could have bookings point to them as well. cc @zdg2102

terinpw:
closes #573


recidiviz/persistence/entity_matching.py
```diff
@@ -34,11 +34,19 @@
 
 def match_and_return_error_count(
         session: Session, region: str,
-        ingested_people: List[entities.Person]) -> int:
+        ingested_people: List[entities.Person],
+        orphaned_entities: List[Entity]) -> int:
```
zdg2102: Is there anything you could think to rename this that would make it more clear what it is from the caller's perspective? Since the caller isn't passing orphaned entities, the caller is passing an empty list they expect to be filled up with orphaned entities. `orphaned_entities_bucket`? `orphaned_entities_container`? `empty_list_for_orphaned_entities`?

recidiviz/tests/persistence/entity_matching_test.py
```diff
@@ -91,13 +91,71 @@ def test_matchPeople_errorCount(self):
...
+            bookings=[expected_booking])
+
+        orphaned_entities = []
+        error_count = entity_matching.match_and_return_error_count(
+            session, _REGION, [ingested_person], orphaned_entities)
+
+        # Assert
+        self.assertEqual(error_count, 0)
+        self.assertEqual(ingested_person, expected_person)
+        self.assertCountEqual(orphaned_entities, [expected_orphaned_bond])
```
zdg2102: Does `assertCountEqual` also check that the contained items are equal?

terinpw: Yeah, terrible name, but it does.

recidiviz/persistence/entity_matching.py
terinpw: Chatted offline w @zacharylawrence and @zdg2102. Returning it as part of a tuple with error count.

terinpw:
@jamwalla - agreed. We were debating about whether to turn it into a class. If we add anything more I'm all in on it too.

Issue #1158, opened 2019-03-01
Figure out how to actually rotate IPs on luminati


arian487:
Currently, the IPs don't rotate over some time period and we end up getting rate limited on Jailtracker.

arian487:
I've sent a message to luminati support, but it seems the only way to do this is to use the proxy manager, I'm investigating how much work this is, but in the interim its worth lowering the jailtracker QPS to ensure we don't get banned.

arian487:
I spoke with them on the phone, we do in fact need to use the proxy manager.  I've lowered the jailtracker QPS for now and will deprioritize this off of the end of fellowship.

PR #1297, opened 2019-03-14
Proxy rotation for luminati

arian487:
See: https://luminati.io/faq#integration-rotation

Fixes https://github.com/Recidiviz/recidiviz-data/issues/1158


jessex:
So we found a way to do this without the proxy manager? How did you figure this approach out?

PR #1302, opened 2019-03-14
Update jurisdiction_id to be nonnullable

terinpw:


zdg2102:
Just hold off on running until we do the drop tomorrow

PR #1303, opened 2019-03-14
Add explanation of migration and release timing to README

zdg2102:
Also add explanation of migration naming style


zdg2102:
@rasmi good call, updated

PR #1318, opened 2019-03-15
Match over ingested entities instead of db ones

terinpw:


recidiviz/persistence/entity_matching.py
```diff
@@ -89,36 +89,38 @@ def match_people_and_return_error_count(
...
         try:
-            match_person(db_person=db_person, ingested_people=ingested_people,
-                         orphaned_entities=orphaned_entities)
+            match_person(ingested_person=ingested_person, db_people=db_people,
+                         orphaned_entities=orphaned_entities,
+                         matched_db_ids=matched_db_ids)
         except Exception as e:
-            logging.error('Found error while matching db person with id %s: %s',
-                          db_person.person_id, str(e))
+            logging.error(
```
jamwalla: this is what we were doing before, but I found out you can use `logging.exception` and it gets the exception from scope!

recidiviz/persistence/entity_matching.py
```diff
@@ -241,23 +244,25 @@ def _is_match_with_relationships(*, db_entity, ingested_entity):
...
-        if ingested_obj:
+        if db_obj:
             db_id = getattr(db_obj, name + '_id')
-            logging.info('successfully matched %s with id %s',
+            logging.info('successfully matched to %s with id %s',
                          name, db_id)
-            setattr(ingested_obj, name + '_id', db_id)
-        else:
+            setattr(ing_obj, name + '_id', db_id)
+            matched_db_ids.add(db_id)  # type: ignore
```
jamwalla: can you explain this one? if this is a mypy bug I've been trying to link to the issue in mypy.

recidiviz/persistence/entity_matching.py
```diff
@@ -287,29 +292,31 @@ def match_holds(
...
+    matched_db_ids: Set[int] = set()
+    for ingested_hold in ingested_booking.holds:
+        db_hold: entities.Hold = _get_only_match(
+            ingested_hold, db_booking.holds, utils.is_hold_match)
 
-        if ingested_hold:
+        if db_hold:
             logging.info(
-                'Successfully matched hold with ID %s', db_hold.hold_id)
+                'Successfully matched to hold with ID %s', db_hold.hold_id)
```
jamwalla: In this and similar logs should we be saying "successfully matched hold {ingested_hold} to db hold with id {id}"?

recidiviz/persistence/entity_matching.py
terinpw: So it's annoying because technically entities.Sentence/Bonds have their sentence/bond_id fields as Optional. But because they're from the db, we in code know that the ids are always present. 


I thought adding the ignore would be the easiest workaround, but let me know if you have a different idea.

recidiviz/persistence/entity_matching.py
terinpw: Now that they're debug level i'm down for this. I'll do it in a follow up PR though since it's in several places.

PR #1321, opened 2019-03-15
call flask methods outside threads

colincadams:
hopefully this will fix errors like: 
https://pantheon.corp.google.com/errors/CPKkzo2V06Op3gE?time=P1D&project=recidiviz-123

although it isn't clear why the app contexts don't work correctly across threads


colincadams:
ran pylint, pytest, and mypy locally

PR #1332, opened 2019-03-18
use gevent to properly use async workers

### LINKED ISSUE #1325: Requests occassionally 502
Its unclear exactly why, the only error we get is:

[error] 33#33: *1126 recv() failed (104: Connection reset by peer) while reading response header from upstream, client: 2002:ab0:8703::, server: , request: "POST /scraper/work/us_ky_allen HTTP/1.1", upstream: "http://172.17.0.1:8080/scraper/work/us_ky_allen", host: "recidiviz-123.appspot.com"

Looking around it doesn't appear there are good reasons for this, we may have to contact Cloud folks directly.

### LINKED ISSUE #1327: Some requests take a very long time
Some requests take > 200 seconds to run, it appears a lot of that time is spent waiting on either the publisher client or datastore.  Its possible its because these are global variables and each request is an eventlet thread and not a separate process, so they wait on each other.  This could be indirectly causing https://github.com/Recidiviz/recidiviz-data/issues/1325

colincadams:
fixes #1325 fixes #1327


gunicorn.conf.py
```diff
@@ -15,11 +15,16 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 # =============================================================================
 """Configures gunicorn"""
+import multiprocessing
+from gevent_grpc_worker import GeventGrpcWorker
 
-# Recommended number of workers based on instance size:
-# https://cloud.google.com/appengine/docs/standard/python3/runtime#entrypoint_best_practices
-workers = 2
+workers = multiprocessing.cpu_count() * 2 + 1
```
rasmi: Can you link: http://docs.gunicorn.org/en/stable/design.html#how-many-workers ?

rasmi: (not strictly necessary!)

PR #1345, opened 2019-03-19
Add BQ export to tasks and cron.

rasmi:


recidiviz/calculator/bq/export_manager.py
```diff
@@ -97,5 +106,50 @@ def export_all_then_load_all():
...
+    """Worker function to handle BQ export task requests.
+
+    Form data must be a bytes-encoded JSON object with parameters listed below.
+
+    URL Parameters:
+        table_name: Table to export then import. Table must be defined
+            in export_config.TABLES_TO_EXPORT.
+    """
+    # Verify this was actually a task queued by our app
+    if 'X-AppEngine-QueueName' not in request.headers:
```
colincadams: `authenticate_request` should do this for us, unless you want to prevent this from being able to be called manually?

recidiviz/calculator/bq/export_manager.py
```diff
@@ -42,18 +48,21 @@ def export_table_then_load_table(
...
     """
     if not dataset_ref:
         dataset_ref = bq_utils.client().dataset(
             export_config.BASE_TABLES_BQ_DATASET)
 
     export_success = cloudsql_export.export_table(table)
-    if export_success:
-        bq_load.start_table_load_and_wait(
-            dataset_ref, table)
+    if export_success: # pylint: disable=no-else-return
```
colincadams: nit: instead of disabling this why not just remove the else and put the logging error and return False at the higher scope?

rasmi: I think this style is more readable (if success, do this, else do that) but open to changing if you think the other way is better. WDYT?

recidiviz/calculator/bq/export_manager.py
rasmi: I wasn't aware, I can remove it. I guess this should never be called manually (since it just calls the function above which can be called manually), but it's not important it to prevent it. What do you think?

rasmi: I'll just remove it!

rasmi: Removed!

recidiviz/calculator/bq/export_manager.py
```diff
@@ -97,5 +106,45 @@ def export_all_then_load_all():
...
+
+    A task is created for each table defined in export_config.TABLES_TO_EXPORT.
+
+    Re-creates all tasks if any task fails to be created.
+    """
+    try:
+        for table in export_config.TABLES_TO_EXPORT:
+            queues.create_bq_task(table.name, '/export_manager/export')
+        return ('', HTTPStatus.OK)
+    except: # pylint: disable=bare-except
```
colincadams: missed this the first time, flask will automatically catch an exception, log it, and return an INTERNAL_SERVER_ERROR, so this likely isn't necessary. If you still want the custom logging prefix you an catch the exception, log and re-raise? But I think leaving it with now try except at all is the cleanest.

rasmi: Done, thanks!

Issue #1344, opened 2019-03-19
Investigate why pubsub pull takes minutes


arian487:
With all of the 502s gone now that we are fully running gevent, these calls should not take minutes.  Publish calls are all fast and async now, but the pull from batch_and_persist is still slow (this is less game breaking but still worth improving).

PR #1346, opened 2019-03-19
Add logging and move ack

arian487:


recidiviz/persistence/batch_persistence.py
```diff
@@ -100,18 +100,31 @@ def _get_batch_messages(scrape_key):
     """
 
     def inner():
-        return pubsub_helper.get_subscriber().pull(
-            pubsub_helper.get_subscription_path(scrape_key,
-                                                pubsub_type=PUBSUB_TYPE),
+        subscriber = pubsub_helper.get_subscriber()
+        logging.debug('Got the pubsub subscriber')
```
colincadams: note that after @jamwalla's change yesterday, debug logs don't show up by default. we should either change these to info or up the log level sent to stackdriver for now

recidiviz/persistence/batch_persistence.py
```diff
@@ -100,18 +100,31 @@ def _get_batch_messages(scrape_key):
...
+        logging.debug('Pulling messages off pubsub')
         response = pubsub_helper.retry_with_create(
             scrape_key, inner, pubsub_type=PUBSUB_TYPE)
+        logging.debug('Finished pulling messages, read %s messages',
+                      len(response.received_messages))
         if response.received_messages:
+            logging.debug('Acking the read messages')
+            # TODO(1339): Move this back to bottom of method pending
+            #  scraper_start fix.
+            _ack_messages(response.received_messages)
```
colincadams: looks like `_ack_messages` requires a scrape key

recidiviz/persistence/batch_persistence.py
```diff
@@ -100,18 +100,31 @@ def _get_batch_messages(scrape_key):
...
-        return pubsub_helper.get_subscriber().pull(
-            pubsub_helper.get_subscription_path(scrape_key,
-                                                pubsub_type=PUBSUB_TYPE),
+        subscriber = pubsub_helper.get_subscriber()
+        logging.debug('Got the pubsub subscriber')
+        logging.debug('Getting subscription path')
+        sub_path = pubsub_helper.get_subscription_path(
+            scrape_key, pubsub_type=PUBSUB_TYPE)
+        logging.debug('Got the subscription path')
+        return subscriber.pull(
```
colincadams: regarding latency, see this note from [the docs](https://cloud.google.com/pubsub/docs/pull):
> Note that to achieve low message delivery latency with synchronous pull, it is important to have many simultaneously outstanding pull requests. As the throughput of the topic increases, more pull requests are necessary. In general, asynchronous pull is preferable for latency-sensitive applications.

Maybe we should try asynchronous pull or sending many synchronous pull requests simultaneously to see if it lowers our latency? Asynchronous seems like it might be our best option.

recidiviz/persistence/batch_persistence.py
arian487: Fixed.

recidiviz/persistence/batch_persistence.py
arian487: Agreed, lets see what this logging brings up when deployed.

recidiviz/persistence/batch_persistence.py

recidiviz/persistence/batch_persistence.py
colincadams: ack cc #1344

PR #1350, opened 2019-03-19
Clear pubsub queue on scraper start 

### LINKED ISSUE #1339: Update scraper start API to correctly handle concurrent scrapes of the same region
For a region, we should only start a new scrape session if there is no active scrape session. 

Once we start a new scrape session, we should clear all messages in the pub sub queue for that region.

terinpw:
closes #1339


recidiviz/ingest/scrape/scraper_control.py
```diff
@@ -33,7 +33,8 @@
 from recidiviz.ingest.models.scrape_key import ScrapeKey
 from recidiviz.ingest.scrape import (docket, ingest_utils, scrape_phase,
                                      sessions, tracker)
-from recidiviz.utils import monitoring, regions
+from recidiviz.ingest.scrape.constants import PUBSUB_TYPE
```
colincadams: I think either `PUBSUB_TYPE` needs to be renamed to `BATCH_PUBSUB_TYPE` or moved into the `batch_persistence` file. Otherwise it is hard to tell whether this is the pubsub type for batch or the docket.

terinpw: Fixed!

Issue #1271, opened 2019-03-12
create public repo


colincadams:

jessex:
https://github.com/Recidiviz/pulse-data lives!

@colincadams has admin access to perform the cutover mirror. We'll keep this open until the mirror is up and running.

PR #1352, opened 2019-03-19
automate copybara

### LINKED ISSUE #1270: automate copybara
We want travis to run copybara for us after each commit

colincadams:
fixes #1270


mirror/export.sh
```diff
@@ -0,0 +1,11 @@
+git config --global user.email "travis@travis-ci.com"
+git config --global user.name "Travis CI"
+echo "https://user%40domain.com:${GH_TOKEN}@github.com" > ~/.git-credentials
```
colincadams: @jessex Here we need an account to authenticate with github. We could use our existing robot account or a new one. Once we have the account, we need to create a personal access token for it and store it as an encrypted variable in Travis so that we can access it here.

jessex: Let's use the existing helper bot! You should have all the access you need to set up the encrypted access token, but let me know if you hit any snags.

jessex:
Is this still a WIP, @colincadams? It looks great to me but don't want to submit a premature review.

colincadams:
other than filling in the git credentials this is good to go!

colincadams:
Before we can merge this, we need to fix #1271 and replace the `publicUrl` in `copy.bara.sky`. Otherwise I think the builds will break.

jessex:
The repo is created now. You should be unblocked to make the cutover!

colincadams:
rerequesting review as this has changed some!

PR #1358, opened 2019-03-19
use pre-built docker image for deployments

colincadams:


.travis.yml
```diff
@@ -43,6 +39,14 @@ after_success:
   - coveralls
 
 before_deploy:
+  # App engine doesn't allow '.' in the version name
+  - VERSION=$(echo $TRAVIS_TAG | tr '.' '-')
+
+  # Push the docker image to GCR
+  - IMAGE_URL=us.gcr.io/recidiviz-staging/appengine/default.$VERSION:latest
```
jessex: This is private by default, right? That's not a Docker image accessible to the world on push?

colincadams: oh good catch! I think this should fix it? I hate editing the deploy part of this file because I have no way to test it

jessex: Yeah it's my least favorite thing, too: testing changes in CI/CD/CR/deployment/Docker land. Trial and error at its dullest. If we have to test this by trial and error, let's make sure that either 1) this can't succeed in deploying a generally public image, or 2) if it can succeed in deploying a generally public image then it's fully deletable.

colincadams: I confirmed locally that when not logged into docker I cannot pull or push to gcr for our projects.

PR #1360, opened 2019-03-19
add region and scrape information to BatchPersistError

jamwalla:
we were getting `recidiviz.persistence.validator.ValidationError`s without info about where the bad data came from!


recidiviz/persistence/batch_persistence.py
```diff
@@ -43,6 +44,11 @@
 
 class BatchPersistError(Exception):
     """Raised when there was an error with batch persistence."""
+    def __init__(self, region: str, scrape_type: ScrapeType,
```
colincadams: note: you don't need to pass in the exception as a parameter. In python 3.7, doing raise in an except block will set the old exception as the `__context__` of the new exception, and it will get printed out first.

From https://docs.python.org/3/library/exceptions.html:
> When raising (or re-raising) an exception in an except or finally clause __context__ is automatically set to the last exception caught; if the new exception is not handled the traceback that is eventually displayed will include the originating exception(s) and the final exception.

jamwalla: oh, sweet!! 🙏🏽

PR #1365, opened 2019-03-19
travis has a very old gcloud

colincadams:


.travis.yml
```diff
@@ -48,7 +48,7 @@ before_deploy:
 
   # Authenticate gcloud
   - gcloud auth activate-service-account recidiviz-staging@appspot.gserviceaccount.com --key-file client-secret-staging.json
-  - gcloud auth configure-docker
+  - gcloud beta auth configure-docker  # use beta as travis has an old gcloud
```
jessex: So the old version doesn't have the built-in `auth` command, but it does have a built-in `beta` command which lets you use commands not built-in to the version yet? Seems reasonable to me, though I wonder if we considered just upgrading the `gcloud` version in the environment?

colincadams: In the version they use the `configure-docker` was still in beta:
https://travis-ci.org/Recidiviz/recidiviz-data/builds/508638093#L4018 

I attempted to just upgrade the local gcloud installation, but that failed as well:
https://travis-ci.org/Recidiviz/recidiviz-data/builds/508652472#L444

We could install our own version of gcloud, which is what we used to do when we installed the emulators on travis prior to the docker image, but that would add time and seemed more heavyweight than necessary.

PR #1368, opened 2019-03-20
use our own deploy script for travis instead of built-in gae

colincadams:
The builtin gae deploy provider seems broken in that `docker_build: local` doesn't work properly and `image_url: $IMAGE_URL` is not supported yet. I started a [thread on their forum](https://travis-ci.community/t/gae-deploy-doesnt-support-image-url-and-docker-build-is-deprecated/2724), will see if they get back to us.

We can work around this by using our own deploy script, similar to `deploy_production.sh` and use that instead. Their abstraction isn't very useful to us anyway, and this gives us more flexibility in the future.

Hopefully this works 🤞


travis/deploy_staging.sh
```diff
@@ -0,0 +1,32 @@
...
+# use instead of 'auth configure-docker' as travis has an old gcloud
+gcloud docker --authorize-only
+
+# Push the docker image to GCR
+IMAGE_URL=us.gcr.io/recidiviz-staging/appengine/default.$VERSION:latest
+docker tag recidiviz-image $IMAGE_URL
+docker push $IMAGE_URL
+
+# Deploy application
+gcloud app deploy staging.yaml --project recidiviz-staging --version $VERSION --image-url $IMAGE_URL
```
zacharylawrence: Nit: Could you add the newline here?

colincadams: done

PR #1371, opened 2019-03-20
Add aggregate mappings for bq view

zacharylawrence:
Note: The query generated from this PR is tested in the BQ UI and successfully combines the aggregate tables into a view!

Co-Authored-By: ohinds <ohinds@users.noreply.github.com>


recidiviz/calculator/bq/views/view_manager.py
```diff
@@ -24,14 +24,15 @@
 from typing import List
 
 from recidiviz.calculator.bq import bq_utils
-from recidiviz.calculator.bq.views import bqview
+from recidiviz.calculator.bq.views import bqview, aggregate_views
```
rasmi: Can you do this import on a separate line?

zacharylawrence:
> Great! I didn't relook at the mappings for views other than HI.

Should be fine since the others were left as is.

recidiviz/calculator/bq/views/view_manager.py
zacharylawrence: Done!

recidiviz/calculator/bq/views/aggregate_views.py
```diff
@@ -0,0 +1,682 @@
...
+            select_statement = null().label(new_view_column_name)
+            if source_column is not None:
+                select_statement = source_column.label(new_view_column_name)
+
+            select_statements.append(select_statement)
+
+        return Session().query(*select_statements)
+
+
+def union(queries: List[Query]) -> Query:
```
jessex: I think we could replace this with a snappy lil `reduce` like:
```
return reduce((lambda left, right: left.union(right)), queries)
```

Haven't tested it but I think that works.

recidiviz/calculator/bq/views/aggregate_views.py
```diff
@@ -0,0 +1,682 @@
...
+    FlFacilityAggregate, GaCountyAggregate, \
+    HiFacilityAggregate, KyFacilityAggregate, NyFacilityAggregate, \
+    TxCountyAggregate, PaFacilityPopAggregate, PaCountyPreSentencedAggregate, \
+    TnFacilityAggregate, TnFacilityFemaleAggregate
+
+
+@attr.s(frozen=True, kw_only=True, auto_attribs=True)
+class Mappings:
+    # Required Primary Key
+    fips: Column
```
jessex: I think it would be useful to also include the high-level region in each row, so it's easy for the human analyst to quickly spot where they're looking.

recidiviz/calculator/bq/views/aggregate_views.py
```diff
@@ -0,0 +1,682 @@
...
+    held_for_federal_male: Optional[Column]
+    held_for_federal_female: Optional[Column]
+    total_held_for_other_male: Optional[Column]
+    total_held_for_other_female: Optional[Column]
+
+    def to_query(self) -> Query:
+        """Create a query to SELECT each column based on the Mapping's name."""
+        select_statements = []
+        for new_view_column_name, source_column in attr.asdict(self).items():
+            select_statement = null().label(new_view_column_name)
```
jessex: What is this line doing exactly?

recidiviz/calculator/bq/views/aggregate_views.py
```diff
@@ -0,0 +1,682 @@
...
+    aggregation_window=GaCountyAggregate.aggregation_window,
+    resident_population=None,
+    jail_capacity=GaCountyAggregate.jail_capacity,
+    custodial=GaCountyAggregate.total_number_of_inmates_in_jail,
+    admissions=None,
+    jurisdictional=None,
+    male=None,
+    female=None,
+    # We can't compute `sentenced` because summing `sentenced_to_state` and
+    # `serving_county_sentence` discards people in `other` who are sentenced
```
jessex: Do we have nowhere we can use `serving_county_sentence`? I'm guessing no other state is publishing that information or else it would be in `Mappings` already but it seems a shame to toss it out (though of course it's discoverable elsewhere).

Are there any other fields we are not copying from the aggregate table into the mappings?

recidiviz/calculator/bq/views/aggregate_views.py
```diff
@@ -0,0 +1,682 @@
...
+from recidiviz.calculator.bq.views.bqview import BigQueryView
+from recidiviz.persistence.database.schema import CaFacilityAggregate, \
+    FlFacilityAggregate, GaCountyAggregate, \
+    HiFacilityAggregate, KyFacilityAggregate, NyFacilityAggregate, \
+    TxCountyAggregate, PaFacilityPopAggregate, PaCountyPreSentencedAggregate, \
+    TnFacilityAggregate, TnFacilityFemaleAggregate
+
+
+@attr.s(frozen=True, kw_only=True, auto_attribs=True)
+class Mappings:
```
jessex: Are we capturing per-facility aggregates anywhere in here? I don't see a field for it in `Mappings` and a quick scan of the appended mappings below doesn't seem to show any of the per-facility aggregates making use of the facility name.

recidiviz/calculator/bq/views/aggregate_views.py
```diff
@@ -0,0 +1,682 @@
...
+    held_for_doc_male=TnFacilityAggregate.tdoc_backup_population - TnFacilityFemaleAggregate.tdoc_backup_population,
+    held_for_doc_female=TnFacilityFemaleAggregate.tdoc_backup_population,
+    held_for_federal_male=None,
+    held_for_federal_female=None,
+    total_held_for_other_male=TnFacilityAggregate.federal_and_other_population + TnFacilityAggregate.tdoc_backup_population - (TnFacilityFemaleAggregate.federal_and_other_population + TnFacilityFemaleAggregate.tdoc_backup_population),
+    total_held_for_other_female=TnFacilityFemaleAggregate.federal_and_other_population + TnFacilityFemaleAggregate.tdoc_backup_population,
+))
+
+_AGGREGATE_QUERY = union([m.to_query() for m in _AGGREGATE_MAPPINGS])
+AGGREGATE_VIEW = BigQueryView(
```
rasmi: Please change to STATE_AGGREGATE_VIEW

recidiviz/calculator/bq/views/aggregate_views.py
```diff
@@ -0,0 +1,682 @@
...
+    view_id='aggregate',
```
rasmi: Please change to `combined_state_aggregates` or something like that.

recidiviz/calculator/bq/views/aggregate_views.py
zacharylawrence: Good point, I added a comment for this. The idea is that the unioned queries require that all the fields are selectable (even if just NULL)

recidiviz/calculator/bq/views/aggregate_views.py
zacharylawrence: Done!

recidiviz/calculator/bq/views/aggregate_views.py
zacharylawrence: Done!

zacharylawrence:
Note: The query generated from this PR is tested in the BQ UI and successfully combines the aggregate tables into a view!

recidiviz/calculator/bq/views/aggregate_views.py
zacharylawrence: No, we're currently truncating everything to the county level. Notably we're not summing to the county level (ie. 2 facilities in the same FIPS will give us 2 rows in the output with the same FIPS)!

The next task in combining these aggregate reports is to add `jurisidction_id` to aggregate reports (tracked in #1396).

The reasoning behind this is that this new BQ View will combine data from both the "state-reported aggregate reports" and "scraped website data". Since all the data scraped from websites is aggregated to the jurisdiction level, this view needs to be too.

recidiviz/calculator/bq/views/aggregate_views.py
zacharylawrence: Right. Since the new BQ View we're creating will be the the combination of both the "state-reported aggregate reports" and "scraped website data", any column in this mapping needs to be present from either multiple aggregate_reports or available from scraped websites.

For `serving_county_sentence` specifically, @ohinds decided that it's not a column that cross-county questions would be asked about.

So I guess the heuristic is, only list the column if:
1. It can be found in multiple aggregate reports or a websites
1. We want to ask questions referencing that column in cross-county analysis

jessex: That's super helpful. Thanks for the explanation!

recidiviz/calculator/bq/views/aggregate_views.py
zacharylawrence: Good point, we should also add a constant for whether the row is from the aggregate reports or scraped website data. I've made a todo (#1399) to track adding both of these since they're both literal constants (not columns). I'll add this in a follow-up PR.

recidiviz/calculator/bq/views/aggregate_views.py
zacharylawrence: Great! I also threw these notes into `mappings.py`

PR #1372, opened 2019-03-20
Fix verbose travis

colincadams:


travis/deploy_staging.sh
```diff
@@ -1,32 +1,37 @@
-#!/bin/bash set -v
-#
 # Use verbose mode so that the commands are printed to Travis. Do not expand
-# variables as some are secrets.
+# variables (set -x) as some are secrets.
```
jessex: -x or -v?

colincadams: set -x would expand them, so we want to use set -v instead. would different wording help?

jessex: My first reading of the comment was that you were saying `set -x` is a thing we are doing or should do, but if you're saying that `set -x` is a thing we should absolutely never do, I think we should be _slightly_ more explicit. Or I could just be overly pedantic. That wouldn't be new.

colincadams: done!

PR #1373, opened 2019-03-20
include all relevant entities in EntityMatchingErrors

jamwalla:


recidiviz/persistence/entity_matching.py
```diff
@@ -32,6 +32,36 @@
 from recidiviz.persistence.errors import EntityMatchingError
 
 
+class MultipleDatabaseEntityMatchingError(EntityMatchingError):
```
terinpw: Let's add these to persistence/errors?

terinpw: Also how do we feel about `MatchedMultipleDatabaseEntitiesError` / `MatchedMultipleIngestedEntitiesError`?  imo having matched at the beginning is clearer

recidiviz/persistence/entity_matching.py
```diff
@@ -107,7 +137,7 @@ def match_person(
         *, ingested_person: entities.Person,
         db_people: List[entities.Person],
         orphaned_entities: List[Entity],
-        matched_db_ids: Set[int]) -> None:
+        matched_db_ids: Dict[int, entities.Person]) -> None:
```
terinpw: can we rename `matched_db_ids` to maybe `matched_people_by_db_id`? Or something more clear

recidiviz/persistence/entity_matching.py
```diff
@@ -116,11 +146,10 @@ def match_person(
...
         # person, raise an error.
         if db_person.person_id in matched_db_ids:
-            raise EntityMatchingError(
-                'matched db person {} to two ingested entities'.format(
-                    db_person.person_id))
+            matches = [ingested_person, matched_db_ids[db_person.person_id]]
+            raise MultipleIngestedEntityMatchingError(db_person, matches)
         ingested_person.person_id = db_person.person_id
-        matched_db_ids.add(db_person.person_id)  # type: ignore
+        matched_db_ids[db_person.person_id] = ingested_person  # type: ignore
```
terinpw: nit - do the cast (int) up here too to eliminate `# type: ignore`

PR #1375, opened 2019-03-20
Make snapshot writer handle provided end dates

zdg2102:


recidiviz/persistence/database/update_historical_snapshots.py
```diff
@@ -189,73 +198,126 @@ def _fetch_most_recent_snapshots_for_entity_type(
...
         context_registry: '_SnapshotContextRegistry') -> None:
     """For every entity in all record trees rooted at |root_people|, determines
     if the entity has a provided start time (i.e. the time the entity was
-    created or otherwise began) and sets it on the |context_registry|
+    created or otherwise began) and/or a provided end time (i.e. the time the
+    entity was completed or closed) and sets it on the |context_registry|
     """
     for person in root_people:
-        earliest_admission_date = None
+        earliest_booking_date = None
```
jamwalla:
```python
earliest_booking_date = min(b.admission_date for b in person.bookings
                            if b.admission_date is not None and
                            b.admission_date_inferred is False)
```

jamwalla: oh, no, it's more complicated than that. but what do you think about pulling out a `_get_earliest_booking_date(bookings)` function? I think it'd be easier to read separated out logically from the `_set_provided_start_and_end_time_for_booking_descendant` work this loop is doing.

recidiviz/persistence/database/update_historical_snapshots.py
```diff
@@ -189,73 +198,126 @@ def _fetch_most_recent_snapshots_for_entity_type(
...
         context_registry: '_SnapshotContextRegistry') -> None:
-    """Sets |entity| provided start time on |context_registry| according to
-    type of |entity|
+    """Sets |entity| provided start time and/or provided end time on
+    |context_registry| according to type of |entity|
     """
 
+    # If this method is called during graph exploration on parent booking,
+    # do nothing
+    if isinstance(entity, schema.Booking):
```
jamwalla: it doesn't look like `_get_related_entities` returns the parameter entity? if it does, could it have a flag like `include_this_entity=True`?

recidiviz/persistence/database/update_historical_snapshots.py
```diff
@@ -189,73 +198,126 @@ def _fetch_most_recent_snapshots_for_entity_type(
...
     # be treated as having the same start time as the parent booking
-    if parent_booking_admission_date:
-        start_time = _date_to_datetime(parent_booking_admission_date)
+    #
+    # Note the inverse of this does NOT apply to end time. We cannot assume
+    # child entities end when the parent booking ends.
+    if parent_booking_start_date:
+        start_time = _date_to_datetime(parent_booking_start_date)
 
     if type(entity).__name__ in _BOOKING_DESCENDANT_START_DATE_FIELD:
```
jamwalla: can this and the below be pulled out into a one line function with a meaningful name? `get_descendant_start_date_field_name` or something?

recidiviz/persistence/database/update_historical_snapshots.py
```diff
@@ -189,73 +198,126 @@ def _fetch_most_recent_snapshots_for_entity_type(
...
         if start_date:
             start_time = _date_to_datetime(start_date)
 
+    if type(entity).__name__ in _BOOKING_DESCENDANT_END_DATE_FIELD:
+        field_name = _BOOKING_DESCENDANT_END_DATE_FIELD[type(entity).__name__]
+        end_date = getattr(entity, field_name)
+        if end_date:
+            end_time = _date_to_datetime(end_date)
+
     if start_time:
```
jamwalla: this should be refactored, move it below line 283. same for end_time

recidiviz/persistence/database/update_historical_snapshots.py
```diff
@@ -265,27 +327,83 @@ def _write_snapshots(session: Session, context: '_SnapshotContext',
...
+
+    # Validate provided start and end times
     if context.provided_start_time and \
-            context.most_recent_snapshot is None and \
             context.provided_start_time.date() < snapshot_time.date():
-        new_historical_snapshot.valid_from = context.provided_start_time
+        provided_start_time = context.provided_start_time
+
+    if context.provided_end_time and \
+            context.provided_end_time.date() < snapshot_time.date():
```
jamwalla: why compare dates and not datetimes?

recidiviz/persistence/database/update_historical_snapshots.py
zdg2102: Discussed offline, all these date values need to be shared through the context of the tree, so it's probably simplest to read this way

recidiviz/persistence/database/update_historical_snapshots.py
zdg2102: Leaving this check keeps this robust if we decide to add cyclic relationships to the graph of the record tree (which we used to have before, so probably worth accounting for as a possibility)

recidiviz/persistence/database/update_historical_snapshots.py
zdg2102: Discussed offline, `start_time` can be set in two different conditionals so we can't move it inside either

recidiviz/persistence/database/update_historical_snapshots.py
zdg2102: We had to make this change in #1305, because datetimes out of Postgres have timezones and datetimes created for new entities don't, so the comparison raises an error

recidiviz/persistence/database/update_historical_snapshots.py
zdg2102: Done

Issue #873, opened 2019-02-06
Fix jails taking >12h to scrape


colincadams:
Currently `us_pa`, `us_fl_hendry`, and `us_fl_nassau` are all [still running at 9am each morning](https://app.google.stackdriver.com/metrics-explorer?isAutoRefresh=false&project=recidiviz-123&timeSelection=%7B%22timeRange%22:%22custom%22,%22start%22:%222019-02-05T18:44:20.674Z%22,%22end%22:%222019-02-06T18:44:20.674Z%22%7D&xyChart=%7B%22dataSets%22:%5B%7B%22timeSeriesFilter%22:%7B%22filter%22:%22metric.type%3D%5C%22cloudtasks.googleapis.com%2Fqueue%2Ftask_attempt_count%5C%22%20resource.type%3D%5C%22cloud_tasks_queue%5C%22%20metric.label.%5C%22response_code%5C%22%3D%5C%22ok%5C%22%22,%22perSeriesAligner%22:%22ALIGN_RATE%22,%22crossSeriesReducer%22:%22REDUCE_SUM%22,%22secondaryCrossSeriesReducer%22:%22REDUCE_NONE%22,%22minAlignmentPeriod%22:%22300s%22,%22groupByFields%22:%5B%22resource.label.%5C%22queue_id%5C%22%22%5D,%22unitOverride%22:%221%22%7D,%22targetAxis%22:%22Y1%22,%22plotType%22:%22LINE%22%7D%5D,%22options%22:%7B%22mode%22:%22COLOR%22%7D,%22constantLines%22:%5B%5D,%22timeshiftDuration%22:%220s%22,%22y1Axis%22:%7B%22label%22:%22y1Axis%22,%22scale%22:%22LINEAR%22%7D%7D). All the other scrapers are finished by 1am.

colincadams:
Maybe we can just bump up the qps until we have a more long term solution?

jessex:
@arian487 another potential impetus to prioritize design for multi-day ingest.

PR #1380, opened 2019-03-20
resume instead of start, pause instead of stop

arian487:
Fixes https://github.com/Recidiviz/recidiviz-data/issues/873


colincadams:
This doesn't work because queues can be shared right? I think if there is motivation to do this we need to do something fancier.

One possibility would be to ack all the tasks on the queue for a particular region and persist the queue requests for those elsewhere, then on resume re add them to the queue. However since there aren't transactions, we would need to make sure we are careful so that no tasks can slip through.

Maybe to handle this it is worth doing something similar to what we do for stop now, where if a task starts and its session is "paused" then it writes its queue request somewhere instead of running.

In either case, I think it's worth designing this more fully before we move forward.

PR #1381, opened 2019-03-20
setup copybara for intial transfer

colincadams:


mirror/copy.bara.sky
```diff
@@ -8,6 +8,15 @@ filesToPublish = glob(
         "recidiviz/ingest/scrape/{regions,vendors}/**",
         "recidiviz/tests/ingest/scrape/{regions,vendors}/**",
         "recidiviz/tools/scraper_template/**",
+        # old paths
```
jessex: :+1: These all looks right to me!

PR #1394, opened 2019-03-21
Add metric for scraper work and remove -1 path

colincadams:


recidiviz/ingest/scrape/base_scraper.py
```diff
@@ -58,6 +58,15 @@
     Task
 from recidiviz.persistence import batch_persistence, persistence
 
+class ParsingError(Exception):
+    """Exception containing the text that failed to parse"""
+    _MAX_TEXT_SIZE = 50 * 1024  # 50 KiB
```
jessex: What constrains us to 50 KiB, besides good sense? Should we make that explicit?

recidiviz/ingest/scrape/worker.py
```diff
@@ -18,19 +18,28 @@
...
 
+from opencensus.stats import aggregation, measure, view
 from recidiviz.ingest.models.scrape_key import ScrapeKey
 from recidiviz.ingest.scrape import sessions
 from recidiviz.ingest.scrape.task_params import QueueRequest
 from recidiviz.utils import monitoring, regions
 from recidiviz.utils.auth import authenticate_request
 
+m_tasks = measure.MeasureInt("ingest/scrape/task_count",
+                             "The count of scrape tasks that occured", "1")
```
jessex: occurred

Fun fact: one of the most misspelled words in the English language, for both native and non-native writers

jessex: Also below on line 39.

recidiviz/ingest/scrape/base_scraper.py
colincadams: the max log entry size is 100k: 
https://cloud.google.com/logging/quotas

I set this to 50k to be conservative, since this doesn't include the stacktrace or context exception. By default if you exceed the log entry size, stackdriver drops it instead of truncating it.

This brings up a good point though, we could instead have the log formatter in `structured_logging` do this truncation. Then we don't have control over what gets truncated, but cleans up this abstraction bleeding.

jessex: Hm, tough trade off. I feel like it would be _ideal_ to delegate things like truncation as deep into the logging stack as possible. Ideally nothing in our codebase would even care and Stackdriver would handle this itself, but if it has to be in our codebase then I think `structured_logging` out on the fringe is the right place. Plus, we should think that nothing we'd log elsewhere would approach the max size, right? Even when we log ingest info objects we log only a small sample which I don't think would be near this size. What do you think?

colincadams: I agree having it on the fringe is better.

It should be rare that we approach the max size, but it has happened before (which is why we now just log a sample of ingest info :grin:) and could happen again. Should be rare enough that truncating the message as a whole, even if that means cutting off more useful info, seems fine. And if we notice we are truncating log messages we should fix the underlying issue.

colincadams:
Correct, when an exception bubbles up to Flask it returns a 500:
http://flask.pocoo.org/docs/1.0/errorhandling/#handling

We can attach different error handlers if desired.

recidiviz/ingest/scrape/worker.py
colincadams: good catch!

colincadams:
I think I'd prefer to just truncate for now

PR #1400, opened 2019-03-21
dedup ingest people while batching

jamwalla:
as we collect IngestInfos from messages, dedup the Person objects. we do this before validating and converting since identical ingest_infos may be returned from different tasks and should not trigger validation errors (eg if they have the same external_id). rewrite of #1354


recidiviz/persistence/batch_persistence.py
```diff
@@ -179,35 +179,34 @@ def _get_proto_from_messages(messages):
...
                 failed_tasks[task_hash] = batch_message
+
+    deduped_ingest_info = _dedup_people(ingest_infos_from_messages)
+    base_proto = ingest_utils.convert_ingest_info_to_proto(deduped_ingest_info)
     logging.info('Generated proto for %s people', len(base_proto.people))
     return base_proto, failed_tasks
 
 
+def _dedup_people(ingest_infos: List[IngestInfo]) -> IngestInfo:
+    """Combines a list of IngestInfo objects into a single IngestInfo with
```
zacharylawrence: Could you add a quick comment explaining that we do this proto -> ingest_info.py conversion because it strips the generated ids?

colincadams: This doesn't actually do any conversion right? The batch message contains an ingest_info.py, that gets converted from a proto during the `from_serializable` call

jamwalla: right, no extra conversion at this step!

jamwalla:
Nah, as we discussed, let's trust the website to distinguish actually-differing people - and we'll make sure to keep them distinct in #1408

PR #1069, opened 2019-02-22
Start emulators in docker

zacharylawrence:
Co-authored-by: Colin Adams <colinadams@google.com>

PR #1403, opened 2019-03-21
remove emulator start instructions

colincadams:


colincadams:
pytest now brings them up for you, added in #1069 but we didn't update the docs

PR #1404, opened 2019-03-21
use threads to concurrently call pubsub

arian487:
This will hopefully protect us from pubsub guarantees (or lack thereof).


recidiviz/persistence/batch_persistence.py
```diff
@@ -105,36 +108,55 @@ def _get_batch_messages(scrape_key):
...
+                scrape_key, inner, pubsub_type=BATCH_PUBSUB_TYPE).\
+                received_messages
+            if recv_messages:
+                logging.info('Finished pulling messages, read %s messages',
+                             len(recv_messages))
+                q.put(recv_messages)
+                _ack_messages(messages, scrape_key)
+        # We occasionally get timeouts, we want to catch these and not kill the
+        # thread.
+        except Exception as e:
```
colincadams: I think that we should communicate back whether or not the pull succeeded, and then if all five fail with an exception we should fail batch and persist. Could do this using a `ThreadPoolExecutor`, or maybe we should be spawning these with gevent?

Otherwise if the first few requests go okay, but then we get a bunch of exceptions (i.e. we hit some quota limit or pubsub goes down), we will just continue as if those were all the messages that exist.

recidiviz/persistence/batch_persistence.py
```diff
@@ -38,17 +38,17 @@
...
 
 class BatchPersistError(Exception):
     """Raised when there was an error with batch persistence."""
 
-    def __init__(self, region: str, scrape_type: ScrapeType,
-                 scraper_start_time: datetime.datetime):
+    def __init__(self, region: str, scrape_type: ScrapeType):
         msg_template = "Error when running '{}' for region {} at {}"
-        msg = msg_template.format(scrape_type, region, scraper_start_time)
+        msg = msg_template.format(scrape_type, region)
```
colincadams: Why remove `scraper_start_time`? Make sure to update the msg_template - if you inline the template I think pylint will catch this.

recidiviz/persistence/batch_persistence.py
```diff
@@ -105,36 +105,62 @@ def _get_batch_messages(scrape_key):
...
-            logging.info('Finished acking the messages')
-            messages.extend(response.received_messages)
-        else:
+        pull_futures = []
+        all_failed = True
+        pulled_messages = []
+        for _ in range(NUM_PULL_THREADS):
+            t = pool.submit(async_pull)
+            pull_futures.append(t)
+        for pull_future in pull_futures:
```
colincadams: Can we do something like this:
```
pull_futures = [pool.submit(async_pull) for _ in range(NUM_PULL_THREADS)]
results, failed = zip(future.result() for future in concurrent.futures.as_completed(pull_futures))
if all(failed):
  raise
pulled_messages = itertools.chain(results)

recidiviz/persistence/batch_persistence.py
arian487: It doesn't feel too necessary and also it would be annoying to pipe to the function.

PR #1406, opened 2019-03-21
Make the proxy optional, true by default

arian487:
some regions return a bit too slowly for luminati, luminati also doesn't accept arbitrary ports so we must turn it off for these regions.


recidiviz/ingest/scrape/regions/us_fl_st_johns/manifest.yaml
```diff
@@ -4,3 +4,4 @@ base_url: http://www2.sjso.org/smartwebclient/jail.aspx
 timezone: America/New_York
 environment: production
 jurisdiction_id: 12109001
+should_proxy: false
```
colincadams: some newlines missing

PR #1408, opened 2019-03-21
choose best person match if multiple matches

### LINKED ISSUE #375: Disambiguate if a scraped entity matches to >1 database entity

terinpw:
fixes #375

Co-Authored-By: jamwalla <jamwalla@users.noreply.github.com>


jessex:
We only need to do this at the `Person` level and not anything else in the tree?

recidiviz/persistence/entity_matching.py
```diff
@@ -468,6 +469,27 @@ def _get_only_match(
...
+def _get_best_person_match(
+        ingested_person: entities.Person, db_people: List[entities.Person],
+        matched_db_ids: Iterable[int]) -> Optional[entities.Person]:
+    """Given the |ingested_person| returns the best match in the |db_people| if
+    any matches can be found."""
+
+    matches: List[entities.Person] = _get_all_matches(
+        ingested_person, db_people, utils.is_person_match)
+    matches = [p for p in matches if
+               cast(int, p.person_id) not in matched_db_ids]
```
jessex: We want the person id to _not_ be in this given list of matched database ids, in order to dub this a possible match? That seems backwards to me but I must be over- or under- or mis-thinking this.

recidiviz/persistence/entity_matching_utils.py
```diff
@@ -51,6 +53,30 @@ def is_person_match(
                 db_entity=db_entity, ingested_entity=ingested_entity))
 
 
+def _entity_to_set(entity: entities.Entity) -> Set[Tuple[str, str, Any]]:
+    """Returns a set of tuples for all fields and children of the |entity|,
+    in the form (ENTITY_NAME, FIELD_NAME, FIELD_VALUE)"""
+    fields = set()
+    for k, v in attr.asdict(entity, recurse=False).items():
```
jessex: I haven't looked at the docs, but `recurse=False` implies to me that this doesn't crawl the entire object tree, but only looks at top-level fields. Yet the docs say that this does look at children, so what am I missing?

recidiviz/persistence/entity_matching_utils.py
```diff
@@ -51,6 +53,30 @@ def is_person_match(
...
+            fields.update(_entity_to_set(v))
+        elif isinstance(v, list):
+            for sub_entity in v:
+                fields.update(_entity_to_set(sub_entity))
+        else:
+            fields.add((entity.get_entity_name(), k, v))
+    return fields
+
+
+def person_similarity(person_a: entities.Person,
```
jessex: Clever approach for ranking similarity! I'm curious, though: what are the cases where we expect multiple matches to occur? That helps evaluate the fitness of this approach.

recidiviz/persistence/entity_matching_utils.py
```diff
@@ -51,6 +53,30 @@ def is_person_match(
                 db_entity=db_entity, ingested_entity=ingested_entity))
 
 
+def _entity_to_set(entity: entities.Entity) -> Set[Tuple[str, str, Any]]:
```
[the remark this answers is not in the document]
rasmi: Case is similar to this, but in our approach, entity_name takes care of the nesting prefixes: https://stackoverflow.com/a/6027615

recidiviz/persistence/entity_matching.py
```diff
@@ -468,6 +469,27 @@ def _get_only_match(
...
+
+    if not matches:
+        return None
+
+    if len(matches) == 1:
+        return matches[0]
+
+    return max(matches,
```
rasmi: I think this is the correct heuristic, but it may be worth selecting a subset of the keys in `entity_matching_utils.person_similarity` that take precedence over raw cosine similarity (where all keys are weighted equally), or discarding a subset of keys we don't want to count towards a similarity score.

recidiviz/persistence/entity_matching_utils.py
rasmi: This is an implementation of cosine similarity -- a good tool to keep handy when thinking of heuristics/metrics for similarity in text-based problems!

recidiviz/persistence/entity_matching.py
rasmi: One more thing to consider, if the DB person is a fully-contained subset of both matches, then they'll both have the same similarity score -- maybe you'd want to choose the one with the larger number of total fields in that case to break the tie. 

You should also consider prioritizing fields in the case where matches have the same similarity score (e.g. if both match 6 fields but A's matched fields are charges and B's matched fields are less important data like bond amounts (for example?), prefer A).

Or have fields where if there isn't perfect match, then the candidate is discarded (e.g. name -- but I guess those essential fields are already handled by the first pass db entity matching, right?)

recidiviz/persistence/entity_matching_utils.py
jessex: Thanks for the coaching, Rasmi! This is an area I'm still learning.

rasmi: I only learned of it recently, I'm glad it came in handy!

recidiviz/persistence/entity_matching.py
jamwalla: We're trying to select a person in the database that matches our ingested person. `matched_db_ids` tracks the ids of database people we've already matched, so we want to not select these people since they've been matched to a different ingested person.

recidiviz/persistence/entity_matching_utils.py
jamwalla: Yeah, thanks for the suggestion Rasmi! The motivation for this ranking (at the person level) is to distinguish between people that look the same - potentially, same name, birthdate, compatible open-ness of recent booking. In these cases we think the best way to disambiguate is to look at differences further down in the tree; for example, distinguishing two identical people by their charges.

recidiviz/persistence/entity_matching_utils.py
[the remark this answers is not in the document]
jamwalla: Yep, we're flattening the object and dumping it into a set, so our tuples will won't collide across entities. For example, `('charge', 'status', PRESENT_WITHOUT_INFO) != ('bond', 'status', PRESENT WITHOUT INFO)`.

The weirdness here is that we *will* have collisions on repeated fields; for example if multiple charges have the same status we'll only have that charge-status-value once. Maybe a multiset (?) or counter is the right approach, then?

recidiviz/persistence/entity_matching.py
jamwalla: The perfect match case is handled by the first pass, but I think you're right that there are "weights" we should tiebreak with! If the person to match on is a full subset of two other people, wouldn't we want to choose the smaller one, since it has fewer non-matching fields?

jamwalla:
> We only need to do this at the `Person` level and not anything else in the tree?

`Person` is a weird case where it's possible to have collisions and we need to just pick one. Everywhere else:
 - EntityMatchingErrors at the `Booking` level mean a db person has multiple open bookings, violating a database constraint.
- EntityMatchingErrors at the `Hold` level mean an ingested person has multiple identical holds, violating a deduping that we do in the converter.
- EntityMatchingErrors on other repeated fields (`Charge`, `Bond`, `Sentence`) shouldn't occur because we'll pick "best" matches and drop/orphan entities as necessary.

recidiviz/persistence/entity_matching_utils.py
terinpw: For context - we've seen a few regions where there are multiple people w the same name and admission date with different sets of charges. 

Without this PR, we'll write both people (person A and B) into the DB on the first days scrape. On the next day when we're matching the ingested people - matching for both of the ingested version of Person A and B will fail as they match to 2 DB entities (and therefore they won't be written). Then the db versions of A and B will get marked released as they won't have been updated by the second day's scrape. 

As long as there are conflicts we'd continue this cycle of writing both people into our DB on Day 1 and then on Day 2 we'd fail to match the next day and therefore marking them released.

rasmi:
Another approach we discussed in person is doing a dict diff type comparison on the objects (where order doesn't matter for lists) then taking the object with the minimum diff as the match.

recidiviz/persistence/entity_matching_utils.py
jessex: So with this change, on Day 2 we'd match to whichever of Person A or B is the "closest" match, but we'd still keep the other one in the database, because we're not 100% sure that A and B are themselves matches. Is that right?

jamwalla: yep, if they were scraped on the same day we call them different people if they're not identical. The alternative, I guess, would be to "entity match" ingested entities first?

jessex: I would think the odds of a particular source providing non-identical but same-in-the-real-world persons, _none of whom we had never seen before_, in a single ingest session are low, _except in the case of the first day we ever scrape_.

Keep in mind that most of these sources do already have duplicates before we ever attempt ingest. So if our attempts at deduplication rely on the passage of time as a component of the  deduplication mechanism (as it does right now), then the first ever ingest of data from a source risking bringing in those duplicates and effectively cementing them in our database until they've phased out.

Does that make sense? Is that a real risk or do we have other safeguards I'm not considering?

jessex: Discussed offline. There is an edge case here wherein depending on the order of people scraped, we might match an ingested person to the wrong database person, but it seems like it should be a rare case.
* Here's the edge case: we had two near-identical people that we scraped earlier, one is released and then later ingested again, we ingest the still-incarcerated and the back-again person, and depending on how similar the back-again person is and the order in which the two are scraped, we might match the back-again person to the still-incarcerated person and turn the latest ingested still-incarcerated person into a brand new person in the database).
* Terin and James will think a bit to see if there's an easy fix for that or if it's too costly to handle.

More generally, we discussed the fact that today if two near-identical people are in the source database at the same time, then they will be ingested into two different people in our database. At the state level, this might not be the desired functionality, i.e. we will want to proactively deduplicate non-identical entities in the source while ingesting. Terin and James will file an issue with more detail should we need to make that change later.

terinpw: Filed #1447 to capture state level concerns.

jamwalla: Filed #1450 for the edge case!

PR #1409, opened 2019-03-21
Create backup manager endpoint

zdg2102:
This is just the endpoint, no cron. Once we deploy this, I'll be able to manually trigger the endpoint using my user auth to validate the behavior. I'll add the cron after the behavior is validated.


recidiviz/backup/backup_manager.py
```diff
@@ -0,0 +1,124 @@
...
+import flask
+import google.auth
+from googleapiclient import discovery
+
+from recidiviz.utils import metadata
+from recidiviz.utils import secrets
+from recidiviz.utils.auth import authenticate_request
+
+
+# Weekly backups for 6 months: 24 weekly backups
```
jessex: It's actually 26 if we're going for half a year.

recidiviz/backup/backup_manager.py
```diff
@@ -0,0 +1,124 @@
...
+    backup_runs = list_result['items']
+    manual_backup_runs = [backup_run for backup_run in backup_runs
+                          if backup_run['type'] == 'ON_DEMAND']
+    logging.info('Backup list request completed with %s total backup runs '
+                 'and %s manual backup runs',
+                 str(len(backup_runs)),
+                 str(len(manual_backup_runs)))
+
+    if len(manual_backup_runs) > _MAX_COUNT_MANUAL_BACKUPS:
+        # Backup runs are returned most recent first
```
jessex: Nitpicky but if the return order in the API changed, intentionally or otherwise, we'd likely start killing off our newest backup every time and effectively break the entire workflow. Is it possible to quickly verify that this is actually the oldest one we're deleting?

recidiviz/backup/backup_manager.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+@backup_manager_blueprint.route('/update_long_term_backups')
+@authenticate_request
+def update_long_term_backups() -> None:
+    """Create a new manual backup and delete the oldest manual backup once the
+    maximum number has been reached
+    """
+    credentials, _ = google.auth.default()
+    client = discovery.build('sqladmin', 'v1beta4', credentials=credentials)
```
colincadams: Since this won't be called often this is likely fine, but for other clients (i.e. cloud tasks) we create a single lazy-initialized client to save network roundtrips for client setup. Could do that here as well for consistency (see recidiviz/common/queues.py)

recidiviz/backup/backup_manager.py
zdg2102: The return order is explicitly specified in the documentation, so I figured it was safe to rely on (https://developers.google.com/resources/api-libraries/documentation/sqladmin/v1beta4/python/latest/sqladmin_v1beta4.backupRuns.html)

But I agree it's probably better to be safe than sorry, so I added a manual sort here

recidiviz/backup/backup_manager.py
zdg2102: (6 * 4) != (52 / 2). Clearly calendars should be metric to fix this :stuck_out_tongue: Updated to 26

recidiviz/backup/backup_manager.py
```diff
@@ -0,0 +1,144 @@
...
+            client, project_id, delete_operation['name'], 'delete')
+        logging.info('Backup delete operation completed')
+
+    logging.info('All backup operations completed successfully')
+
+
+def _await_operation(client: discovery.Resource,
+                     project_id: str,
+                     operation_id: str) -> None:
+    waiting = True
```
colincadams: nit: I'm no python expert, but it seems more pythonic to inline this
```
while True:
  if done:
    break

**Comment by zdg2102** [comment]
@arian487 once I've validated the behavior and set up the cron, I'll update the database README before closing #1113.

@rasmi I added the license text.

I'm not too worried about the timeout, since it doesn't take a very long time to trigger a backup manually through the Cloud SQL UI. We can always increase it if it starts to give problems.

I don't really think there's much logic worth testing here, I'm really just fetching an ID and passing it to another call. All the complexity is on the Cloud SQL side, and I'd have to mock all that out in tests anyway.

And I'm planning on updating `cron.yaml` in a separate PR. I'm going to merge this one first and then test it by manually hitting the endpoint after it's deployed.

File: recidiviz/backup/backup_manager.py [hunk 5]
```diff
@@ -0,0 +1,144 @@
...
+        errors = operation['error'].get('errors', [])
+        error_messages = ['code: {}\n message: {}'.format(
+            error['code'], error['message'])
+                          for error in errors]
+        joined_error_messages = '\n'.join(error_messages)
+        logging.error('Backup %s operation finished with %s errors:\n%s',
+                      operation_type,
+                      str(len(errors)),
+                      joined_error_messages)
+        raise RuntimeError('Error in backup operations. See preceding logs.')
```
colincadams: I think it would be easier to debug if we put the error string as a part of the exception instead of logging it separately

recidiviz/backup/__init__.py
```diff
@@ -0,0 +1,16 @@
+# Recidiviz - a platform for tracking granular recidivism metrics in real time
+# Copyright (C) 2018 Recidiviz, Inc.
```
colincadams: 2019

recidiviz/backup/backup_manager.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+def _get_instance_id() -> str:
```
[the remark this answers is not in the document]
zdg2102: Done

recidiviz/backup/backup_manager.py
```diff
@@ -0,0 +1,124 @@
...
+    """Create a new manual backup and delete the oldest manual backup once the
+    maximum number has been reached
+    """
+    credentials, _ = google.auth.default()
+    client = discovery.build('sqladmin', 'v1beta4', credentials=credentials)
+
+    project_id = metadata.project_id()
+    instance_id = _get_instance_id()
+
+    logging.info('Making request for backup insert operation')
```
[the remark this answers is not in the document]
zdg2102: Good point. Changed to "creating"

recidiviz/backup/backup_manager.py
zdg2102: I initially copied that pattern, and then I realized this is just a single method that's only called once a week, so the number of initializations will always be equal to the number of calls anyway. So I just instantiated it here to be a bit more readable.

recidiviz/backup/backup_manager.py
zdg2102: Done

recidiviz/backup/backup_manager.py
zdg2102: Done

recidiviz/backup/__init__.py
zdg2102: Done

rasmi:
Not sure if it's important since I think this should fail loudly, but any call to .execute() may throw a `googleapiclient.errors.HttpError` -- might be worth catching if you want certain behavior if failure happens after a specific step/request in the process.

zdg2102:
@rasmi I think it should be fine. All that really matters is that the delete doesn't happen if the insert fails, which will still happen if it throws `HttpError`

PR #1410, opened 2019-03-22
make batch and persist threading more succinct

colincadams:
Follow up to #1404 

@arian487 we were missing another `*` when passing into `itertools.chain`


rasmi:
Maybe use [`itertools.chain.from_iterable()`](https://docs.python.org/3/library/itertools.html#itertools.chain.from_iterable)?

colincadams:
good call @rasmi, thanks!

PR #1414, opened 2019-03-22
Write fips as strings to correctly zero pad

zacharylawrence:
#1413


rasmi:
Is this PR also meant to handle FIPS in jurisdiction_id?

zacharylawrence:
> Is this PR also meant to handle FIPS in jurisdiction_id?

Nope, that'll be done in a follow-up PR.

PR #1433, opened 2019-03-22
Add length check constraint for FIPS codes

zdg2102:


rasmi:
Can we do the same for jurisdiction_id = 8?

PR #1436, opened 2019-03-22
remove people w exceptions during entity_match

### LINKED ISSUE #1282: Do not persist ingested people if they failed entity matching
We currently loop over database people and attempt to match them with ingested people. If an exception is thrown, we increment an error counter, but we don't remove any entities from ingested people (which is eventually persisted)

Instead we should loop over ingested people and attempt to match them with the database people. This way, if one ingested person fails to be matched, we can easily remove them from the list of people to be persisted.

terinpw:
closes #1282


recidiviz/persistence/entity_matching.py
```diff
@@ -48,9 +50,22 @@
 monitoring.register_views([matching_errors_by_entity_view])
 
 
+@attr.s
+class EntityMatchingOutput(DefaultableAttr):
+    matched_people: List[entities.Person] = attr.ib(factory=list)
+    error_count: int = attr.ib(default=0)
+    orphaned_entities: List[entities.Entity] = attr.ib(factory=list)
+
+    def __iadd__(self, other):
```
jamwalla: honestly this is lit. but maybe rather than partially implement the assignment operators we could call this `update`?

terinpw: Chatted offline w @jamwalla and added the __add__ method and made the object immutable.

recidiviz/persistence/entity_matching.py
```diff
@@ -48,76 +49,90 @@
 monitoring.register_views([matching_errors_by_entity_view])
 
 
+@attr.s(frozen=True)
```
zacharylawrence: Just noting that if you want to always force kwargs, you could pass `kw_only=True` here too

PR #1437, opened 2019-03-22
Quotes coding style

### LINKED ISSUE #270: Decide on pattern for single versus double quotes
We're inconsistent across the codebase as to when we use single quotes or double quotes. I don't think there's an official Python standard, but we should at least adopt a reasonable pattern across our codebase. [Here's one such pattern](https://stackoverflow.com/questions/56011/single-quotes-vs-double-quotes-in-python) that seems sensible.

I don't think we should try to bake this into linting. Feels fragile and perhaps not too necessary.

jjdressel11:
Updated a lot of files to conform to our new python quoting style, following [this pattern](https://stackoverflow.com/questions/56011/single-quotes-vs-double-quotes-in-python). Updated the contribution guidelines to reflect this. 

From now on, this quote style will be manually enforced.

Closes #270


recidiviz/calculator/bq/bq_load.py
```diff
@@ -74,8 +74,8 @@ def start_table_load(
         ]
     except KeyError:
         logging.exception(
-            'Unknown table name "%s". Is it listed in '
-            'export_config.TABLES_TO_EXPORT?', table_name)
+            "Unknown table name \"%s\". Is it listed in "
```
jessex: I wonder if instead of escaped double quotes inside of human-readable text we should just use single quotes? 

`Unknown table name "blah". Is it listed in export_config.TABLES_TO_EXPORT?`

`Unknown table name 'blah'. Is it listed in export_config.TABLES_TO_EXPORT?`

I think the first is more formal but at the micro- but non-zero readability cost of having to scan over escaped characters.

jjdressel11:
Resolving merge conflicts now...

recidiviz/calculator/bq/bq_load.py
jjdressel11: Yeah, there were a few places where this happened, and I think it's totally fine to use single quotes in these cases.

CONTRIBUTING.md
```diff
@@ -154,3 +154,28 @@ The [linting configuration](.pylintrc) should adhere to this style guide and
 help you follow it automatically. You can use Pylint support for warning
 suppression if you find it necessary, but this may be argued against during
 code review.
+
+### Python Quoting Style
```
jessex: This looks great! I think this shows that we are likely to add notes here and there to "our" style guide and that in time we should just put that in a separate document, and point to that from the contribution guidelines. Having a stand alone document to reference on style guide is nicer than pointing folks to a sub-section of these guidelines, and it keeps this doc more focused.

But I don't think we should do that now! I just filed #1438 to do some time later on.

jjdressel11: Agreed!

recidiviz/persistence/persistence.py
```diff
@@ -248,19 +248,22 @@ def write(ingest_info, metadata):
         session = Session()
         try:
             logging.info("Starting entity matching")
-            entity_matching_errors, orphaned_entities = \
-                entity_matching.match(session, metadata.region, people)
+            entity_matching_output = entity_matching.match(
+                session, metadata.region, people)
+            people = entity_matching_output.people
+            # entity_matching_errors, orphaned_entities = \
```
jessex: I think we want to drop this comment.

jjdressel11: ✅

PR #1476, opened 2019-03-25
match historical bookings

jamwalla:
Co-Authored-By: jamwalla <jamwalla@users.noreply.github.com>


jamwalla:
fixed in #1482 and #1504 instead.

Issue #1478, opened 2019-03-25
Remove goaway grpc wrapper when fixed


arian487:
When the goaway issues are fixed, we should remove the retry wrapper.

PR #1479, opened 2019-03-25
Add a retry wrapper around goaway errors

arian487:
When these issues are fixed, we should remove this code.  https://github.com/Recidiviz/recidiviz-data/issues/1478


recidiviz/tests/common/common_utils_test.py
```diff
@@ -34,3 +40,38 @@ def test_is_generated_id(self):
     def test_is_not_generated_id(self):
         id_str = "id_str"
         self.assertFalse(is_generated_id(id_str))
+
+    def test_retry_grpc_no_raise(self):
+
+        def fn(a, b=1):
+            self.call_count += 1
```
colincadams: can you just define `call_count = 0` at the start of the test method?

recidiviz/common/queues.py
```diff
@@ -159,4 +172,9 @@ def create_bq_task(table_name: str, url: str):
             'body': json.dumps(body).encode()
         }
     )
-    client().create_task(format_queue_path(BIGQUERY_QUEUE), task)
+    retry_grpc_goaway(
+        NUM_GRPC_RETRIES,
```
colincadams: This is flipped, although I think I prefer it this way because it keeps the function and it's arguments together. WDYT?

recidiviz/common/common_utils.py
```diff
@@ -49,6 +56,23 @@ def get_trace_id_from_flask():
     return trace_id
 
 
+def retry_grpc_goaway(fn, num_retries, *args, **kwargs):
+    """Retries a function call some number of times"""
+    time_to_sleep = random.randint(5, RETRY_SLEEP)
```
colincadams: `random.uniform` instead? Since many calls fail simultaneously I think it'd be better to spread them out uniformly instead of on discrete second boundaries.

recidiviz/common/common_utils.py
```diff
@@ -49,6 +56,23 @@ def get_trace_id_from_flask():
...
+    for i in range(num_retries + 1):
+        try:
+            return fn(*args, **kwargs)
+        except exceptions.InternalServerError as e:
+            if 'GOAWAY' not in str(e):
+                raise e
+            if i == num_retries:
+                raise e
+            logging.error('GOAWAY received, sleeping %s seconds and retrying',
```
colincadams: I think this would be better as an info or warn log. If it fails on all retries we will get an error at the end as expected.

colincadams: Also the format specifier should be updated to `%d` or `%.2f` if you change to float

PR #1493, opened 2019-03-25
explicitly list fields to match on

### LINKED ISSUE #1488: explicitly list the entity fields we match over
let's define on each Entity the set of fields that define a match for entity matching. maybe add a `.match` method.

jamwalla:
close #1488

Co-authored-by: terinpw <terinpw@users.noreply.github.com>
Co-authored-by: jamwalla <jamwalla@users.noreply.github.com>


recidiviz/persistence/entity_matching_utils.py
```diff
@@ -16,18 +16,60 @@
...
+    'level',
+    'fee_dollars',
+    'charging_entity',
+    'court_type',
+    'case_number',
+}
+
+_HOLD_MATCH_FIELDS = {
+    'jurisdiction_name',
+    'status',
```
terinpw: from @ohinds - remove status and raw text from hold

recidiviz/persistence/entity_matching_utils.py
```diff
@@ -16,18 +16,60 @@
...
+    'bond_agent',
+}
+
+_SENTENCE_MATCH_FIELDS = {
+    'sentencing_region',
+    'min_length_days',
+    'max_length_days',
+    'date_imposed',
+    'completion_date',
+    'projected_completion_date',
```
terinpw: per @ohinds remove completion and projected completion from sentence

recidiviz/persistence/entity_matching_utils.py
```diff
@@ -16,18 +16,60 @@
...
 import logging
-from typing import Callable, Optional
+from typing import Optional, Set
 
 import deepdiff
 
-from recidiviz.common.constants.bond import BondStatus
 from recidiviz.persistence import entities
 from recidiviz.persistence.errors import PersistenceError
 from recidiviz.persistence.persistence_utils import is_booking_active
```
terinpw: For all of these we should only match on raw text and not the enum itself, as our mappings could change.

PR #1512, opened 2019-03-26
Fix all the enum parsing errors from last night.

ohinds:


recidiviz/ingest/scrape/regions/us_al_marion/us_al_marion_scraper.py
```diff
@@ -38,7 +38,7 @@ def populate_data(self, content, task: Task,
 
         person = scraper_utils.one('person', scraped_data)
 
-        if person.gender and person.gender == 'H':
+        if person.gender and person.gender == 'H' or person.gender == 'W':
```
jamwalla: if gender = 'W' we want to set ethnicity = hispanic?

also this never needed the `if person.gender` check, huh

recidiviz/ingest/scrape/regions/us_ar_faulkner/us_ar_faulkner_scraper.py
```diff
@@ -30,5 +30,6 @@ def __init__(self):
     def get_enum_overrides(self) -> EnumOverrides:
         overrides_builder = super(UsArFaulknerScraper,
                                   self).get_enum_overrides().to_builder()
-        overrides_builder.add('A', Race.ASIAN)
+        overrides_builder.ignore('B', Gender)
```
jamwalla: map to race?

PR #1379, opened 2019-03-20
Add community supervision and incarceration.

ohinds:
Formly info in sentence.

PR #1518, opened 2019-03-26
State Schema - incarceration and supervision

jessex:
Apologies for the massive PR. This adds the first pass at a state level schema, developed based on notes from Oliver and feedback from a few different partners. Its entry point into the current schema is to add `Incarceration` and `Supervision` as child entities for `Sentence`, which shifts the latter's focus to the imposition of the sentence itself, with the child entities denoting the results of that sentence. 

@ohinds has issued a PR that includes incarceration and supervision entities, but this should be a strict superset of that because this includes more fields on the two primary new entities and also more entities (children of those two). The one difference here is that #1379 adds incarceration as an array to sentence and mine adds only a single incarceration (we both add an array of supervisions, though). That's worth discussing here.

This covers everything except for an actual database migration and updates to the `database.write` flow itself to save into these new tables, which I understand @ohinds to be working on. That should be discussed, as well. Otherwise, this appears to introduce no regression and lays the groundwork to begin ingest state-level data, though obviously these new entities will be used at the county-level as well. 

This is ready for full review, and is best navigated by going through commit by commit, which are mostly neatly isolated.


arian487:
As a high level comment, I think a doc outlining the design, schema, and requirements would be helpful.  Not necessary for this PR but IMO this deserves its own doc.  It also feels like there should be a bit more separation here.  Our original schema was more of a jails schema, so I'm not sure if it makes sense to add these state level concepts to the ingest_info models for examples.  They should probably have their own models honestly, and even their own converters and database files.  

Ideally we'd have a base model that both jail ingest info and state ingest info inherit from for example and expand some fields as well as add new concepts (assessment for example).  The same goes for the converter and anything else.  Mixing the logic might get hairy, and separating the logic out will be more readable, but also allow for much easier changes in the future. 

Right now we're mixing sentences for example, is it true that the new sentence definition should apply to both the state level schema and the jails schema?  If not, someone reading the code later will be very confused why we have fields that are never touched by any jail scraper for example.

Ideally I'd have a converter/base_converter.py which has some base level stuff that is shared, and then a converter/jail/ directory which has jail_converter.py as well as converter definitions for some jail stuff (maybe jail sentences are different, and call for different converters, and perhaps could even have a base which is shared by jail/state).  Similarly you have converter/state/ with state_converter.py and all of the extra converters that youi don't need elsewhere.

None of this is required for this PR, but I feel confident that you will be much, much happier in the future if you separate out the logic a bit

jessex:
Thanks for the feedback, @arian487!

A schema doc is definitely coming. I am still holding out hope for more documentation to land before the end of this week that I can leverage. I also didn't want to invest in thorough docs until this was ready to go.

As for the general comments on separation, I think it's possible that we'd want to define a cleaner separation between county jails and state prisons but I am not sure there's enough information yet to call for a complete bifurcation just yet -- my thinking being, it's easy to take what I've done here and split between county and state, but relatively challenging to re-consolidate if the split has already occurred. Case in point, I've just dangled all of this off of `Sentence` and am not populating it from any counties yet. If I wanted to take this into a state-specific flow, I could just rip it all out in one fell swoop, for the most part.

But I do very much value that feedback and I think it may be the route we eventually pursue! I'd be curious to hear @ohinds' reflections on that comment, given that he had begun pursuing the same route.

recidiviz/common/constants/assessment.py
```diff
@@ -0,0 +1,61 @@
...
+
+"""Constants related to an assessment entity."""
+
+import recidiviz.common.constants.enum_canonical_strings as enum_strings
+from recidiviz.common.constants.entity_enum import EntityEnum, EntityEnumMeta
+
+
+class AssessmentClass(EntityEnum, metaclass=EntityEnumMeta):
+    MENTAL_HEALTH = enum_strings.assessment_class_mental_health
+    OTHER = enum_strings.assessment_class_other
```
zdg2102: We've mostly been avoiding having "Other" values in any of the enums (with the exception of fields like Race where a lot of the ingested sources explicitly use the word "Other" in their own classifications) and opting for either extending the enum with new values or setting the field as None. Is this "Other" value coming from the classification used in the ingested source?

recidiviz/common/constants/incident.py
```diff
@@ -0,0 +1,59 @@
...
+    'VIOLENT': IncidentOffense.VIOLENT,
+    'VIOLENCE': IncidentOffense.VIOLENT,
+}
+
+
+_INCIDENT_OUTCOME_MAP = {
+    'PRIVILEGE LOSS': IncidentOutcome.PRIVILEGE_LOSS,
+    'LOSS OF PRIVILEGE': IncidentOutcome.PRIVILEGE_LOSS,
+    'SHOT': IncidentOutcome.SHOT,
+    'REPRIMAND': IncidentOutcome.SHOT,
```
zdg2102: Is this mapping wrong, or does `SHOT` have a different meaning here than I'm reading it as?

recidiviz/ingest/models/ingest_info.proto
```diff
@@ -72,6 +92,7 @@ message Booking {
     optional string arrest_id = 10;
     repeated string charge_ids = 11;
     repeated string hold_ids = 12;
+	optional string prior_booking_id = 14;
```
zdg2102: Might just be an artifact of the github UI, but the indentation here looks odd

recidiviz/persistence/converter/assessment.py
```diff
@@ -0,0 +1,52 @@
...
+                              proto, metadata.enum_overrides)
+    new.assessment_class_raw_text = fn(normalize, 'assessment_class', proto)
+    new.assessment_type = fn(AssessmentType.parse, 'assessment_type',
+                             proto, metadata.enum_overrides)
+    new.assessment_type_raw_text = fn(normalize, 'assessment_type', proto)
+    new.assessment_score = fn(parse_int, 'assessment_score', proto)
+    new.assessment_level = fn(normalize, 'assessment_level', proto)
+
+    # TODO convert this into a map of key-value pairs when we first ingest this
+    new.assessment_metadata = fn(normalize, 'assessment_metadata', proto)
```
zdg2102: I don't really understand from the context, what is this field intended to represent?

recidiviz/persistence/converter/booking.py
```diff
@@ -60,6 +60,7 @@ def copy_fields_to_builder(booking_builder, proto, metadata):
     new.external_id = fn(parse_external_id, 'booking_id', proto)
     new.projected_release_date = fn(parse_date, 'projected_release_date', proto)
     new.facility = fn(normalize, 'facility', proto)
+    new.prior_booking_id = fn(parse_external_id, 'prior_booking_id', proto)
```
zdg2102: Is this meant for an external ID we're ingesting from the source system (i.e. not our own primary key)?

recidiviz/persistence/database/schema.py
```diff
@@ -281,6 +381,60 @@ class PersonHistory(Base, DatabaseEntity, _PersonSharedColumns):
     valid_to = Column(DateTime)
 
 
+class _AssessmentSharedColumns:
```
zdg2102: I'm somewhat confused about this type. What is it and why does it apply directly to Person?

recidiviz/persistence/database/schema.py
```diff
@@ -318,6 +472,10 @@ def __new__(cls, *_, **__):
     def person_id(self):
         return Column(Integer, ForeignKey('person.person_id'), nullable=False)
 
+    @declared_attr
+    def prior_booking_id(self):
```
zdg2102: If this is ingested as an external ID from the source system, when do we normalize it to our corresponding internal primary key?

recidiviz/tests/ingest/models/ingest_info_test.py
```diff
@@ -91,7 +116,41 @@ def test_prune(self):
...
+                                                   ),
+                                                   ingest_info.Incident()
+                                               ],
+                                               parole_decisions=[
+                                                   ingest_info.ParoleDecision()
+                                               ]
+                                           ),
+                                           supervisions=[
+                                               ingest_info.Supervision(),
+                                               ingest_info.Supervision(
```
zdg2102: Can we construct some of these in separate calls outside of this block so we can avoid this level of nesting?

recidiviz/tests/persistence/converter/disciplinary_action.py
```diff
@@ -0,0 +1,85 @@
+# Recidiviz - a data platform for criminal justice reform
```
zdg2102: I think the filename would be `disciplinary_action_test` to match the pattern we're using elsewhere

